PR #161, opened 2022-06-07
Add Swahili translations

nickmwangemi:


africanlii/templates/africanlii/set_language.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}
```
longhotsummer, line 1: All of this will have to move to peachjam (rather than africanlii) at some point, so I recommend doing it from the start.

longhotsummer:
@nickmwangemi i've merged in updated translations into main, which will clash with yours. They're now managed on crowdin interprise and you should have an invitation.

You should be able to simply resolve the conflicts by replacing your .po files with those from main.

nickmwangemi:
@longhotsummer Noted. Will do.

africanlii/templates/peachjam/set_language.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}
+
+<form action="{% url 'set_language' %}" method="post">{% csrf_token %}
+    <input name="next" type="hidden" value="{{ redirect_to }}">
+    <select name="language" onChange="form.submit();">
```
ro6ley, line 5:
```suggestion
    <select  class="form-select" name="language" onChange="form.submit();">
```

ro6ley (dismissed):
Just the minor suggestion to use bootstrap's select widget, otherwise LGTM.

PR #163, opened 2022-06-08
Multiple Document Versions Support

### LINKED ISSUE #128: As a multi-lingual user I want to be able to choose to view different language versions of a document that are available
Some documents are available in multiple languages.

In the metadata block, add an option that (a) helps the user realise there are multiple languages, (b) is clear which one they're looking at, and (c) provides links to open alternate language versions.

Perhaps the simplest is to make the current language in the `Language: English` metadata block a bootstrap dropdown, that shows the alternate language versions as links.

Since there can also be multiple date versions, only show languages with documents at the current document date. This prevents the user from unexpectedly changing document dates when they change the language.

### LINKED ISSUE #129: As an advanced reader I want to be able to choose historical versions of a document
Some documents are available in multiple versions (dates).

In the metadata block, add an option that (a) helps the user realise there are multiple dates, (b) is clear which one they're looking at, and (c) provides links to open alternate date versions.

Perhaps the simplest is to make the current date in the `Date: xxx` metadata block a bootstrap dropdown, that shows the alternate date versions as links.

Bear in mind that there may be multiple dates AND multiple languages. How should it be handled if there are dates that aren't available in the current language?

In time, we'll add a rich timeline for multiple date versions, but this is a quick-and-easy win for now.

ro6ley:
This PR adds support for multiple document versions based on date and language.

* [x] Make the current date in the `Date: xxx` metadata block a bootstrap dropdown, that shows the alternate date versions as links.
* [x] Make the current language in the `Language: English` metadata block a bootstrap dropdown, that shows the alternate language versions as links.

### Screenshots
* **Multiple language versions for the same date:**
  image

* **Multiple date versions for the same language:**
  image

* **Document without other date or language versions:**
  image

closes #128 
closes #129

### DIFF
--- a/africanlii/views/generic_views.py
+++ b/africanlii/views/generic_views.py
@@ -1,4 +1,4 @@
-from django.views.generic import ListView
+from django.views.generic import DetailView, ListView
 
 from africanlii.forms import BaseDocumentFilterForm
 from africanlii.models import AuthoringBody, Court
@@ -56,10 +56,10 @@ class FilteredDocumentListView(ListView, BaseDocumentFilterForm):
         return context
 
 
-class DocumentVersionsMixin:
-    """Helper mixin for document detail views to fetch other versions of a document based
-    on the document's work_frbr_uri and separate them based on language and date.
-    """
+class BaseDocumentDetailView(DetailView):
+    slug_field = "expression_frbr_uri"
+    slug_url_kwarg = "expression_frbr_uri"
+    context_object_name = "document"
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
@@ -72,6 +72,8 @@ class DocumentVersionsMixin:
         context["language_versions"] = all_versions.filter(date=self.object.date)
 
         # date versions that match current document language
-        context["date_versions"] = all_versions.filter(language=self.object.language)
+        context["date_versions"] = all_versions.filter(
+            language=self.object.language
+        ).order_by("-date")
 
         return context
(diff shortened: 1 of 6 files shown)


africanlii/views/judgment.py
```diff
@@ -17,7 +20,7 @@ def get_queryset(self):
 
 
 @registry.register_doc_type("judgment")
-class JudgmentDetailView(DetailView):
+class JudgmentDetailView(DocumentVersionsMixin, DetailView):
```
longhotsummer, line 23: It's starting to look like a common `DocumentDetailView` base class would be useful that would be used by all document detail views.

africanlii/views/generic_views.py
```diff
@@ -53,3 +54,24 @@ def get_context_data(self, **kwargs):
...
+        # get all versions that match current document work_frbr_uri
+        all_versions = CoreDocument.objects.filter(
+            work_frbr_uri=self.object.work_frbr_uri
+        ).exclude(pk=self.object.pk)
+
+        # language versions that match current document date
+        context["language_versions"] = all_versions.filter(date=self.object.date)
+
+        # date versions that match current document language
+        context["date_versions"] = all_versions.filter(language=self.object.language)
```
longhotsummer, line 75: Let's sort these most recent first (`-date`).

africanlii/views/judgment.py
ro6ley: I had this at some point 😅

longhotsummer (approved):

PR #165, opened 2022-06-09
Indigo documents ingestor

### LINKED ISSUE #67: As an editor I'd like documents to be ingested from Indigo automatically
Indigo (edit.laws.africa, etc.) is authoritative for certain document types, such as legislation. Peachjam must be able to pull information from a remote source. It must be configurable and robust and retry in the face of failure.

# Ingestors

Ingestors are a new concrete model in Peachjam. An ingestor is created for each document source (roughly speaking). An ingestor is a way for the an administrator to monitor and configure connections to external data sources. An ingestor is associated with an Adapter. An adapter is a concrete class (not a Django model) that holds the business logic to fetch information from a source and load it into Peachjam.

An ingestor can be configured with settings that the adapter uses to do its work, such as URL, username or password required to connect to the remote source.

## Ingestor data model

* name: string, required, unique
* enabled: boolean, defaults to false, required
* adapter: char field, can be blank, list of choices is taken from the list of adapter classes registered with the system
* last_refreshed_at: datetime, nullable -- the last time the adapter ran successfully; see below for how this is used

Indigo signoff has a similar concept (PublicationOutlet) here: https://github.com/laws-africa/indigo-signoff/blob/master/indigo_signoff/models/outlets.py#L21

## IngestorSettings data model

* ingestor: FK
* name: string, required
* value: string, optional
* (ingestor, name) must be unique

This represents arbitrary key-value settings pairs for an ingestor.

See how Indigo Signoff does it here: https://github.com/laws-africa/indigo-signoff/blob/master/indigo_signoff/models/outlets.py#L149

See how Indigo Signoff allows the adapter to validate the configuration settings here: https://github.com/laws-africa/indigo-signoff/blob/master/indigo_signoff/admin.py

## Adapter data model

The Adapter base class is a regular python class that takes a the settings as a dictionary. It implements `clean_settings()` which raises validation errors if the settings are bad - this is used by the admin form to let the adapter complain if the ingestor settings aren't correct.

It has a name property (or a name variable) which provides a unique name for the adapter, for use when linking the ingestor to an adapter.

```python
class IngestorAdapter:
  def __init__(self, settings=None):
    self.settings = settings

  def clean_settings(self):
    # validate that self.settings are decent
  ...
```

The adapter class must also have two methods that actually perform the updates (described below). The first finds new documents that has changed or been added since the last time the ingestor ran. The second updates a particular document only.

See Indigo Signoff base adapter here: https://github.com/laws-africa/indigo-signoff/blob/master/indigo_signoff/adapters.py

# How ingestors work

Ingestors use [background tasks](https://github.com/arteria/django-background-tasks) to fetch content in the background.

## Daily updates

Every day, each ingestor is told to look for updated information since it was last run. A background task wakes up and queues up a new background task, one for each active ingestor. The background tasks should be linked to the ingestor using the `creator` param when creating the task.

When the task runs:

* The ingestor tells its adapter to check for updates since the last time the ingestor ran, based on the `last_refreshed_at` timestamp (which may be null).
* The adapter does whatever necessary to identify what content is new, based on the timestamp (simplest if the adapter is given the ingestor object, rather than just the timestamp).
* The adapter can either update what is new, or (preferred) it queues up new background tasks to update just a single updated item. The latter has better failure semantics if there are hundreds of updates. The new background tasks should also be linked to the ingestor using the `creator` keyword.
* Finally, the ingestor updates its `last_refreshed_at` timestamp to `now()` and saves itself.

The secondary background tasks (for updating a single document) run by instantiating the ingestor, getting the adapter, and telling the adapter that it must update a single document (based on some key or other information).

## Ad-hoc updates

A new UI lists ingestors and lets admin users poke the ingestors to get them to run immediately. This is useful when we know there is new content to ingest.

A user can also tell an ingestor to refresh itself completely, which effectively sets `last_refreshed_at` to null and then queues up a full update task.

The UI also shows the status and last run time of each ingestor.

## Triggered updates

In future, we may have webhooks that can receive notifications from other systems that there is new information, which then effectively trigger a "daily update" run for an ingestor.

### LINKED ISSUE #87: As a user I don't want to see TOC entries below the paragraph level
When importing the TOC from indigo, remove children for:

* any element that has `basic_unit = True`
* anything where the type is `paragraph`

actlikewill:
- Sets up the ingestor and adapter for ingesting documents from indigo

closes #67 
closes #87


peachjam/adapters/adapters.py
```diff
@@ -0,0 +1,54 @@
...
+        from africanlii.models import GenericDocument
+
+        logger.info(f"Doc update {document['title']}")
+        field_data = {
+            "title": document["title"],
+            "created_at": document["created_at"],
+            "updated_at": document["updated_at"],
+            "work_frbr_uri": document["frbr_uri"],
+            "expression_frbr_uri": document["expression_frbr_uri"],
+            "date": document["publication_date"],
```
longhotsummer, line 42:
```suggestion
            "date": document["expression_date"],

```

longhotsummer: Also set language

peachjam/adapters/adapters.py
```diff
@@ -0,0 +1,54 @@
...
+        self.client = requests.session()
+        self.client.headers.update(
+            {
+                "Authorization": f"Token {token}",
+            }
+        )
+        self.url = url
+
+    def get_doc_list(self):
+        r = self.client.get(self.url).json()
```
longhotsummer, line 21: `self.client.raise_for_status()`

peachjam/adapters/adapters.py
```diff
@@ -0,0 +1,54 @@
...
+            "date": document["publication_date"],
+        }
+
+        if document["publication_document"]:
+            source_url = document["publication_document"]["url"]
+            field_data["source_url"] = source_url
+            # Here we would download the source file
+            # download_source_file(source_url)
+
+        doc = GenericDocument(**field_data)
```
longhotsummer, line 51: Use the `Legislation` document type.

peachjam/models/ingestors.py
```diff
@@ -0,0 +1,40 @@
...
+from peachjam.plugins import plugins
+from peachjam.tasks import update_document
+
+
+def adapter_choices():
+    # TOOD: use these choices in the ingestor admin form
+    return [(key, p.name) for key, p in plugins.registry["ingestor-adapter"].items()]
+
+
+class Ingestor(models.Model):
```
longhotsummer, line 13: Also create an `IngesterSettings` model.

peachjam/forms.py
```diff
@@ -0,0 +1,18 @@
...
+from peachjam.models import Ingestor
+from peachjam.plugins import plugins
+
+
+def adapter_choices():
+    return [(key, p.name) for key, p in plugins.registry["ingestor-adapter"].items()]
+
+
+class IngestorForm(forms.ModelForm):
+    adapter = forms.ChoiceField(choices=adapter_choices())
```
longhotsummer, line 12:
```suggestion
    adapter = forms.ChoiceField(choices=adapter_choices)

```

peachjam/models/ingestors.py
```diff
@@ -0,0 +1,41 @@
+from django.db import models
+from django.utils import timezone
+
+from peachjam.adapters import IndigoAdapter  # noqa
```
actlikewill, line 4: @longhotsummer I have imported the adapter here even though I am not using it in this file because the plugin decorator needs to run to have the adapter in the registry. Is there a better way to do this to have the adapters in the registry?

peachjam/adapters/adapters.py
```diff
@@ -0,0 +1,150 @@
...
+                if (k == "type" and v == "paragraph") or (
+                    k == "basic_unit" and v is True
+                ):
+                    d["children"] = []
+                elif isinstance(v, dict):
+                    remove_sub_paragraph(v)
+                elif isinstance(v, list):
+                    for i in v:
+                        if isinstance(i, dict):
+                            remove_sub_paragraph(i)
```
longhotsummer, line 125: Items are lists of dictionaries, so this should be easier:


```suggestion
            if d['type'] == 'paragraph' or d['basic_unit']:
                d['children'] = []
            else:
                for kid in d['children']:
                    remove_sub_paragraph(kid)
```

peachjam/adapters/adapters.py
```diff
@@ -0,0 +1,150 @@
...
+
+class Adapter:
+    def __init__(self, settings):
+        self.client = requests.session()
+        self.client.headers.update(
+            {
+                "Authorization": f"Token {settings['token']}",
+            }
+        )
+        self.url = settings["url"]
```
longhotsummer, line 22: We can't make any assumptions about what the adapter wants to do with the settings.

```suggestion
    def __init__(self, settings):
        self.settings = settings
```

peachjam/adapters/adapters.py
```diff
@@ -0,0 +1,150 @@
...
+        """Update the document identified by some opaque id, returned by check_for_updates."""
+        raise NotImplementedError()
+
+    @classmethod
+    def name(cls):
+        return cls.__name__
+
+
+@plugins.register("ingestor-adapter")
+class IndigoAdapter(Adapter):
```
longhotsummer, line 40: Setting up the `self.client` object based on the settings must happen here, as part of the constructor, after the superclass's constructor has been called.

peachjam/adapters/adapters.py
```diff
@@ -0,0 +1,150 @@
...
+            "toc_json": toc_json,
+            "content_html": content_html,
+            "work": work,
+        }
+
+        doc, _ = Legislation.objects.update_or_create(
+            expression_frbr_uri=expression_frbr_uri, defaults={**field_data}
+        )
+        if document["publication_document"]:
+            self.download_source_file(document["publication_document"], doc)
```
longhotsummer, line 106: I'd like to treat this a bit differently. The idea behind peachjam's sourcefile is that it's the "original" file which the user can download.

With Indigo, the publication document is strictly the original source file, but often that is a huge gazette that doesn't have all amendments.

Instead, we need to download the rendered PDF from Indigo and attach that as the source file.

To get the PDF, fetch `document["url"] + ".pdf"` much like you do for the publication document below, and then stash that file.

longhotsummer:
@actlikewill looking great! Minor comments added, we should be ready to deploy this very soon.

peachjam/adapters/adapters.py
actlikewill: Much simpler. Thanks 🙂

peachjam/adapters/adapters.py
actlikewill: Changed

peachjam/adapters/adapters.py
actlikewill: Changed. I am assuming that the api always generates a pdf at that url and so it will always be present?

longhotsummer: Yes, correct.

longhotsummer (approved):

Issue #126, opened 2022-05-24
As a Swahili speaker I'd like to be able to read the site in Swahili


longhotsummer:
* ensure all strings are translated to Swahili on crowdin
* add a toggle (at the right of the menu bar?) that lets the user change the language of the site
* use django's [cookie-based language system](https://docs.djangoproject.com/en/4.0/topics/i18n/translation/#the-set-language-redirect-view) (so that the URLs don't change)

longhotsummer:
DONE!

PR #170, opened 2022-06-10
Mark translation strings

nickmwangemi:


longhotsummer (approved):

longhotsummer:
LGTM. Just to note that it doesn't close #126 quite yet since the strings still need to be translated once this is merged.

PR #179, opened 2022-06-13
Add missing load i18n tags

nickmwangemi:


peachjam/templates/account/password_reset_from_key_done.html
```diff
@@ -1,4 +1,5 @@
 {% extends "account/base.html" %}
+{% load i18n %}
```
ro6ley, line 2: the i18n tags are for templates that have translation strings, this one doesn't atm. But there are some strings that need to be tagged for translation like the `Reset password` etc

nickmwangemi: Missed that. I'll update.

ro6ley (approved):

PR #187, opened 2022-06-21
Annotation of Citation Links

nickmwangemi:
This PR contains logic for annotating citation links


peachjam_api/urls.py
```diff
@@ -9,4 +9,9 @@
 
 urlpatterns = [
     path("", include(router.urls)),
+    path(
+        "documents/<int:pk>/enrichments/citation-links",
+        views.CitationLinkDetail.as_view(),
```
longhotsummer, line 14: The `CitationLinkDetail` view doesn't apply the `pk` filter to anything, so this view is no different to a normal listing view -- which is fine. Perhaps simply add a `citation-links` endpoint to the `router` above, that is a normal CRUD view? I don't see a reason right now to have two separate views.

peachjam_api/serializers.py
```diff
@@ -54,3 +54,17 @@ class Meta:
...
+
+
+class CoreDocumentSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CoreDocument
+        fields = "__all__"
+
+
+class CitationLinkSerializer(serializers.ModelSerializer):
+    document = CoreDocumentSerializer(read_only=True)
```
longhotsummer, line 66: let's make this just the document id

peachjam_api/urls.py
nickmwangemi: Done.

peachjam_api/serializers.py
nickmwangemi: Done.

longhotsummer (approved):

PR #201, opened 2022-06-23
Run tasks in separate process

ro6ley:


longhotsummer (approved):
Don't forget to `dokku scale peachjam tasks=1` after deploying

PR #208, opened 2022-06-24
Applying caching

nickmwangemi:


longhotsummer:
Wow, I was a complete idiot and didn't realise we were already doing this!

There's something odd going on, though. This doesn't seem to be having any effect. No cache-header is present in the response.

```
$ http -p Hh https://agp.africanlii.org/documents/akn/aa-au/judgment/eacj/2022/3/eng@2022-04-07/source.pdf
GET /documents/akn/aa-au/judgment/eacj/2022/3/eng@2022-04-07/source.pdf HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: agp.africanlii.org
User-Agent: HTTPie/3.2.1

HTTP/1.1 200 OK
Connection: keep-alive
Content-Disposition: inline; filename="Rugo Farm Company v Attorney General of the Republic of Burundi (Reference 14 of 2018) [2022] EACJ 3 (07 April 2022);.pdf"
Content-Language: en
Content-Type: application/pdf
Date: Fri, 24 Jun 2022 16:24:05 GMT
Referrer-Policy: same-origin
Server: nginx
Strict-Transport-Security: max-age=15724800; includeSubdomains
Transfer-Encoding: chunked
Vary: Accept-Language
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
```

longhotsummer:
Ah. Streaming responses (eg. file streaming responses) don't have caching headers applied.

longhotsummer:
Simplest is to change the views to use HttpResponse rather than FileResponse. That will both cache on the server, then, and add cache headers on the client.

nickmwangemi:
@longhotsummer I'll work on that

PR #210, opened 2022-06-27
Document modelling adjustments

### LINKED ISSUE #191: Document modelling adjustments
We need to move a bunch of models from `africanlii` into `peachjam`. Doing it now is easier than doing it later. We can either do a rather complicated dance to preserve existing data, or we can effectively create new models and re-import data.

Move these models into peachjam:

* GenericDocument
* DocumentNature
* AuthoringBody + Court (merge into one model called "Author")
* LegalInstrument
* Legislation

This will also impact the frontend in terms of the homepage, listing page facets, and search.

Let's take the opportunity to drop and re-index all ES content, and rename the index to include an app-specific identifier (eg. `africanlii_documents`).

### LINKED ISSUE #195: Merge Court and AuthoringBody into one model
Call it Author. Later we can add differentiating fields if they are necessary.

This should be done as part of #191

nickmwangemi:
closes #191
closes #195


peachjam/models/author.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
+
+
+class Author(models.Model):
+    name = models.CharField(max_length=255, null=False)
```
longhotsummer, line 5: I think both of these should be unique (independently)

nickmwangemi:
Closes #195

peachjam_search/documents.py
```diff
@@ -31,12 +37,12 @@ class SearchableDocument(Document):
...
 
     # GenericDocument, LegalInstrument
-    authoring_body = fields.KeywordField()
+    author = fields.KeywordField()
     nature = fields.KeywordField()
 
     class Index:
         # TODO: make this configurable per website
-        name = "africanlii_documents"
+        name = settings.PEACHJAM["APP_NAME"]
```
longhotsummer, line 45: The app name may have spaces, may be uppercased, etc. because it's shown to the user. So I suggest putting a setting such as

```python
PEACHJAM["ES_INDEX"] = os.environ.get("ES_INDEX", slugify(PEACHJAM["APP_NAME"]))
```

that would provide  a reasonable default and allow it to be overridden through an environment variable.

longhotsummer: And then you don't need to adjust `APP_NAME` is settings.py

longhotsummer (approved):

PR #212, opened 2022-07-01
Admin

longhotsummer:
* move admin into peachjam
* update admin forms to group related fields, put important ones first, and hide advanced ones

image


actlikewill (commented):
When saving a document from the admin, if fails if there is no work. 'CoreDocument has no work is the error'

actlikewill:
```
 if self.work_frbr_uri and (
            not self.work or self.work.frbr_uri != self.work_frbr_uri
        )
```
This logic in the coredocument save method seems to always result in false so no work gets created.

longhotsummer:
@actlikewill thanks, fixed.

actlikewill (approved):
Looks good.

PR #218, opened 2022-07-04
Unified UI document view

### LINKED ISSUE #200: Unified UI  for document display
In our current html syntax that handles rendering a document, different html is rendered depending on what document type is present (`akn, html or pdf`). This results in visual lack of consistency between the `akn`, `html`  and `pdf` documents.
PDF Document View:
pdf-document-view

AKN Document:
akn-document-view

HTML Document: 
html-document

**Proposal:**
- First and foremost create a three-column grid UI for which akn, html and pdf document views use:
- - PDF View: left column: (Tabbed content: Page previews, and search),  centre column: pdf, right column: empty
- - AKN View: Left column: (Tabbed content: TOC, and search),  centre column: akn document, right column: la-gutter
- - HTML View: Left column: Search, centre column: html document, right-column empty

- Visibility of left and right columns :
- - Left column should be toggleable in terms of visibility. For the desktop view. We will have cta button that will toggle visibility. On  mobile view it will be the same except, Left column will be overlayed above the centre column on the left (Like a side drawer). So we don't squish the content in mobile view  
- - Right column will be visible in desktop but hidden on mobile, (until start using the la-gutter properly with enrichments)

**Things to consider:**
- Thought needs to go into the color alignment used. i.e what background greys are we using
- PDF pages are canvas elements with text layers overlaid on top of it. So the width of those pages must remain the same size, otherwise, text placement will not be correct if resized. Currently, on mobile view, the page is the same size, just horizontally scrollable, we should ensure that this is preserved
- Document search on mobile: If we're going with the left column being a left-side drawer on mobile, perhaps when we search for documents, on click of the Result snippet, the drawer closes, and the normal ui interaction happens (scrolling document to corresponding text)

musangowope:
- Implement side navigation menu for tablet/mobile view
- Style pdf, html and akn document to have the same look and feel

DEMO [LATEST]:
https://www.loom.com/share/b5f06ea7462f40b3abd685382352bda8

resolves #200


longhotsummer (approved):
Awesome, let's just pull the nav-tab styling out into a global style.

Issue #190, opened 2022-06-21
Adjustments to document listing tables


longhotsummer:
* ensure all document listing tables are sorted by descending date (most recent first)
* in the author-based listing views, replace Jurisdiction column with document type
* in the document-type based listing views, replace jurisdiction with the name of the authoring body (or court)
* on the date column, add `text-nowrap` or add CSS to prevent wrapping

longhotsummer:
Re-opening because the jurisdiction replacement isn't working in production because checking the URLs isn't sufficient.

Additionally, it must show either author or document type, not jurisdiction at all.

PR #220, opened 2022-07-05
Document listing tables adjustments

nickmwangemi:


longhotsummer:
@nickmwangemi we need to do this differently, checking the request.path isn't robust enough.

longhotsummer (approved):

nickmwangemi:
Closes #190

PR #222, opened 2022-07-05
Import content from uploaded files

longhotsummer:
* add a new "upload file" input element when creating (not editing) a document
* the content is processed through a document pipeline to clean the HTML and stored as `content_html`
* the uploaded file is stored as the source file

### DIFF
--- a/peachjam/admin.py
+++ b/peachjam/admin.py
@@ -12,7 +12,7 @@ from import_export.admin import ImportMixin
 from treebeard.admin import TreeAdmin
 from treebeard.forms import movenodeform_factory
 
-from peachjam.forms import IngestorForm, RelationshipForm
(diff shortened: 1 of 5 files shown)


longhotsummer:
`docpipe` also has a more generic `soffice_convert` function which can replace the existing `docx_to_pdf` conversion in peachjam.

ro6ley (approved):
LGTM

actlikewill (approved):
Works well. I've noticed that the conversion results in line heights that are little tight. We need to ensure that docx files have line heights of above 1.5 before uploading.

longhotsummer:
Nice catch. I still need to determine what styles we strip from the HTML. That will come after we’ve had a chance to do some testing.

Greg
--
(iPhone)

PR #229, opened 2022-07-06
Fix docx file conversion to pdf

nickmwangemi:


africanlii/views/documents.py
```diff
@@ -56,5 +56,12 @@ def render_to_response(self, context, **response_kwargs):
                 file = open(f"{temp_dir.name}/{filename}", "rb")
             else:
                 file = self.object.source_file.file.open()
-            return view_attachment(file)
+            response = FileResponse(
```
longhotsummer, line 59: You're reverting your change from https://github.com/laws-africa/peachjam/pull/208 -- we can't use FileResponse. This needs to be a normal HttpResponse, very much like how `view_attachment` does it above.

africanlii/views/documents.py
```diff
@@ -56,5 +56,8 @@ def render_to_response(self, context, **response_kwargs):
                 file = open(f"{temp_dir.name}/{filename}", "rb")
             else:
                 file = self.object.source_file.file.open()
-            return view_attachment(file)
+            return HttpResponse(
+                file.read(), content_type=self.object.source_file.mimetype
```
longhotsummer, line 60: If we've converted the docx to a pdf, then the content type must be for the pdf, not the docx.

longhotsummer: Or, put another way, it should always be a PDF.

I still think line 54 should check for a non-PDF and convert it to PDF, rather than checking for docx specifically. Because `.doc` also needs to be converted.

africanlii/views/documents.py
```diff
@@ -50,11 +52,18 @@ def render_to_response(self, context, **response_kwargs):
...
+        if hasattr(self.object, "source_file") and infile:
+            if infile.name.endswith(".pdf"):
+                return HttpResponse(
+                    self.object.source_file.file.open(), content_type="application/pdf"
+                )
+            insuffix = os.path.splitext(infile.name)[1].replace(".", "")
+            primary_file, files = soffice_convert(infile, insuffix, "pdf")
+
+            response = HttpResponse(primary_file.read(), content_type="application/pdf")
+            response[
```
longhotsummer, line 65: this can go

africanlii/views/documents.py
```diff
@@ -50,11 +52,18 @@ def render_to_response(self, context, **response_kwargs):
...
     def render_to_response(self, context, **response_kwargs):
-        if hasattr(self.object, "source_file") and self.object.source_file.file:
-            if self.object.source_file.file.name.endswith(".docx"):
-                temp_dir, filename = convert_docx_to_pdf(self.object.source_file.file)
-                file = open(f"{temp_dir.name}/{filename}", "rb")
-            else:
-                file = self.object.source_file.file.open()
-            return view_attachment(file)
-        raise Http404
+        infile = self.object.source_file.file
```
longhotsummer, line 55: if `source_file` exists, it should always have a file, so you don't need to check for that.

africanlii/views/documents.py
```diff
@@ -50,11 +52,18 @@ def render_to_response(self, context, **response_kwargs):
...
+        if hasattr(self.object, "source_file") and infile:
+            if infile.name.endswith(".pdf"):
+                return HttpResponse(
```
longhotsummer, line 58: rather than copy-and-paste on the HttpResponse on this line and a few lines down, get a single file object and then have a single HttpResponse line.

PR #232, opened 2022-07-07
Add ckeditor to admin

### LINKED ISSUE #216: As an editor I'd like CKEditor to make editing HTML fields easier
* ckeditor 5
* fields: content_html
* don't use ckeditor (or make the field readonly) if the content_html_is_akn field is set.

Disallow these ckeditor features:

* images (until we can sort this out in detail)

actlikewill:
closes #216 

ckeditor

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Set up Python Environment
       uses: actions/setup-python@v2
       with:
-        python-version: "3.7"
+        python-version: "3.8"
     - uses: actions/setup-node@v2
       with:
         node-version: '14'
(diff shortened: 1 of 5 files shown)


longhotsummer:
@actlikewill if we need python 3.8 (if that's the problem here), I don't see a huge reason not to upgrade to that. 3.7 is EOL in 11 months.

actlikewill:
@longhotsummer Do we need to alter dynamically alter the widget using javascript? i.e when the user checks the content_html_akn box we hide the field or make it readonly?

longhotsummer (dismissed):
Looks good. Disabling the field as you've done it is sufficient for now.

longhotsummer:
We probably need to make `headnote_holding` on judgment also an  HTML field with CKeditor, I'll confirm shortly.

longhotsummer:
Let's go ahead and merge this and we can add more fields later if/when required.

PR #238, opened 2022-07-07
Filter out values that are not true for authors facet

### LINKED ISSUE #237: One author has a null value. Enforce that authors always have values
Currently on https://agp.africanlii.org/generic_documents/
author-null
One of author values is null in the array, causing the `ListFacets` component to break because it expects an `author` value that is true

Currently, we do have enforcements in the admin which require us to fill out the Author and code when adding an author.  I suspect the rule was not enforced during the bulk import? im guessing some documents just dont have authors?

Addressing this issue:
- Add logic to handle cases null values in vue component?
- Enforce that all authors have a value

musangowope:
add logic to filter out items in the authors iterable that are not truthy. This ensures  `ListFacets.vue` does not break as it expects truthy values
resolves #237

### DIFF
--- a/africanlii/views/generic_views.py
+++ b/africanlii/views/generic_views.py
@@ -29,7 +29,11 @@ class FilteredDocumentListView(ListView, BaseDocumentFilterForm):
             authors = []
         else:
             authors = list(
-                set(a for a in self.model.objects.values_list("author__name", flat=True) if a)
+                set(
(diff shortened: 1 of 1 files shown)


actlikewill (approved):

longhotsummer (dismissed):
this is better done in the db, but for now this is fine

PR #240, opened 2022-07-07
Change document urls to be frbr uri

### LINKED ISSUE #228: Chore: change document urls to just be the FRBR URI
Currently our document URLs are `https://example.com/documents/<frbr-uri>`. Rather, they should just be `https://example.com/<frbr-uri>`, eg. `https://example.com/akn/ke/act/...`. All FRBR URIs will start with `/akn/`, so:

* [ ] change the document view url to be `/akn/<path:frbr_uri>/`
* [ ] ensure that the view updates the `frbr_uri` kwarg to add the `/akn/` to the start of the FRBR URI before trying to work with it.
* [ ] update the search javascript to use the new url pattern

actlikewill:
closes #228


actlikewill:
Hi @ro6ley What do you think about this approach to this task? I am going with using re_path and matching urls that start with akn.

africanlii/urls.py
```diff
@@ -68,6 +68,10 @@
     ),
     path("feeds/all.xml", CoreDocumentAtomSiteNewsFeed(), name="atom_feed"),
     path("i18n/", include("django.conf.urls.i18n")),
+    re_path(
+        r"^(?P<expression_frbr_uri>akn/.*)/$",
```
ro6ley, line 72: Using re_path is fine, maybe tweak the regex closer to our frbr_uri format

longhotsummer: Does the normal django `path` segment not work? That allows basically anything, including slashes.

For reversing, remember that you'll need to strip the leading `/` from the FRBR URI.

longhotsummer: Also, just call this `frbr_uri`. We're going to need to add suspport for processing it as both an FRBR expression URI, and an FRBR work URI.

africanlii/views/documents.py
```diff
@@ -26,6 +26,11 @@ class DocumentDetailViewResolver(View):
     """Resolver view that returns detail views for documents based on their doc_type."""
 
     def dispatch(self, request, *args, **kwargs):
+        expression_frbr_uri = kwargs.get("expression_frbr_uri")
```
longhotsummer, line 29: You can make the `expression_frbr_uri` param (or actually `frbr_uri`) an actual parameter to the method, rather than having to use `kwargs`. We know it's always going to be there.

africanlii/views/documents.py
```diff
@@ -25,15 +25,14 @@ class HomePageView(TemplateView):
 class DocumentDetailViewResolver(View):
     """Resolver view that returns detail views for documents based on their doc_type."""
 
-    def dispatch(self, request, *args, **kwargs):
-        expression_frbr_uri = kwargs.get("expression_frbr_uri")
+    def dispatch(self, request, frbr_uri, *args, **kwargs):
         # add the leading slash if not present
-        if expression_frbr_uri[0] != "/":
-            kwargs["expression_frbr_uri"] = f"/{expression_frbr_uri}"
+        print(args)
```
longhotsummer, line 30:
```suggestion
```

africanlii/urls.py
actlikewill: Changed. @ro6ley I think that the regex is sufficient for this purporse since it captures everything that starts with akn/ and strips off the trailing slash. It will simply throw a 404 if no document exists with that frbr_uri.

actlikewill:
@longhotsummer We also have some more urls that are have document at the beginning eg /documents/akn/.../source.pdf, 
Should these also be changed?

longhotsummer:
Yes, let's do that for consistency's sake.

longhotsummer (approved):

longhotsummer (dismissed):

nickmwangemi (approved):

PR #244, opened 2022-07-11
Citations

longhotsummer:
Generic citation extraction framework.

### DIFF
--- a/peachjam/analysis/citations.py
+++ b/peachjam/analysis/citations.py
@@ -1,0 +1,109 @@
+from collections import defaultdict
+
+from docpipe.pipeline import Pipeline, PipelineContext, Stage
+
+
+class CitationContext(PipelineContext):
+    document = None
+    """Document being processed."""
+    html = None
+    """Parsed HTML."""
+    html_text = None
+    """Extracted HTML in text format. """
+    text = None
+    """Extracted text. """
+    use_html = False
+    """Use HTML rather than text? """
+    citation_links = []
+    """Extracted citation links that must be stored. """
+
+
+class ExtractHtml(Stage):
+    """Extract HTML from the document into html_text.
+
+    Reads: document.content_html
+    Writes: html_text
+    """
+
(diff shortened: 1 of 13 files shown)


ro6ley (approved):

actlikewill:
@longhotsummer  I'm having trouble extracting the citations from the documents. Not sure if I'm doing it correctly

longhotsummer:
> @longhotsummer I'm having trouble extracting the citations from the documents. Not sure if I'm doing it correctly

in HTML documents, they'll show up as links.

image

Here's an example you can use: [Test document for extracting citations.docx](https://github.com/laws-africa/peachjam/files/9120001/Test.document.for.extracting.citations.docx)


PDF document citation links aren't yet displayed.

peachjam/analysis/citations.py
```diff
@@ -0,0 +1,59 @@
...
+            return self.extract_citations_from_html(document)
+        else:
+            # markup the source file by extracting the text and assuming we'll render it as a PDF
+            return self.extract_citations_from_source_file(document)
+
+    def extract_citations_from_html(self, document):
+        html = lxml.html.fromstring(document.content_html)
+        for matcher in self.matchers:
+            matcher().markup_html_matches(document.expression_uri(), html)
+        document.content_html = lxml.html.tostring(html, encoding="unicode")
```
ro6ley, line 30: What's the reason behind not storing these citation links like when they are extracted from the source document in `extract_citations_from_source_file()` below?

longhotsummer: HTML citations are marked up in the html itself, as `a` elements, and therefore HTML extractions don't need the CitationLinks. We can't do that for PDF, so they're stored as CitationLinks.

longhotsummer:
> @longhotsummer We may need to add the poppler-utils dependency to the docker config

Good catch, added.

actlikewill (approved):
Looks great and works well.

PR #253, opened 2022-07-13
Peach jam settings object

longhotsummer:
Makes it possible to limit choices for language and jurisdictions without having to mess with countries_plus or languages_plus


peachjam/admin.py
```diff
@@ -34,14 +34,20 @@
...
     Relationship,
     SourceFile,
     Taxonomy,
+    pj_settings,
 )
 from peachjam.resources import GenericDocumentResource, JudgmentResource
 
 
+class PeachJamSettingsAdmin(admin.ModelAdmin):
+    filter_horizontal = ("document_languages",)
```
actlikewill, line 48:
```suggestion
    filter_horizontal = ("document_languages", "document_jurisdictions",)
```
We can use this for both languages and jurisdictions.

PR #257, opened 2022-07-14
Footer tweaks

nickmwangemi:


nickmwangemi:
Adjust footer from...
image


To...
image

PR #259, opened 2022-07-14
Improved footer logo responsive wrapping

musangowope:
- Reduce size of srt logo
- Add styling to improve footer logo responsive wrapping when scaling down to mobile

DEMO:
https://www.loom.com/share/fdfed80d629b40319ec3cb997922d646


longhotsummer (approved):
I've found `img-responsive` is a useful bootstrap attribute, too. So you rely on the container to determine the size and the image fits within it.

PR #260, opened 2022-07-14
Move views into peachjam from africanlii

nickmwangemi:


longhotsummer (dismissed):
Looks good, just sanity check that you're not breaking the i18next-parser requirement.

longhotsummer (approved):

Issue #235, opened 2022-07-07
ensure editors can capture new content in admin area


longhotsummer:

longhotsummer:
Done

PR #266, opened 2022-07-18
Formalise creation of FRBR URIs

longhotsummer:
To make it easier for editors, we need to enforce (and help generate) work FRBR URIs. Some notes are [in this google doc](https://docs.google.com/document/d/1SIZLdiRkdm1ZkKgi4nLor_bdZOkZLSd39QkccMq6fE0/edit#heading=h.92keeuyypx9j).

The idea behind this PR is that in the general case, we help editors to fill in the components of an FRBR URI and ensure it's valid. In the specific case, like for legislation or a judgment, some or all of the FRBR URI may be set by the system, and we don't want the user to make changes.

* make the components of the FRBR URI actual fields (doctype, subtype, actor, date, number)
* always generate and validate an FRBR URI when saving
* allow document models to provide different lists of possible options for the "doctype" field, to help ensure that certain document models are always used for certain content

Outstanding:

* [ ] populate the necessary fields when importing from spreadsheets or indigo
* [ ] auto-generate the `frbr_uri_number` field in the admin area based on the slugified version of the title field


longhotsummer:
https://github.com/laws-africa/cobalt/issues/71 means we have to do some hoop jumping for judgment FRBR URI creation.

longhotsummer:
@actlikewill please can you make the changes to populate the `frbr_uri_*` fields appropriately when importing with django-import-export, and with the indigo ingestor

longhotsummer:
Closes #235

actlikewill:
We need to convert the frbr uri values in the spreadsheet to strings especially the frbr_uri_number column. It gets imported as a float value with a decimal point, which messes up the frbr_uri. It can be handled within the code but I think it would be better if we did it in the spreadsheet.

peachjam/adapters/adapters.py
```diff
@@ -74,30 +75,24 @@ def update_document(self, url):
             LegalInstrument,
             Legislation,
             Locality,
-            Work,
         )
 
         logger.info(f"Updating document ... {url}")
 
         document = self.client_get(f"{url}.json").json()
-        frbr_uri = document["frbr_uri"]
```
longhotsummer, line 83: keep this, and parse it with `FrbrUri.parse` which will then give you all the constituent fields, including `author`, which we don't expose in the Indigo API.

peachjam/resources.py
```diff
@@ -67,17 +66,7 @@ def before_import_row(self, row, **kwargs):
...
         row["jurisdiction"] = frbr_uri.country
         row["locality"] = frbr_uri.locality
-
-        work, _ = Work.objects.update_or_create(
-            frbr_uri=frbr_uri, defaults={"title": row["title"]}
-        )
-        row["work"] = work.id
-        row["author"] = row["author_obj"]["code"]
-
-        Author.objects.get_or_create(
```
longhotsummer, line 77: let's keep this, and do the same thing for the Indigo adapter.

peachjam/resources.py
```diff
@@ -67,17 +66,7 @@ def before_import_row(self, row, **kwargs):
...
-            code=row["author_obj"]["code"],
-            name=row["author_obj"]["name"],
-        )
+        row["frbr_uri_number"] = str(row["frbr_uri_number"])
```
longhotsummer, line 69: rely on the parsed frbr_uri's `.number` instead

actlikewill:
@longhotsummer I have made the requested changes. There is still an issue with the ingestor that I want to fix. Its not updating correctly. But I can do that on a separate pr.

peachjam/admin.py
```diff
@@ -362,6 +400,13 @@ class IngestorAdmin(admin.ModelAdmin):
     inlines = [IngestorSettingInline]
     readonly_fields = ("last_refreshed_at",)
     form = IngestorForm
+    actions = ["reset_ingestor_refresh_date"]
+
+    def reset_ingestor_refresh_date(self, request, queryset):
```
longhotsummer, line 405: @actlikewill I think it would be useful if this also queued up the background task  to refresh this ingestor.

PR #271, opened 2022-07-19
Minor changes:

nickmwangemi:
## Minor Changes:
- Change "Author" to Regional Body" in homepage and in filter facets
image
image

- Order recent document list by document date
image

- Show Regional Body, rather than jurisdiction in document lists that show jurisdiction
image

- When showing document type in listing views, needs to be user-friendly document type
image

## Bug fix:
- Selecting multiple years in the document facet view now shows documents with all years
image

### DIFF
--- a/africanlii/forms.py
+++ b/africanlii/forms.py
@@ -15,7 +15,6 @@ class BaseDocumentFilterForm(forms.Form):
     def __init__(self, data, *args, **kwargs):
         self.params = QueryDict(mutable=True)
         self.params.update(data)
-        print(data)
 
         super().__init__(self.params, *args, **kwargs)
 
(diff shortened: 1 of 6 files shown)


africanlii/forms.py
```diff
@@ -11,23 +12,29 @@ class BaseDocumentFilterForm(forms.Form):
...
         if year and exclude != "year":
-            queryset = queryset.filter(date__year=year)
+            queryset = queryset.filter(date__year__in=year)
 
         if alphabet and exclude != "alphabet":
             queryset = queryset.filter(title__istartswith=alphabet)
 
         if author and exclude != "author":
-            queryset = queryset.filter(author__name__iexact=author)
+            queryset = queryset.filter(author__name__in=author)
```
longhotsummer, line 35: @actlikewill can you pls double check that this is doing the right thing?

I also don't know if there's a way to let the form fields give us multiple choices for those fields (just curious -- this mechanism works too)

longhotsummer (approved):

PR #272, opened 2022-07-19
Move read-only attribute to correct element

musangowope:
Data attributes (including readonly) were not on the correct element. It was supposed to be on `data-component="RelationshipEnrichments"`

### DIFF
--- a/peachjam/templates/peachjam/_document_content.html
+++ b/peachjam/templates/peachjam/_document_content.html
@@ -118,13 +118,15 @@
   <div class="document-content__inner">
     <div class="navigation" id="navigation-column">
     </div>
-    <div class="content-and-enrichments" data-component="RelationshipEnrichments">
+    <div class="content-and-enrichments"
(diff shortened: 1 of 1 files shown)


musangowope:
@longhotsummer I noticed in`RelationshipEnrichment/index.ts` even though `getButton` is used to place the floating `Add relationship` button in the `gutter`, no click event is bound to that button to run `addEnrichment`, which trigger showing the modal
```
addEnrichment (target: IRangeTarget): void {
    this.listComponent.creating = {
      id: null,
      predicate_id: null,
      predicate: {},

      subject_work_id: this.workId,
      subject_work: {
        frbr_uri: this.workFrbrUri
      },
      subject_target_id: target.anchor_id,
      subject_documents: [],

      object_work: {},
      object_target_id: null,
      object_documents: []
    };
```
Hence nothing happens on click of `Add relationship`, is this the intended effect for now?

longhotsummer (approved):

PR #280, opened 2022-07-21
Ingestor update fix

actlikewill:
- The issue here was that the frbr uris coming from indigo and the ones that we are generating on peachjam do not match.
- The adapter tries to create a new document from the frbr_uri provided by indigo since it can't find a match on peachjam
- This results in db integrity errors when performing the update from the ingestor.
- By reusing the frbr_uri generation methods on the core document model, we can now get the correct value and perform the update.

### DIFF
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -72,6 +72,7 @@ class IndigoAdapter(Adapter):
 
         from peachjam.models import (
             Author,
+            CoreDocument,
             DocumentNature,
             GenericDocument,
             LegalInstrument,
@@ -84,7 +85,6 @@ class IndigoAdapter(Adapter):
         document = self.client_get(f"{url}.json").json()
         frbr_uri = FrbrUri.parse(document["frbr_uri"])
         title = document["title"]
-        expression_frbr_uri = document["expression_frbr_uri"]
         toc_json = self.get_toc_json(url)
         content_html = self.client_get(url + ".html").text
         jurisdiction = Country.objects.get(iso__iexact=document["country"])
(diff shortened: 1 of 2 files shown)


peachjam/adapters/adapters.py
```diff
@@ -95,27 +95,35 @@ def update_document(self, url):
...
             "toc_json": toc_json,
             "content_html": content_html,
+        }
+
+        frbr_uri_data = {
+            "locality": locality,
+            "jurisdiction": jurisdiction,
             "frbr_uri_subtype": frbr_uri.subtype,
             "frbr_uri_number": frbr_uri.number,
             "frbr_uri_doctype": frbr_uri.doctype,
```
longhotsummer, line 112: Also pull out `frbr_uri_date`

peachjam/adapters/adapters.py
```diff
@@ -95,27 +95,35 @@ def update_document(self, url):
...
         if frbr_uri.actor:
-            field_data["frbr_uri_actor"] = frbr_uri.actor
+            frbr_uri_data["frbr_uri_actor"] = frbr_uri.actor
             field_data["author"] = Author.objects.update_or_create(code=frbr_uri.actor)[
                 0
             ]
 
+        doc = CoreDocument(**frbr_uri_data)
+        doc.work_frbr_uri = doc.generate_work_frbr_uri()
+        expression_frbr_uri = doc.generate_expression_frbr_uri()
```
longhotsummer, line 125: Add two asserts here: the generated work and expression FRBR URIs and the ones provided by indigo must match

peachjam/resources.py
```diff
@@ -72,7 +79,6 @@ def before_import_row(self, row, **kwargs):
         row["frbr_uri_date"] = frbr_uri.date
 
         if frbr_uri.actor:
-            Author.objects.update_or_create(code=frbr_uri.actor)
```
actlikewill, line 75: @longhotsummer  I have removed this because creating the author with only the code without the name results in unpredictable errors since we are using the case insensitive lookup. This means that we have to ensure that the authors already exist on peachjam to avoid errors. I am open to suggestions on how to handle this.

peachjam/adapters/adapters.py
actlikewill: I have added the check below.

peachjam/resources.py
longhotsummer: that's fine, as long as we get a reasonable error telling us what is missing so that we can create it.

longhotsummer (approved):

PR #282, opened 2022-07-22
Fix compiling scss error which is block deploys

musangowope:
During deploy, `python manage.py compilescss` scans all through all django templates for `scss` link references and finds the corresponding `scss file` in static to compile. The problem is that it is not intelligent enough to decipher what we had. (Not able to decipher `APP_STYLES`)
```
<link rel="stylesheet" href="{% sass_src APP_STYLES  %}" type="text/css">
```
Hence deploys are blocked

**Solution:**
We have to put an actual value. I took the alternative approach to solve this problem which was extending `main.html` and override `block head-css` in each sub app of `peachjam` (`liiweb, africanlii, lawlibrary`)

resolves https://github.com/laws-africa/peachjam/issues/283

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -161,3 +161,9 @@ cython_debug/
 staticfiles
 .DS_Store
 .sass-cache
+
+#Ignore app static build styles
+peachjam/static/stylesheets/peachjam.css
+africanlii/static/stylesheets/africanlii.css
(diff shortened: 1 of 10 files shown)


longhotsummer (approved):
Ugh :facepalm:

PR #288, opened 2022-07-26
Home + Judgement Listing + Detail page

nickmwangemi:
- Created Law library Homepage template
- Created Liiweb list page template
- Created Judgement list page template
- Created Court Detail page template (Judges by court)

DEMO:
https://www.loom.com/share/0d195ae8737f4e8d9ae1cc5909315bb9


musangowope:
@nickmwangemi please could you fix the tests when you get the chance

lawlibrary/templates/peachjam/header.html
```diff
@@ -10,10 +10,14 @@
...
 {% block nav-items %}
   <li class="nav-item">
-    <a class="nav-link {% ifequal request.path '/judgements/'%} active {% endifequal%}" href="#">{% trans 'Judgements' %}</a>
+    <a class="nav-link {% ifequal request.path '/judgements/'%} active {% endifequal%}" href="{% url 'judgment_list' %}">
+      {% trans 'Judgements' %}
+    </a>
   </li>
   <li class="nav-item">
-    <a class="nav-link {% ifequal request.path '/legislation/'%} active {% endifequal%}" href="#">{% trans 'Legislation' %}</a>
+    <a class="nav-link {% ifequal request.path '/about/'%} active {% endifequal%}" href="#">
```
longhotsummer, line 18: Incorrect href

lawlibrary/urls.py
```diff
@@ -4,5 +4,8 @@
 
 urlpatterns = [
     path("", views.HomePageView.as_view(), name="home_page"),
+    path("judgments/", views.JudgmentListView.as_view(), name="judgment_list"),
+    path("judgments/<str:code>/", views.CourtDetailView.as_view(), name="court_detail"),
+    path("judgments/<str:code>/<int:year>/", views.YearView.as_view(), name="year"),
```
longhotsummer, line 9: Not crucial, but I suggest naming the route and the view class to as "CourtYearView" or similar. Year is rather unspecific.

lawlibrary/views/court.py
```diff
@@ -0,0 +1,66 @@
...
+    paginate_by = 20
+    model = Judgment
+    template_name = "lawlibrary/court_detail.html"
+
+    def get_base_queryset(self):
+        if "year" in self.kwargs:
+            return self.model.objects.filter(
+                author=self.author, date__year=self.kwargs["year"]
+            )
+        return self.model.objects.filter(author=self.author)
```
longhotsummer, line 21: You're repeating yourself here. Rather, get the basic filtered queryset based on the author, and then do the year filtering on top of that if a year is present:

```python
qs = self.model.objects.filter(author=self.author)
if "year" in self.kwargs:
    qs = qs.filter(date__year=self.kwargs["year"])
```

lawlibrary/views/court.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def get_queryset(self):
+        self.author = get_object_or_404(Author, code=self.kwargs["code"])
+        return super().get_queryset()
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        years = list(
```
longhotsummer, line 30: I know we use this pattern elsewhere, but as our DB grows we'll land up loading 1000s of years into memory just to make them distinct. Much more efficient to get the db to do it:

```python
years = self.model.objects.filter(author=self.author).order_by().values_list("date__year", flat=True).distinct()
```

This is a pattern we can re-use elsewhere. The `order_by()` is important to turn off the default ordering, which would be doing the wrong thing.

Here's an example of what the SQL is, just to confirm:

```
>>> str(Judgment.objects.filter(author__name="foo").order_by().values_list("date__year", flat=True).distinct().query)
'SELECT DISTINCT EXTRACT(\'year\' FROM "peachjam_coredocument"."date") FROM "peachjam_judgment"
INNER JOIN "peachjam_author" ON ("peachjam_judgment"."author_id" = "peachjam_author"."id")
INNER JOIN "peachjam_coredocument" ON ("peachjam_judgment"."coredocument_ptr_id" = 
"peachjam_coredocument"."id") WHERE "peachjam_author"."name" = foo'
```

lawlibrary/templates/lawlibrary/court_detail.html
```diff
@@ -0,0 +1,46 @@
+{% extends 'peachjam/layouts/document_list.html' %}
+{% load i18n %}
+
+{% block title %}{% trans 'Judgement List' %}{% endblock %}
+
+{% block page-title %}
+    <h1 class="m-0">{% trans 'Judgements List' %}</h1>
```
longhotsummer, line 7: Rather use the court (author) name; if there's a year, append it.

eg. "Supreme Court" and "Supreme Court - 2021"

lawlibrary/templates/lawlibrary/court_detail.html
```diff
@@ -0,0 +1,46 @@
+{% extends 'peachjam/layouts/document_list.html' %}
+{% load i18n %}
+
+{% block title %}{% trans 'Judgement List' %}{% endblock %}
```
longhotsummer, line 4: As below

lawlibrary/templates/lawlibrary/court_detail.html
```diff
@@ -0,0 +1,46 @@
...
+              {{ y }}
+            </a>
+          {% endif %}
+        {% endfor %}
+      </div>
+    {% endif %}
+  </div>
+
+  <select class="form-select d-block d-lg-none" data-component="NavigationSelect">
+    <option value="{% url 'court_detail' court.code %}">Any</option>
```
longhotsummer, line 35:
```suggestion
    <option value="{% url 'court_detail' court.code %}">All years</option>
```

lawlibrary/templates/lawlibrary/court_detail.html
```diff
@@ -0,0 +1,46 @@
...
+
+{% block content %}
+  <div class="d-none d-lg-flex align-items-center">
+    {% if years %}
+      <i class="bi bi-calendar"></i>
+      <div class="d-flex">
+        <a class="p-2"
+           href="{% url 'court_detail' court.code %}"
+        >
+          Any
```
longhotsummer, line 18:
```suggestion
          All years
```

liiweb/templates/liiweb/layouts/list.html
```diff
@@ -0,0 +1,39 @@
+{% extends "peachjam/layouts/main.html" %}
+
+{% block page-content %}
+  <section class="pt-5 pb-5">
+    <div class="container">
+      {% block grouped-links %}
+        {% if groups %}
+          {% for group in groups %}
+            <h4 class="heading-underlined">
```
longhotsummer, line 9: Two thoughts on this:

1. this template is only used by one other template -- it seems premature to make it re-usable (especially since line 21 refers to courts)
2. please put the court class groupings into three columns: so "Magistrate Court" is one column, "Supreme Court" the next etc. eg. `col-md-3` should work, even if it wraps between rows.

longhotsummer (changes requested):
This is looking super.

@nickmwangemi There are some comments about the backend

@musangowope There are some comments about the frontend and templates

I'd love to get this merged by the end of today, I'm free after lunch to work through these smaller details and get this done!

lawlibrary/urls.py
musangowope: done

lawlibrary/templates/peachjam/header.html
musangowope: removed about About link since we dont actually have the page

lawlibrary/templates/lawlibrary/court_detail.html
musangowope: done

lawlibrary/templates/lawlibrary/court_detail.html
musangowope: done

lawlibrary/templates/lawlibrary/court_detail.html
```diff
@@ -1,10 +1,26 @@
 {% extends 'peachjam/layouts/document_list.html' %}
 {% load i18n %}
 
-{% block title %}{% trans 'Judgement List' %}{% endblock %}
+{% block title %}
+  {% if year %}
+    {% blocktrans with court_name=court.name year=year%}
+      {{ court_name }} - {{ year }}
+    {% endblocktrans %}
```
longhotsummer, line 8: No need to translate this at all, it's just variables.

longhotsummer: But it won't hurt, you can leave it in.

liiweb/templates/liiweb/layouts/list.html
musangowope: refactored. Put the list.html logic in judgement_list.html (obviously without the block logic) and deleted list.html

peachjam/settings.py
```diff
@@ -65,7 +65,7 @@
     "sass_processor",
     "import_export",
     "treebeard",
-    "background_task",
+    # "background_task",
```
longhotsummer, line 68: I think this needs to be added back?

app.json
```diff
@@ -1,7 +1,7 @@
 {
   "scripts": {
     "dokku": {
-      "predeploy": "python manage.py compilescss && python manage.py collectstatic --noinput -i *.scss",
+      "predeploy": "python manage.py migrate && python manage.py compilescss && python manage.py collectstatic --noinput -i *.scss",
```
longhotsummer, line 4: We run migrations in postdeploy in all our other apps. I think we should aim for consistency unless there's a good argument to have this here?

peachjam/settings.py
nickmwangemi: Yes, I've added it back. It was part of the deploy troubleshooting.

app.json
nickmwangemi: I've reverted. It was part of the deploy troubleshooting.

longhotsummer (approved):
Nice!

Issue #293, opened 2022-08-01
ALL listings page ordered by date


musangowope:
https://agp.africanlii.org/authors/8/
https://agp.africanlii.org/authors/7/

longhotsummer:
Please also sanity check all the other listing pages (NOT the homepage "recent" view, though) -- descending date order should be the default for the document listing tables.

PR #298, opened 2022-08-01
Order authors and legislation views by date

musangowope:
Sanity checked:
- Judgements
- Generic documents
- Legal instruments
- legislation

DEMO:
https://www.loom.com/share/6cab1b0b57f44ad488b980ea33feea1e

resolves https://github.com/laws-africa/peachjam/issues/293

### DIFF
--- a/peachjam/views/legislation.py
+++ b/peachjam/views/legislation.py
@@ -12,10 +12,6 @@ class LegislationListView(FilteredDocumentListView):
     context_object_name = "documents"
     paginate_by = 20
 
-    def get_queryset(self):
-        queryset = super(LegislationListView, self).get_queryset()
-        return queryset.order_by("-date")
-
 
 @registry.register_doc_type("legislation")
 class LegislationDetailView(BaseDocumentDetailView):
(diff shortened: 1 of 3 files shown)


peachjam/views/legislation.py
```diff
@@ -12,6 +12,10 @@ class LegislationListView(FilteredDocumentListView):
     context_object_name = "documents"
     paginate_by = 20
 
+    def get_queryset(self):
+        queryset = super(LegislationListView, self).get_queryset()
```
longhotsummer, line 16: Push this down into FilteredDocumentListView, so that all list views get it. You can also simplify it a bit:

```python
def get_queryset(self):
    return super().get_queryset().order_by("-date")
```

longhotsummer: The `super(LegislationListView, self)` is the old way of calling a superclass's method, which is no longer required and is simplified to just `super()`.

musangowope: done

longhotsummer (approved):

PR #301, opened 2022-08-03
When importing word, strip font and line-height

longhotsummer:
Also reset line-height to normal for the document text. Otherwise it uses bootstrap's adjusted line height.

Before:
image

After:
image


actlikewill (approved):
Looks much better.

PR #307, opened 2022-08-10
Add notices

nickmwangemi:
image


peachjam/views/legislation.py
```diff
@@ -17,3 +23,46 @@ class LegislationListView(FilteredDocumentListView):
...
+            )
+
+        # TODO: Out-of-date legislation
+
+        # TODO: Up-to-date legislation
+
+        return notices
+
+    def get_repeal_info(self):
+        repeal_info = json.dumps(self.object.metadata_json.get("repeal", None))
```
longhotsummer, line 59: This looks odd to me -- I thought the data in `metadata_json` should already be decoded into dictionaries, since it's stored as JSONField?

peachjam/views/generic_views.py
```diff
@@ -138,3 +138,8 @@ def get_context_data(self, **kwargs):
             context["display_type"] = None
 
         return context
+
+    def get_notices(self, notices=None):
```
longhotsummer, line 142: I don't think the `notices` param is necessary?

peachjam/views/legislation.py
```diff
@@ -17,3 +23,46 @@ class LegislationListView(FilteredDocumentListView):
 class LegislationDetailView(BaseDocumentDetailView):
     model = Legislation
     template_name = "peachjam/legislation_detail.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["notices"] = self.get_notices()
```
longhotsummer, line 29: This can go into the superclass, since that's where the default implementation of `get_notices` is.

musangowope:
Perhaps the anchor text should the same color as notice text?
image

peachjam/views/legislation.py
```diff
@@ -17,3 +22,77 @@ class LegislationListView(FilteredDocumentListView):
...
+                            repeal["date"],
+                            repeal["repealing_uri"],
+                            repeal["repealing_title"],
+                        )
+                    ),
+                }
+            )
+
+            msg = (
+                "This is the latest version of this {} as it was when it was repealed."
```
longhotsummer, line 46: This might not be true. If a work is repealed, then all versions of the work are marked repealed.

Instead of calculating this message here, it should be calculated when you're working out if this is the latest expression or not.

longhotsummer: Look at how the messages are determined for openbylaws.org.za: https://github.com/laws-africa/openbylaws.org.za/blob/a9138e619ae5d21df182ff09e621a773c0ab9d76/_includes/work_notices.html#L16

peachjam/views/legislation.py
```diff
@@ -17,3 +22,77 @@ class LegislationListView(FilteredDocumentListView):
...
+                                ).date()
+                                - timedelta(days=1),
+                                latest_expression["expressions"][0][
+                                    "expression_frbr_uri"
+                                ],
+                            )
+                        ),
+                    }
+                )
+        self.add_notices_to_messages(notices)
```
longhotsummer, line 87: The notices shouldn't be added to messages. These are not general system messages that need to be shown the next time the page loads and never again. We're only using the message system's concept of warning vs danger vs info.

peachjam/views/legislation.py
```diff
@@ -17,3 +22,77 @@ class LegislationListView(FilteredDocumentListView):
...
+                    ),
+                }
+            )
+
+        # fetch the points in time
+        points_in_time = self.get_points_in_time()
+        latest_expression = points_in_time[-1]
+
+        for idx, point_in_time in enumerate(points_in_time):
+            if point_in_time != latest_expression:
```
longhotsummer, line 64: This case seems to be missing: For the most up-to-date version of amended legislation, the notice should say "This is the latest version of this (friendly type)".

longhotsummer (changes requested):
@nickmwangemi pls refer back to the original issue and ensure all the required cases are correctly handled. The openbylaws.org.za code referenced there should be used as a guide.

peachjam/views/legislation.py
```diff
@@ -17,3 +22,77 @@ class LegislationListView(FilteredDocumentListView):
...
+                    "This is the version of this {} as it was from {} to {}. "
+                    "<a href={}>Read the version currently in force</a>"
+                )
+                notices.append(
+                    {
+                        "type": messages.WARNING,
+                        "html": format_html(
+                            msg.format(
+                                self.object.metadata_json["type_name"],
+                                point_in_time["expressions"][0]["expression_date"],
```
longhotsummer, line 75: double check the zero -- I think that should be `points_in_time[idx-1]`

peachjam/views/legislation.py
```diff
@@ -17,3 +22,77 @@ class LegislationListView(FilteredDocumentListView):
...
+                        )
+                    ),
+                }
+            )
+
+        # fetch the points in time
+        points_in_time = self.get_points_in_time()
+        latest_expression = points_in_time[-1]
+
+        for idx, point_in_time in enumerate(points_in_time):
```
longhotsummer, line 63: Rather than iterating over all the points in time, you want to find the index of the current point in time in the list. That tells you the start and end dates for that expression (the end date will be "never" if this is the latest point in time).

longhotsummer:
> Perhaps the anchor text should the same color as notice text? image
> 
> In this case. Both texts will be dark red

@musangowope good point -- this should be handled at the top-level SCSS where we can set link colours for the various alerts. I don't think it's a blocker for this feature. If you can make the change and open a PR into Nick's branch, or open a new issue, then Nick can focus on the business logic aspect and you can manage the styling separately.

longhotsummer (approved):

PR #311, opened 2022-08-12
Handle work uri and redirect to document

### LINKED ISSUE #296: When following a citation link, redirect me to the most recent version of a work
The work-based citation links currently don't work: https://agp.africanlii.org/akn/aa-au/statement/resolution/achpr/2000/45/

We need to update the document view to recognise work FRBR URIs (vs. expression FRBR URIs) and find the most recent version of a work and redirect to that document.

longhotsummer:
Also strip the / at the end of the FRBR URI, since it's not actually a part of the FRBR URI

Closes #296


actlikewill (approved):
Works well.

PR #322, opened 2022-08-16
Page based indexing for pdfs

### LINKED ISSUE #300: Add support for indexing PDFs and showing page-based search results
Currently we don't index content for PDF documents at all.

The goal is to index each page separately, so that we can point users directly at the page that matches. This is similar to what we do in gazettes.africa:

image

## Indexing

We need to index the gazette page content in a different field to the general document content, since they're indexed and searched differently.

Add a `pages` nested field to the core document index:

```python
    pages = fields.NestedField(properties={
        'page_num': fields.IntegerField(),
        'body': fields.TextField(analyzer='standard', fields={'exact': Text()}),
    })
```

When indexing, extract the pages and index them separately. Something like:

```python
    def prepare_pages(self, instance):
        """ Extract pages from the PDF.
        """
        # use docpipe.pdf.pdf_to_text to get the raw text, you'll need to download the PDF locally first into a temp file
        text = run_pdf_to_text(instance)
        if not text:
            raise ValueError(f"Couldn't get any text to index for search from {instance}")
        # split on ^L -- form feed / page markers
        page_texts = text.split("\x0C")
        pages = []
        for i, page in enumerate(page_texts):
            i = i + 1
            page = page.strip()
            if page:
                pages.append({
                    'page_num': i,
                    'body': page,
                })

        return pages
```

## Searching

The search view will be updated to add a query backend to add the page search to the query. I don't know if there's an easier way, but here's how we've done it before:

```python
class PageQueryBackend(BaseSearchQueryBackend):
    """ Does a nested page search, and include highlights.
    """
    @classmethod
    def construct_search(cls, request, view, search_backend):
        queries = []
        for search_term in search_backend.get_search_query_params(request):
            queries.append(Q(
                'nested',
                path='pages',
                query=Q('bool', must=[
                    SimpleQueryString(query=search_term, default_operator="and", quote_field_suffix=".exact",
                                      fields=['pages.body'])
                ], should=[
                    MatchPhrase(pages__body={'query': search_term, 'slop': 2}),
                ]),
                inner_hits={
                    '_source': ["pages.page_num"],
                    'highlight': {
                        'fields': {'pages.body': {}},
                        'pre_tags': ["<mark>"],
                        'post_tags': ["</mark>"],
                        'fragment_size': 80,
                        'number_of_fragments': 2,
                    },
                }
            ))
        return queries


class CustomSearchBackend(BaseSearchFilterBackend):
    query_backends = [PageQueryBackend]
```

The result serializer must be updated to extract useful information from the nested page field, if any:

```python
    pages = serializers.SerializerMethodField()

    def get_pages(self, obj):
        """ Serialize nested page hits and highlights.
        """
        pages = []
        if hasattr(obj.meta, 'inner_hits'):
            for page in obj.meta.inner_hits.pages.hits.hits:
                info = page._source.to_dict()
                info['highlight'] = page.highlight.to_dict()
                pages.append(info)
        return pages
```

# Frontend

The frontend must be updated to show the page matches, highlights, and page-based links.

The document detail page should support a `page=2` query parameter and jump to that page when first loading the PDF.

actlikewill:
DEMO:
https://www.loom.com/share/bff6768dcf864ba7828d352ac9324c8a

closes #300

### DIFF
--- a/peachjam_search/views.py
+++ b/peachjam_search/views.py
@@ -2,6 +2,7 @@ from django.utils.decorators import method_decorator
 from django.views.decorators.cache import cache_page
 from django.views.generic import TemplateView
 from django_elasticsearch_dsl_drf.filter_backends import (
+    BaseSearchFilterBackend,
     CompoundSearchFilterBackend,
     DefaultOrderingFilterBackend,
     FacetedFilterSearchFilterBackend,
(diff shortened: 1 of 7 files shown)


peachjam_search/views.py
```diff
@@ -19,6 +24,54 @@
...
+                )
+            )
+        return queries
+
+
+class CustomSearchBackend(BaseSearchFilterBackend):
+    query_backends = [PageQueryBackend]
+
+    def get_query_backends(self, request, view):
+        return self.query_backends
```
actlikewill, line 72: @ro6ley Hi I am implementing this page-based indexing but for some reason I am not getting any results with this custom backend. Not sure where the problem is. Maybe you could take a look?

longhotsummer (approved):
This looks ready to go?

PR #327, opened 2022-08-18
Legislation table vue and styling

musangowope:
- Created `LegislationTable/index.vue` to display Legislation data
- Implemented search functionality in component
- Implemented sorting (ascending or descending) by column heading
- Implemented filtering using `SecondaryFacets.vue` 
- Added mobile support. SecondaryFacets are displayed Offcanvas element on mobile view

DEMO:
https://www.loom.com/share/a7de2bde771742c5b044ba7270a98630

### DIFF
--- a/peachjam/js/components/LegislationTable/index.vue
+++ b/peachjam/js/components/LegislationTable/index.vue
@@ -217,7 +217,8 @@ export default {
           title: 'Repealed',
           name: 'repealed',
           type: 'boolean',
-          value: true
+          value: true,
+          count: this.filteredData.filter(item => item.repealed).length
         }
       ];
     },
--- a/peachjam/js/components/LegislationTable/index.vue
+++ b/peachjam/js/components/LegislationTable/index.vue
@@ -197,6 +197,12 @@ export default {
           value: key
         }));
       };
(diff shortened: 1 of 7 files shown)


peachjam/js/components/LegislationTable/index.vue
```diff
@@ -0,0 +1,319 @@
...
+      <div class="card legislation-table">
+        <div class="card-header">
+          <div class="input-group">
+            <button
+              class="btn btn-outline-secondary dropdown-toggle"
+              type="button"
+              data-bs-toggle="dropdown"
+              aria-expanded="false"
+            >
+              Search by {{ filters[selectedFilterKey].label }}
```
longhotsummer, line 21: Two thoughts:

1. Avoid the word "search", which can imply lots of smarts; rather use "filter", which is clearer that it's going to look for exactly what you type without any smarts
2. can we reduce this to just one input box, and the same filter is applied to both title fields, and if any match then the row is retained. The benefit of this is the user doesn't have to think: typing "Children" and "5 of 2020" both do what the user expects.

musangowope:
All functionality implemented. Next step: sanity with extensive set of Legislation data

PR #335, opened 2022-08-22
Make page-based searching work

longhotsummer:
* create a simple compound filter backend that will correctly join the field-based search with the nested page query search into a single "should" statement
* don't return large source fields which are not shown in the search results (and remove from serializer)
* boost title field
* use SimpleQueryString search which handles quoted searches for us

Produces a search like this:

```
"query": {
    "bool": {
      "should": [
        {
          "simple_query_string": {
            "fields": [
              "title^2",
              "author",
              "citation",
              "judges",
              "content"
            ],
            "query": "africa"
          }
        },
        {
          "nested": {
            "inner_hits": {
              "_source": [
                "pages.page_num"
              ],
              "highlight": {
                "fields": {
                  "pages.body": {}
                },
                "fragment_size": 80,
                "number_of_fragments": 2,
                "post_tags": [
                  "</mark>"
                ],
                "pre_tags": [
                  "<mark>"
                ]
              }
            },
            "path": "pages",
            "query": {
              "bool": {
                "must": [
                  {
                    "simple_query_string": {
                      "default_operator": "and",
                      "fields": [
                        "pages.body"
                      ],
                      "query": "africa",
                      "quote_field_suffix": ".exact"
                    }
                  }
                ],
                "should": [
                  {
                    "match_phrase": {
                      "pages.body": {
                        "query": "africa",
                        "slop": 2
                      }
                    }
                  }
                ]
              }
            }
          }
```

### DIFF
--- a/peachjam_search/views.py
+++ b/peachjam_search/views.py
@@ -2,7 +2,6 @@ from django.utils.decorators import method_decorator
 from django.views.decorators.cache import cache_page
 from django.views.generic import TemplateView
 from django_elasticsearch_dsl_drf.filter_backends import (
-    BaseSearchFilterBackend,
     CompoundSearchFilterBackend,
     DefaultOrderingFilterBackend,
     FacetedFilterSearchFilterBackend,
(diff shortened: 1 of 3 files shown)


peachjam_search/views.py
```diff
@@ -21,51 +21,54 @@
...
+                },
             )
-        return queries
+        ]
 
 
-class CustomSearchBackend(BaseSearchFilterBackend):
-    query_backends = [PageQueryBackend]
+class SearchFilterBackend(CompoundSearchFilterBackend):
+    query_backends = [
```
longhotsummer, line 66: This is how we get it to combine everything into one `should` clause.

actlikewill (approved):
It works!

PR #340, opened 2022-08-25
Compile and store app-prod.js in the repo

longhotsummer:
Otherwise, down-stream dependencies don't have access to the compiled code.


actlikewill (approved):
Works well.

musangowope (approved):

PR #351, opened 2022-08-30
Fetch relationships

nickmwangemi:


peachjam/adapters/adapters.py
```diff
@@ -158,10 +179,78 @@ def update_document(self, url):
...
             if pubdoc and pubdoc["url"]:
-                self.download_source_file(pubdoc["url"], doc, title)
+                self.download_source_file(pubdoc["url"], created_doc, title)
         else:
             # the source file is the PDF version
-            self.download_source_file(f"{url}.pdf", doc, title)
+            self.download_source_file(f"{url}.pdf", created_doc, title)
+
+        self.fetch_relationships(document, created_doc)
+        logger.info("Fetching of relationships is complete!")
```
longhotsummer, line 188: Minor style: put logging together with the code that is responsible for it. So move this to the end of `fetch_relationships`.

peachjam/adapters/adapters.py
```diff
@@ -158,10 +179,78 @@ def update_document(self, url):
...
+
+    def fetch_relationships(self, imported_document, created_document):
+        subject_work = created_document.work
+
+        if imported_document["repeal"]:
+            repealing_work, _ = Work.objects.get_or_create(
+                frbr_uri=FrbrUri.parse(imported_document["repeal"]["repealing_uri"]),
```
longhotsummer, line 195: The database considers `frbr_uri` to be a string, so you don't want to parse it.

peachjam/adapters/adapters.py
```diff
@@ -158,10 +179,78 @@ def update_document(self, url):
...
+                    )
+                    self.create_relationship(
+                        predicate_idx=2,
+                        subject_work=subject_work,
+                        object_work=commencing_work,
+                        relationship="Commencing",
+                    )
+
+    def create_relationship(
+        self, predicate_idx, subject_work, object_work, relationship
```
longhotsummer, line 237: This method is taking two parameters which identify the relationship: `predicate_ix` and `relationship`, so there is repetition. Also, the `relationship` value is only used in a log.

I suggest dropping the `relationship` parameter and using the `name` value of the predicate object for the log.

peachjam/adapters/adapters.py
```diff
@@ -158,10 +179,78 @@ def update_document(self, url):
...
+    def fetch_relationships(self, imported_document, created_document):
+        subject_work = created_document.work
+
+        if imported_document["repeal"]:
+            repealing_work, _ = Work.objects.get_or_create(
+                frbr_uri=FrbrUri.parse(imported_document["repeal"]["repealing_uri"]),
+                title=imported_document["repeal"]["repealing_title"],
+            )
+            self.create_relationship(
+                predicate_idx=1,
```
longhotsummer, line 199: Magic numbers are always a bit dangerous, since it's unclear if they're correct and easy to go wrong when someone later adds a new one or decides to sort them in the array differently, etc. A more explicit alternative is to use a dict and pass in the predicate slug, rather than a magic number.

```python
self.predicates = {
    # this is implicitly the slug
    "amended-by": {
        "name": "amended by",
        "verb": "is amended by",
        "reverse_verb": "amends",
    },
    ...
}

...

self.create_relationship("amended-by", ...)

...

predicate = self.predicates[slug]
predicate, _ = Predicate.objects.get_or_create(slug, defaults={... slug=slug ...})
```

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -116,21 +116,21 @@ <h5>{% translate 'Related documents' %}</h5>
 
                           <ul class="list-unstyled">
                             {% for rel in document.relationships_as_subject %}
-                              {% if rel.object_document %}
+                              {% if rel.object_work %}
```
longhotsummer, line 119: Nice, we didn't previously have work details or FRBR URIs, so this should be better.

longhotsummer (changes requested):
Looks good, minor style changes suggested.

peachjam/adapters/adapters.py
nickmwangemi: Done

peachjam/adapters/adapters.py
nickmwangemi: Done

peachjam/adapters/adapters.py
nickmwangemi: Nice catch!

nickmwangemi:
@longhotsummer Should be good to go now.

longhotsummer (approved):
great

Issue #356, opened 2022-08-31
Detail page should prompt to load large pdf


musangowope:
> > @longhotsummer On the task to prompt to load large files, what is the threshold for a large file? Also what is the best way to handle this, should we alter the response on the document source pdf view or should we handle it with the frontend javascript?
> 
> I believe this functionality already exists in Peach Jam for large files? @musangowope can you confirm?

@longhotsummer.  We did discuss the idea for this feature in the past but it does not yet exist. I'm going to go ahead and make a ticket on peachjam for this feature.

_Originally posted by @musangowope in https://github.com/laws-africa/gazettemachine/issues/104#issuecomment-1232577550_

musangowope:
Process:
- If pdf document is large, give user decision to either "Proceed to load document" or "Cancel"
- If the user selects cancel, loading does not proceed. They can download the document from meta block link. 
(Perhaps provide the option to load the document again?)

Questions:
- What size does the pdf need to be for this to happen?

longhotsummer:
Thanks Musango

I don't think there needs to be a cancel. Just the button to load the PDF if they want to.

Since it's a pdf thing, it should be done as part of the pdf viewer. 

I suggest a threshold configurable via the django view class. Set it as a data attribute. 

5MB by default?

PR #358, opened 2022-08-31
Handling large pdf

musangowope:
Add logic to include prompt if pdf file is large

Demo:
https://www.loom.com/share/3359323674114bd59df2dc7585b0152d

resolves https://github.com/laws-africa/peachjam/issues/356

### DIFF
--- a/peachjam/js/components/pdf-renderer.ts
+++ b/peachjam/js/components/pdf-renderer.ts
@@ -20,13 +20,11 @@ class PdfRenderer {
   protected root: HTMLElement;
   protected scrollListenerActive: boolean;
   protected pdfContentMarks: any[];
-  private readonly pdfSize: string | undefined;
   public onPreviewPanelClick: () => void;
   public onPdfLoaded: () => void;
   constructor (root: HTMLElement) {
     this.root = root;
     this.pdf = root.dataset.pdf;
-    this.pdfSize = root.dataset.pdfSize;
     this.pdfContentWrapper = root.querySelector('.pdf-content');
     this.scrollListenerActive = true;
     this.pdfContentMarks = [];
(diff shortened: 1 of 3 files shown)


peachjam/templates/peachjam/_document_content.html
```diff
@@ -149,14 +152,21 @@
...
               <div class="progress">
                 <div class="progress-bar bg-primary progress-bar-striped progress-bar-animated"
                      role="progressbar">
                 </div>
               </div>
             </div>
+
             <div class="pdf-content"></div>
+            <div class="pdf-large-prompt">
+              <p>{% trans 'This current document is more than 5MB. Proceed to load?' %}</p>
```
longhotsummer, line 165: This should tell the user how big it is -- 100MB? 5.1MB? I think this would be clearer:

"This document is 52MB. Do you want to load it?"

peachjam/js/components/pdf-renderer.ts
```diff
@@ -37,40 +35,58 @@ class PdfRenderer {
...
+            this.activatePreviewPanel(current);
+            const scrollableContainer = this.root.querySelector('[data-preview-scroll-container]');
+            if (!scrollableContainer) return;
+            scrollableContainer.scrollTop = (current.offsetTop + current.clientHeight) - (current.offsetHeight * 2);
+          }
+        }, 20));
+      });
     });
+
+    startPdfSetupObserver.observe(this.root, {
```
longhotsummer, line 79: Why use an observer like this? It seems a bit complex and indirect. This component controls everything, why not have a `loadPdf` method that is called either when the button is clicked, or when the component loads (if the PDF isn't too large)?

peachjam/templates/peachjam/_document_content.html
musangowope: refactored

peachjam/js/components/pdf-renderer.ts
musangowope: refactored

longhotsummer (approved):
great

PR #360, opened 2022-09-01
Source File size except block 

actlikewill:
- Adding this except block to guard against the exception thrown when adding files manually from gazette machine.

### DIFF
--- a/peachjam/models/core_document_model.py
+++ b/peachjam/models/core_document_model.py
@@ -314,8 +314,11 @@ class SourceFile(AttachmentAbstractModel):
     file = models.FileField(upload_to=file_location, max_length=1024)
 
     def save(self, *args, **kwargs):
-        # store this so that we don't have to calculate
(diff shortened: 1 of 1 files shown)


longhotsummer (commented):
If `self.file.size` fails, won't `self.file.read()` for the mimetype check also fail? I'm trying to understand why we'd be able to get this far if we can't read the file.

peachjam/models/core_document_model.py
```diff
@@ -314,8 +314,11 @@ class SourceFile(AttachmentAbstractModel):
     file = models.FileField(upload_to=file_location, max_length=1024)
 
     def save(self, *args, **kwargs):
-        # store this so that we don't have to calculate
-        self.size = self.file.size
+        try:
```
longhotsummer, line 317: As mentioned, I suggest we remove the exception handler and instead do:

```python
if not self.size:
    self.size = self.file.size
```

PR #368, opened 2022-09-09
Fix notices

nickmwangemi:
- Display notices 
- Add the text-nowrap class to prevent dates in the recent documents table from wrapping

image


peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -137,6 +137,8 @@ <h5>{% translate 'Related documents' %}</h5>
             </div>
           {% endblock %}
 
+          {% include 'peachjam/_notices.html' with messages=notices %}
```
longhotsummer, line 140: @nickmwangemi pls put the notices beneath the document title, and above the tabs.

longhotsummer (approved):

PR #373, opened 2022-09-13
Children for legislation

longhotsummer:
Each entry in the  serialized JSON for the legislation table now includes `children`, which is a list with a very similar shape to the current list, except `children` and `taxonomies` aren't included.

```javascript
[{
  ...
  "title": "Primary Act",
  "children": [
    {
      "title": "Directions on Measures to Address, Prevent and Combat the Spread of COVID-19 in the Public Transport Services, 2020",
      "citation": "Government Notice 801 of 2020",
      "work_frbr_uri": "/akn/za/act/gn/2020/801",
      "repealed": false
    }
  ]
}, { ... }
]
```

**Front end:**  
add logic to show sublegs in applicable table rows items:

DEMO:
https://www.loom.com/share/a56beb75774747f3b0faf1c7a1aed6eb


musangowope:
@longhotsummer all the necessary front-end logic has been implemented👍🏽

peachjam/js/components/LegislationTable/index.vue
```diff
@@ -122,30 +125,50 @@
...
+              <div
+                v-if="row.children.length"
+                :id="`row-accordion-${index}`"
+                class="accordion-collapse collapse accordion"
+                data-bs-parent=".legislation-table__row"
+              >
+                <div class="d-flex">
+                  <div class="indent" />
+                  <div class="accordion-body flex-grow-1">
+                    <div class="row">
```
longhotsummer, line 137: Since the columns and the info being presented are the same as the parent table -- title and citation -- I think these are redundant and can be removed.

longhotsummer (commented):
@musangowope looks great, small change to remove the titles in the subleg list, since they're the same as the parent table.

peachjam/js/components/LegislationTable/index.vue
musangowope: done

musangowope (dismissed):

longhotsummer:
Looks good!

Issue #369, opened 2022-09-12
Lawlibrary: link SGBV case summaries to cases


longhotsummer:
There are 19 docs, they can probably be imported manually.

https://lawlibrary.org.za/other/sgbv-library

* [x] import them as a simple "document" type, with a subtype of "case summary" and "guide"
* [x] content will be extracted automatically from the word document
* [x] add a taxonomy term
* [x] add a view that filters to that taxonomy - https://github.com/laws-africa/peachjam/pull/374
* [ ] link them to the related case through a relationship ("summarises" / "is summarised by")

longhotsummer:
All documents imported except the following, for which .docx files are missing:

Case Summary: A K v Minister of Safety and Security and Others (3429 of 2013) [2018] ZAECPEHC 82 (22 November 2018)
Case Summary: S v Beale (A 283 of 2018) [2019] ZAWCHC 55 (03 May 2019)
Case Summary: S v ML 2016 (2) SACR 160 (SCA)
Guide to Sexualised Behaviour in Child Sexual Abuse Victims
Guide: A Brief Explanation of Cognitive Development to Better Understand the Child Witness

longhotsummer:
Once the judgments are imported, we can link them through relationships.

longhotsummer:
All remaining cases and guides imported.

longhotsummer:
We're still waiting on the judgments to be imported. I think blocked by #328

longhotsummer:
Outstanding cases to be linked:


Case summary: Du Toit v The Magistrate and Others 2016 (2) SACR 112 (SCA)		7 November 2019
Case summary: Lance Bessick v The State. Case no. A5392010. Judgement 29 May 2012. High Court (Cape Town)		7 November 2019
Case summary: De Klerk v State, KZN High Court (Pietermaritzburg), Appeal No. AR 3262009, 8 October 2009		7 November 2019
Case summary: S v Ngcobo 2018 (1) SACR 479 (SCA)		7 November 2019
Case summary: S v W D (AR 326 of 2009) [2009] ZAKZPHC 48 (08 October 2009)		7 November 2019
Case summary: AK v Minister of safety and Security & Others 2019 (1) SACR 529 (ECP)		7 November 2019
Case summary: Beale v State. Case no. A283/18. 3 May 2019. High Court of South Africa Western Cape Division. (Cape Town)		7 November 2019
[...]

PR #374, opened 2022-09-13
taxonomy detail page (doc listing page)

longhotsummer:
* adds a new taxonomy detail page which lists documents that have that taxonomy
* re-orders the URLs in urls.py to group logically
* make the taxonomy slug field unique
* add a link in lawlibrary for the SGBV library
* make nav links non-bold

Later, we can add a taxonomy listing page. This is all we need right now for lawlibrary.

image


longhotsummer:
Part of #369

peachjam/static/stylesheets/components/_navbar.scss
```diff
@@ -2,7 +2,6 @@
   .nav-item,
   .nav-link {
     color: black;
-    font-weight: bold;
```
actlikewill, line 5: his will have an effect on Africanlii as well. Are we changing the nav font weight across all the platforms?

actlikewill (approved):
Looks good.

PR #377, opened 2022-09-13
Agp and other adjustments

longhotsummer:
* updated translation strings
* improved search result layout
* improved document table layouts - move the date to the right, so that the "empty" column isn't as jarring
* make authors clickable in the document detail view
* in search results, show jurisdiction; show only snippet or pages, not both

image

image

image


actlikewill (approved):
Looks good and works well.

PR #378, opened 2022-09-13
Handle blank soffice output

longhotsummer:


actlikewill:
It now uploads without the error. However, the document now exists without content html. There will be no way for me to know which other documents have errors and isolate them during import.

longhotsummer:
Can you check for empty `content_html`?

actlikewill:
> 

We can check for empty html at the step where we extract. I can push on to this branch if that is ok.

longhotsummer:
> > 
> 
> We can check for empty html at the step where we extract. I can push on to this branch if that is ok.

Yup, go for it -- add whatever you need.

PR #388, opened 2022-09-14
Correctly set parent work when importing

longhotsummer:
Also avoid potential conflict when checking for existing related works.


actlikewill (approved):
Looks good

PR #394, opened 2022-09-14
Guard against a document existing but as the wrong document type

longhotsummer:
There are about 86 lawlibrary imports that are failing because of this.


longhotsummer:
I've deleted the incorrect documents. So this PR will just help with debugging this in future.

actlikewill (approved):
Awesome. This was confusing me quite a bit as well because we had used create_or_update. 😕

PR #401, opened 2022-09-14
Fix pagination for long lists

### LINKED ISSUE #396: Pagination goes crazy for very long lists
image

Two suggestions:

* for the "Recent judgments" list at https://lawlibrary.staging.laws.africa/judgments/ -- just show the most recent 20 judgments and don't show pagination at all
* when showing pagination elsewhere, we should be smarter about showing maybe 5 pages on each side of the current page.

See https://docs.djangoproject.com/en/4.1/ref/paginator/#django.core.paginator.Paginator - could get_elided_page_range be useful? So we effectively have page buttons corresponding to something like `[1, 2, '…', 7, 8, 9, 10, 11, 12, 13, '…', 49, 50]`

actlikewill:
- Adds ellipsis to pagination for long lists.

pagination

closes #396

### DIFF
--- a/peachjam/templates/peachjam/_pagination.html
+++ b/peachjam/templates/peachjam/_pagination.html
@@ -1,4 +1,4 @@
-{% load i18n %}
+{% load i18n peachjam %}
 
 <ul class="pagination">
   {% if page_obj.has_previous %}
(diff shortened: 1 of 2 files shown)


peachjam/templates/peachjam/_pagination.html
```diff
@@ -7,10 +7,17 @@
     </li>
   {% endif %}
 
-  {% for num in paginator.page_range %}
+{% get_proper_elided_page_range paginator page_obj.number as page_range %} <!-- here -->
```
longhotsummer, line 10:
```suggestion
{% get_proper_elided_page_range paginator page_obj.number as page_range %}
```

longhotsummer (approved):

PR #404, opened 2022-09-15
Remove Lawlib pagination on recent judgments

actlikewill:
- removes the pagination for recent judgments
- fixes a small typo.


longhotsummer (approved):
This is fine -- if you want to fix lawlibrary.JudgmentListView to just be a TemplateView, that would be good too :)

PR #407, opened 2022-09-15
Allow override of case number string

longhotsummer:
This is required in some cases where the generated string doesn't capture enough information or follow the right format.


actlikewill (approved):
Looks good.

PR #408, opened 2022-09-15
Paginator bug fix

nickmwangemi:
https://www.loom.com/share/880fe74c709f41cfb3a335e52cbedc7a


peachjam/templatetags/peachjam.py
```diff
@@ -37,3 +38,32 @@ def get_proper_elided_page_range(p, number, on_each_side=3, on_ends=2):
...
+            query_dict.setlist(k, v)
+        else:
+            query_dict[k] = v
+
+    for k in remove_keys:
+        if k in query_dict:
+            del query_dict[k]
+
+    qs = query_dict.urlencode()
+    return f"?{qs}" if qs else ""
```
longhotsummer, line 69: you have "?" both here and in the template. I'd say keep it in the template and drop it here, because this helper doesn't know if it's building the entire querystring or not.

longhotsummer (approved):

PR #409, opened 2022-09-15
Defer expensive fields by default

longhotsummer:
content_html, metadata_json and toc_json are exclusively used in detail views; for listing views, don't load them because it's a huge amount of data which is not used


longhotsummer:
The homepage can currently take 20+ seconds to load, this fix should make it significantly faster.

nickmwangemi (approved):

Issue #403, opened 2022-09-15
Search page improvements


longhotsummer:
* [ ] result must include court for judgments - #429 
* [ ] #493
* [x] don't include facets that don't have values (blocked by #361)
* [x] #405

Notes:
* https://www.loom.com/share/1493f76bff0c48d3b4725bbffecc2a7e
* https://www.loom.com/share/db4fb3e099e34b1399fc147fa20c5a87

longhotsummer:
Closing in favour of individual issues.

PR #410, opened 2022-09-15
Store concept of most recent doc, and keep up to date

longhotsummer:
Ref #403

This allows us to only search for the most recent document when the are multiple documents for a work, at different dates. This means we'll show only the most recent piece of legislation in a search result, and not older documents.

We do this by calculating and storing (only in ES) a new field -- `is_most_recent`. A document is the most recent if it has the latest date of all documents linked to the same work. Note that there can be multiple most-recent documents, since there may be documents at the latest date in different languages.

In order to keep this field up-to-date, we need to detect when an old most-recent-document is now out of date. A signal handler checks when a document has been saved -- if it's the current most recent, it assumes that all the previous documents of the same work could potentially have been the old  most recent one, so it fakes a post-save signal to trigger a re-index.

### DIFF
--- a/peachjam/js/components/FindDocuments/index.vue
+++ b/peachjam/js/components/FindDocuments/index.vue
@@ -426,6 +426,7 @@ export default {
           params.append('page', this.page);
           params.append('ordering', this.ordering);
           params.append('highlight', 'content');
+          params.append('is_most_recent', 'true');
 
           for (const key of Object.keys(this.filters)) {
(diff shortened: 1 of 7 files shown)


longhotsummer:
We'll need to rebuild the search index after this is deployed.

longhotsummer:
I had considered not indexing the older documents at all, but that seemed overly complicated. This also keeps the door open to allow us to deliberately search the older documents, if necessary.

PR #411, opened 2022-09-15
Fix resources for sca upload

actlikewill:
- Fix resource for upload

### DIFF
--- a/peachjam/resources.py
+++ b/peachjam/resources.py
@@ -136,9 +136,11 @@ class JudgmentResource(BaseDocumentResource):
         model = Judgment
 
     def before_import_row(self, row, **kwargs):
-        Author.objects.get_or_create(
-            code=row["court_obj"]["code"], defaults={"name": row["court_obj"]["name"]}
-        )
+        if "court_obj" in row:
+            Author.objects.get_or_create(
+                code=row["court_obj"]["code"],
+                defaults={"name": row["court_obj"]["name"]},
+            )
 
     def after_import_row(self, row, instance, row_number=None, **kwargs):
         super().after_import_row(row, instance, row_number, **kwargs)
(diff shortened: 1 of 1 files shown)


peachjam/resources.py
```diff
@@ -150,12 +152,21 @@ def after_import_row(self, row, instance, row_number=None, **kwargs):
...
-                    document=judgment,
-                    number=case_number["number"],
-                    year=case_number["year"],
-                    matter_type=MatterType.objects.get(name=case_number["matter_type"]),
-                )
+        if "string_override" in row:
+            CaseNumber.objects.create(
+                string_override=row["string_override"], document=judgment
+            )
+        if "case_numbers" in row:
```
longhotsummer, line 159: These must be on the same object. Otherwise we'll get one for the string override, and one for the split out details, which is two case numbers.

peachjam/resources.py
```diff
@@ -150,12 +152,21 @@ def after_import_row(self, row, instance, row_number=None, **kwargs):
...
-                )
+        if "string_override" in row:
+            CaseNumber.objects.create(
+                string_override=row["string_override"], document=judgment
+            )
+        elif "case_numbers" in row:
+            for case_number in row["case_numbers"]:
+                if case_number["number"]:
+                    MatterType.objects.get_or_create(name=case_number["matter_type"])
+                    CaseNumber.objects.create(
```
longhotsummer, line 163: Just throw the string override (if present), into this object and do away with line 155-158. It's fine if it's null or empty. The code will only use it if it's valid.

longhotsummer: Basically, they're not exclusive. The SCA dataset has both the override and the breakdown, and we want to keep both for different purposes.

peachjam/resources.py
```diff
@@ -150,12 +152,21 @@ def after_import_row(self, row, instance, row_number=None, **kwargs):
...
-                    document=judgment,
-                    number=case_number["number"],
-                    year=case_number["year"],
-                    matter_type=MatterType.objects.get(name=case_number["matter_type"]),
-                )
+        if "string_override" in row:
+            CaseNumber.objects.create(
+                string_override=row["string_override"], document=judgment
+            )
+        elif "case_numbers" in row:
```
actlikewill, line 159: @longhotsummer  Changed this to `elif`. We will change this whole section when we update and standardize this code.

PR #415, opened 2022-09-16
Show related documents in a tab

longhotsummer:
* always show all related documents in a tab
* if there are 4 related documents or fewer, show them in a list
* if there are more than 4, then show a link to toggle to the tab
* also: hide the language chooser for lawlibrary

image

image

image


peachjam/js/components/tabs.ts
```diff
@@ -0,0 +1,21 @@
+export class ToggleTab {
+  btn: HTMLElement;
+
+  constructor (btn: HTMLElement) {
```
musangowope, line 4: instead btn I would call it `anchor` since we're getting an `href` attribute from the element. 
If we want to be more explicit, perhaps we could even change HTMLElement to `HTMLAnchorElement`

musangowope (commented):
Looks good. Minor changes

peachjam/js/components/tabs.ts
longhotsummer: done

musangowope (approved):

PR #436, opened 2022-09-19
Fixes resource mimetype

actlikewill:
- fixes appilcation/zip mimetype


longhotsummer (approved):
Fragile but let's go with it for now.

PR #444, opened 2022-09-20
Add logic to temporarily disable an ingestor

nickmwangemi:


longhotsummer (approved):
Does the checkbox show up in the admin view?

nickmwangemi:
> I've updated it show 

image

longhotsummer (approved):

Issue #352, opened 2022-08-30
As a user clicking through to a search result, I want the search to be carried through


longhotsummer:
ie. can we pre-fill the search functionality on the document page so that the search highlights show up?

longhotsummer:
The idea here is when popping open a result from the site search page, add search=foo or something to the url. In the document detail page, when the in-document search component fires up, check the query string and pretend the user entered that search. 

There may be some interaction between waiting for the pdf (if there's no html) to load before kicking off the search.

longhotsummer:
Done

PR #445, opened 2022-09-20
As a user clicking through to a search result, I want the search to be carried through

Sandravaphilips:
- Issue: https://github.com/laws-africa/peachjam/issues/352
- Loom: https://www.loom.com/share/30fe2bc271054ca28bda3fcc36d428c2


peachjam/js/components/FindDocuments/SearchResult.vue
```diff
@@ -4,7 +4,7 @@
       class="h5"
       target="_blank"
       rel="noreferrer"
-      :href="item.expression_frbr_uri"
+      :href="`${item.expression_frbr_uri}?q=${query}`"
```
musangowope, line 7: wrap `q` in `encodeURIComponent`

musangowope, line 7: wrap `query` with `encodeURIComponent`

musangowope (changes requested):
Great first step. The next step would be to accomodate pdf documents. You need to trigger the search only once the pdf document has loaded

musangowope:
Taking videos of work done for prs can also be really valuable. We like to use a tool called [Loom](https://www.loom.com/)

peachjam/js/components/document-content.ts
```diff
@@ -59,6 +59,13 @@ class DocumentContent {
       this.pdfRenderer = new PdfRenderer(root);
       this.pdfRenderer.onPreviewPanelClick = () => { this.navOffCanvas?.hide(); };
       this.pdfRenderer.onPdfLoaded = () => {
+        const searchInput: HTMLInputElement | null = this.root.querySelector('[aria-label="Search document content"]');
+        const searchForm: HTMLFormElement | null = this.root.querySelector('.doc-search__form');
+        if (searchInput) {
+          searchInput.value = 'law';
```
musangowope, line 65: this is a hard-coded value. I'm guessing you were using this for testing purposes. Please set `search.value` to the  `url param` `q`

musangowope (changes requested):
Minor change

musangowope:
it would also be amazing if we could immediately activate the search tab in `in-document search` if  url param `q` exists

longhotsummer:
> it would also be amazing if we could immediately activate the search tab in `in-document search` if url param `q` exists

Yes please!

musangowope (approved):

PR #447, opened 2022-09-22
Django Jazzmin

### LINKED ISSUE #441: Which admin skin should we use?
* https://django-jazzmin.readthedocs.io/
* https://grappelliproject.com/

Ideal requirements:

* tabbed "change" views (this is essential, we don't want to have to implement it ourselves)
* more flexible filters in list views: we want to be able to specify 5/6 different filter types and not have django show all options as a long list down the right side of the page
* works well with other plugins
* solid and well maintained

actlikewill:
- [x] Install django jazzmin
- [x] Add jazzmin settings config
- [x] Customise ui
- [x] Tighten CSS paddings and margins

closes #441 

https://www.loom.com/share/82c3b828bd154552a4964a03d183ec31


peachjam/settings.py
```diff
@@ -83,6 +84,85 @@
 
 ROOT_URLCONF = "peachjam.urls"
 
+JAZZMIN_SETTINGS = {
+    "site_title": "Peachjam",
```
longhotsummer, line 88: Use `PEACHJAM["APP_NAME"]` where possible

peachjam/settings.py
```diff
@@ -83,6 +84,85 @@
...
+    "site_header": "Peachjam",
+    "site_brand": None,
+    "site_icon": None,
+    "site_logo": "/images/africanlii-logo.svg",
+    "welcome_sign": "Welcome to Peachjam Admin",
+    "copyright": "Laws.Africa",
```
longhotsummer, line 94: Leave blank

africanlii/settings.py
```diff
@@ -3,3 +3,7 @@
 INSTALLED_APPS = ["africanlii.apps.AfricanliiConfig"] + INSTALLED_APPS  # noqa
 
 ROOT_URLCONF = "africanlii.urls"
+
+JAZZMIN_SETTINGS["site_title"] = "Africanlii"  # noqa
```
longhotsummer, line 7: All three of these can default to `PEACHJAM["APP_NAME"]` and be set in peachjam.settings so that they don't have to be set here. More automatic, less manual :)

peachjam/settings.py
```diff
@@ -83,6 +84,85 @@
 
 ROOT_URLCONF = "peachjam.urls"
 
+JAZZMIN_SETTINGS = {
+    "site_title": "Peachjam",
+    "site_header": "Peachjam",
+    "site_brand": None,
+    "site_icon": None,
+    "site_logo": "/images/africanlii-logo.svg",
```
longhotsummer, line 92: I think it's too awkward to make a logo work across all the sites, and this can just be unset.

peachjam/settings.py
```diff
@@ -83,6 +84,85 @@
...
+        "peachjam.Predicate": "fa fa-exchange-alt",
+        "django.Site": "fa fa-server",
+        "socialaccount.SocialAccount": "fa fa-users",
+        "socialaccount.SocialToken": "fa fa-key",
+        "socialaccount.SocialApp": "fa fa-puzzle-piece",
+    },
+    "related_modal_active": False,
+    "custom_css": "stylesheets/jazzmin_admin.css",
+    "show_ui_builder": False,
+    "changeform_format": "horizontal_tabs",
```
longhotsummer, line 134: 🚀

peachjam/settings.py
```diff
@@ -83,6 +84,85 @@
...
+        "peachjam.DocumentNature": "fa fa-list",
+        "peachjam.SourceFile": "fa fa-file-archive",
+        "peachjam.Taxonomy": "fa fa-tags",
+        "peachjam.Predicate": "fa fa-exchange-alt",
+        "django.Site": "fa fa-server",
+        "socialaccount.SocialAccount": "fa fa-users",
+        "socialaccount.SocialToken": "fa fa-key",
+        "socialaccount.SocialApp": "fa fa-puzzle-piece",
+    },
+    "related_modal_active": False,
```
longhotsummer, line 131: I think we should make this `true`

longhotsummer:
The themes are simply the compiled CSS from [bootswatch](https://bootswatch.com/) -- see here: https://github.com/farridav/django-jazzmin/blob/fcd2161fe47159bcbee14f2eefc8777d9082c4c9/jazzmin/settings.py#L162

Which is great because it means we can really easily compile our own bootstrap.css file from customised SASS variables and just dump it in. Bootswatch themes are just standard bootstrap with custom CSS vars (and a little extra bootswatch magic). eg. see https://github.com/thomaspark/bootswatch/tree/v5/dist/simplex for the simplex theme.

actlikewill:
Sounds great. I can dig into it.

longhotsummer:
@actlikewill I think this is good to go!

longhotsummer (approved):

PR #457, opened 2022-09-23
Judgment media summary inline

### LINKED ISSUE #453: As a content editor, I want to be able to upload a Judgment media summary file without having to enter the Size and Mime Type.
Editors need to be able to upload judgment summaries that have been captured in a separate document. They currently cannot do this without populating the **Size** and **Mime type** fields, and they do not know how to populate these fields. If these could be made non-mandatory fields, that would solve the problem.

image

actlikewill:
Sets judgment media summary files inline to jugment model
- [x] Create migrations
- [x] Remove old admin

closes #453


peachjam/models/core_document_model.py
```diff
@@ -324,13 +324,22 @@ class AttachmentAbstractModel(models.Model):
...
     size = models.BigIntegerField(default=0)
+    file = models.FileField()
 
     def __str__(self):
         return f"{self.filename}"
 
     class Meta:
         abstract = True
 
+    def save(self, *args, **kwargs):
```
longhotsummer, line 335: The similar code in `SourceFile` can fall away.

peachjam/admin.py
```diff
@@ -85,8 +85,7 @@ class SourceFileAdmin(admin.ModelAdmin):
 admin.site.register(SourceFile, SourceFileAdmin)
 
 
-class SourceFileInline(admin.TabularInline):
-    model = SourceFile
+class BaseAttachmentFileInline(admin.TabularInline):
     extra = 0
     readonly_fields = ("filename", "mimetype", "attachment_link")
```
longhotsummer, line 90: Size as well?

peachjam/models/core_document_model.py
actlikewill: removed. ✔️

peachjam/admin.py
actlikewill: size added. ✔️

longhotsummer (approved):

PR #459, opened 2022-09-23
Strip "position: *" when importing docx

### LINKED ISSUE #456: Image wonky - positioned in the top left corner of the page
Windows|Chrome

https://lawlibrary.org.za/akn/za/judgment/zagpjhc/2022/48/eng@2022-02-11

image

longhotsummer:
Fixes #456 

Disallow any "position" attribute in imported docx HTML since it could move content anywhere on the page.

Before:

image

After:

image

### DIFF
--- a/peachjam/pipelines.py
+++ b/peachjam/pipelines.py
@@ -30,7 +30,11 @@ class RemoveTags(Stage):
 
 class CleanStyles(Stage):
     # styles we explicitly want to remove
-    style_blacklist = {"p": set("line-height".split()), None: set()}
+    style_blacklist = {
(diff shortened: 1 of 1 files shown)


longhotsummer:
There are 260 documents impacted by this which will need to have their HTML re-processed. 259 are judgments, 1 is a generic document.

```sql
select
  id, doc_type
from
  peachjam_coredocument
where
  content_html like '%position: abs%';
```

actlikewill (approved):
Works well.

longhotsummer:
Updated all 11 impacted agp docs.

Issue #400, opened 2022-09-14
Legislation table layout adjustments


longhotsummer:
* make the title column effectively col-8, and the numbered title col-4 (ie. two thirds and one third)
* give the subleg items some bottom padding to space them out
* don't open anything in a new tab

![image](https://user-images.githubusercontent.com/4178542/190197014-8c4f574f-1082-4c41-97fd-473746edf813.png)

Looking at the code, it seems to me that we can use the same indent + col-primary + col-secondary concept on the main rows, and on the children rows -- they don't have to be styled differently and the columns will then line up. Then we add some extra left padding/margin on the children to indent a bit further.

PR #469, opened 2022-09-26
Legislation table layout adjustments

Sandravaphilips:
- Adjusted layout of LegislationTable
- Issue: https://github.com/laws-africa/peachjam/issues/400

Screenshot 2022-09-26 at 19 49 29


musangowope:
Some feedback:
1) give the subleg items some bottom padding to space them out
2) We need to use the same indent + col-primary + col-secondary concept for the both the main rows and children rows. i.e we should be using the same classes for main rows and children rows. Let me know if you like to chat through this

musangowope (approved):
Nice work!

musangowope (dismissed):

longhotsummer:
Sorry guys, this isn't working yet.

image

* The numbered title column must align consistently, otherwise there's visual noise.
* The point of widening the columns is to take advantage of horizontal space for long titles; the long title is now wrapping too soon, especially on mobile

I think the key issue is how you're doing the columns -- it's too complicated, we definitely don't need flex-grid in the mix.

Take a look at this: https://jsfiddle.net/j572u18h/1/

Key ideas:

* make the primary row content and its secondary row content "peers" in separate divs (don't mix the two columns and the secondary content as three peer divs)
* add the indent to the left of the secondary content just with padding on the div
* let flex-basis control the width of the two major columns
* add padding if necessary on the table "cells"

You could probably even get away with using `.row` and `.col-9` and `.col-3` on the two major columns.

longhotsummer:
So it's actually very close, I suggest:

* drop the indent div in the secondary row, using padding on the title column instead -- now the columns will align
* use flex rather than grid -- fixes the early wrapping because we don't need an empty column there
* add padding/margin to the right of the title column so that there is some padding between the two columns

longhotsummer:
And I guess the grid could be used instead of flex, but we need to drop the empty column and add some padding on the right :)

musangowope:
@longhotsummer made discussed changes
https://www.loom.com/share/6ac5d52a13214cc29d778a3e32adf75d

longhotsummer (approved):
nice, much simpler markup

PR #472, opened 2022-09-27
Add robots.txt

nickmwangemi:


peachjam/templates/peachjam/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow: /
```
longhotsummer, line 2: We WANT google to index everything! This should be `Allow: /`

longhotsummer (approved):

PR #474, opened 2022-09-27
Add measure to check that object has GlobalWorkOptions before proceeding

musangowope:


longhotsummer:
Are you able to reproduce the error? This seems like `pdfjsLib` is not defined, rather than anything to do with `GlobalWorkerOptions`.

PR #475, opened 2022-09-27
Serial number override field on Judgment model

### LINKED ISSUE #452: As a content editor uploading judgments in a non-bulk manner, I need to be able to override the MNC generator by inputting a court generated MNC
At the moment the MNC is automatically generated based on the order of uploading judgments and editors working via the editorial interface are unable to override the system. 

This issue blocks the content team from uploading judgments from courts that do assign their MNCs

actlikewill:
- [x] Add serial number override field on Judgments model
- [x] Raise validation error when serial number exists

closes #452


longhotsummer:
@actlikewill I've pushed a change to this. The `clean()` already ensures the FRBR URI is unique (or it should, at least). So when we're generating the MNC, we should fall back to using the serial number override if it's available. Since `serial_number` is an int field but `serial_number_override` is a char field, we can't assign the one to the other.

Let's see if this works...

actlikewill:
@longhotsummer I'd like to merge #477 first before this.

longhotsummer (dismissed):

longhotsummer:
@actlikewill over to you

longhotsummer (dismissed):

peachjam/models/judgment.py
```diff
@@ -76,6 +84,14 @@ def generate_serial_number(self):
         if self.pk:
             query = query.exclude(pk=self.pk)
 
+        serial_nums = [s["serial_number"] for s in query.values("serial_number")]
+
+        if self.serial_number_override:
+            if int(self.serial_number_override) in serial_nums:
+                raise ValidationError("Serial number already in use.")
+            else:
+                return self.serial_number_override
```
actlikewill: @longhotsummer Do we not need to enforce uniqueness for mnc serial numbers?

longhotsummer: I've been thinking about this. In the short term, not having this support is preventing uploaders from adding new content for courts that assign MNCs themselves, where there is basically no chance of overlap. So I'd like to unblock that ASAP.

When I worked on this yesterday I expected the FRBR URI uniqueness to enforce unique serial numbers. However, I realised that the code actually allows duplicate FRBR URIs, because it thinks the two documents belong to the same work.

Can we merge this in to unblock and then re-evaluate?

peachjam/models/judgment.py
```diff
@@ -86,12 +94,16 @@ def __str__(self):
     def assign_mnc(self):
         """Assign an MNC to this judgment, if one hasn't already been assigned."""
         if self.date and hasattr(self, "court"):
+            self.serial_number = self.generate_serial_number()
             if self.mnc != self.generate_citation():
-                self.serial_number = self.generate_serial_number()
```
actlikewill, line 90: @longhotsummer This allows new mnc to be generated on change of serial number override.

actlikewill, line 90: @longhotsummer This allows new mnc to be generated on change of serial number override.

longhotsummer: But it also means that a new serial number will be assigned on older judgments unnecessarily :( 

Let me think about how we do this differently.

longhotsummer:
https://www.loom.com/share/61e2e77d76c0421180447e30af82517c

@actlikewill i've made a change here:

1. when assigning the MNC, only call `generate_serial_number` if the current MNC doesn't match the re-computed MNC (ie. court or year have changed), or if there is an override and the assigned serial doesn't match the override.
2. the judgment code itself doesn't have enough knowledge to know when the override has been cleared and a new serial must be assigned. So the Admin form now clears the serial number if the override is cleared, allowing a fresh serial to be assigned.

longhotsummer (approved):

PR #476, opened 2022-09-27
Fix wrapping on document detail page

nickmwangemi:
Without relationships
image

With relationships
image

### DIFF
--- a/peachjam/templates/peachjam/layouts/document_detail.html
+++ b/peachjam/templates/peachjam/layouts/document_detail.html
@@ -51,7 +51,7 @@
           <div class="container">
             {% block document-metadata %}
               <div class="row">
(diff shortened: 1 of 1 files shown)


longhotsummer:
@nickmwangemi this isn't achieving what the original issue asks for. The second column should effectively not be there if there are no relationships.

nickmwangemi:
> @nickmwangemi this isn't achieving what the original issue asks for. The second column should effectively not be there if there are no relationships.

I'll sort it out

longhotsummer (approved):
nice :)

longhotsummer:
Agh, damn, now the key-value pairs are too far apart! :(

longhotsummer:
I'll open another issue, I think musango needs to brew up some fancier CSS.

PR #481, opened 2022-09-28
Show enabled field in admin for Ingestor

nickmwangemi:
Adds the `enabled` field to the admin page. #444 did not include the necessary changes.

image
image


longhotsummer:
I haven't seen the list-editable setup before. I don't think it needs to be editable in the list view, we don't often make changes to this. Showing it readonly in the list, modifiable in the change form, is fine.

longhotsummer (approved):

Issue #430, opened 2022-09-19
UI Analysis and improvements


musangowope:
Peachjam has gotten to a stage where have built multiple pages with various UI elements. It would be helpful if we could assess each page on Africanlii and Lawlibrary and see how we can improve these pages. Things that come to mind are the following:
- Page margin/padding
- Heading/Typogragraphy
- Background colors
- General UI consistency

longhotsummer:
Suggest that we focus on only this right now:

* consistent padding between breadcrumbs (if any) and bottom of navbar
* consistent padding between breadcrumbs and first heading
* consistent padding between first heading and navbar if there are no breadcrumbs
* on the document detail page page, add a full-width background color on the document title, including the page content tabs and the breadcrumb

PR #483, opened 2022-09-29
UI consistency accross pages

musangowope:
- Add refactor margin classes for pages
- Add light background for header element on document detail page

DEMO:
https://www.loom.com/share/068bf961fe114bb48053c8b558d773c6

resolves https://github.com/laws-africa/peachjam/issues/430

### DIFF
--- a/lawlibrary/templates/lawlibrary/judgment_list.html
+++ b/lawlibrary/templates/lawlibrary/judgment_list.html
@@ -6,7 +6,7 @@
 {% block page-content %}
   <section class="pb-5">
     <div class="container">
-      <h1 class="mb-3 mt-4">{% trans 'Judgments' %}</h1>
+      <h1 class="my-4">{% trans 'Judgments' %}</h1>
 
       {% if grouped_courts %}
         <div class="row mb-3">
--- a/lawlibrary/templates/lawlibrary/judgment_list.html
+++ b/lawlibrary/templates/lawlibrary/judgment_list.html
@@ -6,7 +6,7 @@
 {% block page-content %}
   <section class="pb-5">
     <div class="container">
-      <h1 class="mb-4 mt-2">{% trans 'Judgments' %}</h1>
+      <h1 class="mb-3 mt-4">{% trans 'Judgments' %}</h1>
 
       {% if grouped_courts %}
         <div class="row mb-3">
--- a/peachjam/static/stylesheets/_global.scss
+++ b/peachjam/static/stylesheets/_global.scss
@@ -91,7 +91,7 @@ footer {
 }
 
 .breadcrumb {
-  margin-top: map-get($spacers, 1);
+  margin: 0;
 }
 
 // xs btn
--- a/peachjam/static/stylesheets/_global.scss
+++ b/peachjam/static/stylesheets/_global.scss
@@ -91,7 +91,8 @@ footer {
 }
 
 .breadcrumb {
-  margin: 0;
+  padding-top: 1rem;
+  margin-bottom: 0;
 }
 
 // xs btn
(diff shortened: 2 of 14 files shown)


peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -5,44 +5,46 @@
 
 {% block page-content %}
   <main class="pb-5">
-    <div class="container">
-      {% block breadcrumbs %}
-      {% endblock %}
+    <div class="bg-light d-flex">
```
longhotsummer, line 8: Why d-flex?

peachjam/templates/peachjam/layouts/document_list.html
```diff
@@ -19,7 +19,7 @@
     {% endblock %}
 
     {% block page-header %}
-      <div class="row">
+      <div class="row mt-3">
```
longhotsummer, line 22: Would it make sense to instead add margin to the `h1` inside the `page-title` block? 

I think the africanlii page titles for the document lists should have the same margin treatment as for peachjam.

longhotsummer (changes requested):
This looks great!

It looks like we're converging on: the primary `h1` title in the page header should have `my-4` and breadcrumbs, if visible, have `mt-3`?

peachjam/static/stylesheets/_global.scss
```diff
@@ -91,7 +91,7 @@ footer {
 }
 
 .breadcrumb {
-  margin-top: map-get($spacers, 1);
+  margin: 0;
```
longhotsummer, line 94: LOL, maybe just set the spacing here and drop `mt-3` on each `.breadcrumb`?

peachjam/templates/peachjam/layouts/document_detail.html
musangowope: image
This was an escaping margin-top, that made a gap between the main and body elements. Super strange.
`d-flex` worked to fix this issue, but actually using `py-4` on `h1` is a clearer solution.

musangowope: It seems to be because of collapsing margins 🤔
https://stackoverflow.com/questions/17137262/why-does-the-child-divs-margin-affect-the-margin-of-the-parent

peachjam/static/stylesheets/_global.scss
musangowope: done. Used padding-top instead of margin-top because margin collapsing issue

lawlibrary/templates/lawlibrary/judgment_list.html
```diff
@@ -6,7 +6,7 @@
 {% block page-content %}
   <section class="pb-5">
     <div class="container">
-      <h1 class="mb-4 mt-2">{% trans 'Judgments' %}</h1>
+      <h1 class="mb-3 mt-4">{% trans 'Judgments' %}</h1>
```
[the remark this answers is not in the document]
musangowope: Done, using `my-4` for headings

peachjam/templates/peachjam/layouts/document_list.html
musangowope: done. Used my-4

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -4,45 +4,47 @@
 {% block title %}{{ document.title }}{% endblock %}
 
 {% block page-content %}
-  <main class="pb-5">
-    <div class="container">
-      {% block breadcrumbs %}
-      {% endblock %}
+  <div class="pb-5">
```
musangowope, line 7: page-content was already wrapped in a `main` element tag, so we had a nested `main`, generally isnt semantically correct

longhotsummer (approved):

PR #487, opened 2022-09-30
Work and relationship admin

longhotsummer:
* do away with the custom relationship view and just edit via work inlines
* works are managed automatically and so are readonly

image

image

image


actlikewill (approved):
Looks Good

Issue #486, opened 2022-09-29
Key-value pair spacing in metadata block


longhotsummer:
https://github.com/laws-africa/peachjam/pull/476 made the key-value pair block take up the full width if there was only one column. However, now we have the case where the key-value pairs are too far apart.

I think we should have some custom CSS to make the key-value pair definition lists work "well" out the box (eg. add a specific class on the `dl` and just have plain `dd/dt` tags). I suggest that the `dt` takes up 1/3 of the space with some specific max width, and then the `dd` takes up the rest of the space.

That'll mean it condenses well on mobile, but stops expanding on wide single-column displays.

longhotsummer:
Specific column widths are still showing up in lawlibrary; eg https://lawlibrary.org.za/akn/za/judgment/zacc/2022/35/eng@2022-09-29

![image](https://user-images.githubusercontent.com/4178542/194007291-d3828991-3070-4fd5-8248-a65c7cc097f4.png)

PR #489, opened 2022-09-30
Key pair meta block spacing

musangowope:
https://www.loom.com/share/c092e711d6b74bdab1514dbb2d92a75a

resolves https://github.com/laws-africa/peachjam/issues/486

### DIFF
--- a/peachjam/static/stylesheets/components/_document-content.scss
+++ b/peachjam/static/stylesheets/components/_document-content.scss
@@ -298,3 +298,15 @@
 .document-metadata-list {
   margin-bottom: 0px;
 }
+
+.document-metadata-list__row {
+  display: flex;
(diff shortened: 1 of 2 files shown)


peachjam/static/stylesheets/components/_document-content.scss
```diff
@@ -298,3 +298,15 @@
 .document-metadata-list {
   margin-bottom: 0px;
 }
+
+.document-metadata-list__row {
```
longhotsummer, line 302: `div` can't be a child of `dl`. Just move these styles into `.document-metadata-list`?

longhotsummer (commented):
I suspect lawlibrary and liiweb have additional templates that will need to be updated.

longhotsummer (approved):

PR #504, opened 2022-10-03
Document Inheritance via Django-Polymorphic

### LINKED ISSUE #432: Use django-polymorphic for document inheritance
We have a problem right now that when you load a document object via `CoreDocument.objects`, you only ever get the `CoreDocument` object and not the document subclass instance, such as `Judgment`. This means that we don't get to take advantage of polymorphism to override the behaviour of `CoreDocument` at certain times.

For example, a `Judgment` instance should be able to override properties and methods on `CoreDocument` AND have those changes available when we're updating the elasticsearch index.

django-polymorphic changes this and very neatly ensures that `CoreDocument.objects.all()` (and similar queries) returns the correct polymorphic object.

* install https://django-polymorphic.readthedocs.io/en/stable/
* migrate existing models (see https://django-polymorphic.readthedocs.io/en/stable/migrating.html)
* migrate existing model managers

ro6ley:
This PR adds support for django-polymorphic for document inheritance, w/ migrations.

closes #432


longhotsummer (approved):

actlikewill (changes requested):
@ro6ley You can update the migrations. Otherwise it works well.

longhotsummer (approved):

Issue #321, opened 2022-08-16
Make anchor text the same color as notice text


musangowope:
![image](https://user-images.githubusercontent.com/15995210/184874233-ca7b357e-6cd6-4df6-bf77-7898e7222c85.png)

_Originally posted by @longhotsummer in https://github.com/laws-africa/peachjam/issues/307#issuecomment-1216587529_

PR #505, opened 2022-10-03
Match  anchor color to their type respective type of alert

musangowope:
Bootstrap has a class called `.alert-link` which is used for links inside `.alert`so apply that styling of `.anchor-link` to change child anchor elements of `.alert`

DEMO:
image

resolves https://github.com/laws-africa/peachjam/issues/321


longhotsummer (approved):
nice

PR #513, opened 2022-10-05
Replace `JudgmentMediaSummaryFile` model with `AttachedFiles` model

nickmwangemi:
image


peachjam/models/judgment.py
```diff
@@ -191,18 +191,14 @@ def save(self, *args, **kwargs):
         return super().save(*args, **kwargs)
 
 
-class JudgmentMediaSummaryFile(AttachmentAbstractModel):
-    SAVE_FOLDER = "media_summary_files"
+class AttachedFileNature(models.Model):
```
longhotsummer, line 194: This can move into `peachjam/models/core_document_model.py` since it's no longer judgment-specific.

longhotsummer: Ditto for AttachedFiles

peachjam/models/judgment.py
```diff
@@ -191,18 +191,14 @@ def save(self, *args, **kwargs):
...
-    updated_at = models.DateTimeField(auto_now=True)
+    def __str__(self):
+        return self.name
 
-    class Meta:
-        ordering = ["document", "filename"]
 
-    def __str__(self):
-        return f"{self.filename}"
+class AttachedFiles(AttachmentAbstractModel):
```
longhotsummer, line 202: Do you need to override `SAVE_FOLDER = "attachments"`?

longhotsummer:
This is looking good, Nick. 

The attached files must be added to the metadata block, something like this:

image

peachjam/models/judgment.py
```diff
@@ -204,20 +203,3 @@ def get_case_number_string(self):
     def save(self, *args, **kwargs):
         self.string = self.get_case_number_string()
         return super().save(*args, **kwargs)
-
-
-class JudgmentMediaSummaryFile(AttachmentAbstractModel):
```
longhotsummer, line 209: Note to self: none of our databases have data for this model so we don't need a migration.

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -145,6 +145,25 @@ <h1 class="py-4">{{ document.title }}</h1>
                           </dd>
                         {% endif %}
                       {% endblock %}
+
+                      {% block document-metadata-media-summary-download %}
+                        {% if document.attachedfiles_set %}
+                          <dt>{% translate 'Attachments' %}</dt>
```
longhotsummer, line 151: I'd like this done a bit differently, please. The user shouldn't have to understand our concept of attachments. Rather, show each attached file as its own metadata entry.

```
Language          English
Media summary     Download PDF (3.2MB)
Media summary     Download DOCX (1MB)
Original document Download RTF (1.2MB)
  
```

longhotsummer (changes requested):
Minor change to layout, pls.

longhotsummer:
@actlikewill if https://github.com/laws-africa/peachjam/pull/518 is merged first then this will need to be updated

peachjam/models/judgment.py
nickmwangemi: @longhotsummer Does that mean I remove the migrations file?

longhotsummer: No, the model  migrations are needed, but we don't need to migrate old data (you don't appear to have added that in any case). So It's all good.

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -145,6 +145,33 @@ <h1 class="py-4">{{ document.title }}</h1>
                           </dd>
                         {% endif %}
                       {% endblock %}
+
+                      {% block document-metadata-media-summary-download %}
+                        {% if document.attachedfiles_set %}
+                          {% for attachedfile in document.attachedfiles_set.all %}
+                            {% if attachedfile.nature.name == 'Media summary' %}
+                              <dt>{% translate 'Media summary' %}</dt>
```
longhotsummer, line 153: No need to special case anything, just use `attachedfile.nature.name` as the `dt` value.

longhotsummer (commented):
Great, looks good. Please coordinate with @actlikewill -- one of you needs to update the now outdated use of JudgmentMediaSummaryFile elsewhere in the code (particularly the importer).

longhotsummer (approved):

nickmwangemi:
I've updated and tested the new `AttachedFiles` model together with @actlikewill, and the importer works.

PR #514, opened 2022-10-05
Metadata block refactor

musangowope:
https://www.loom.com/share/2989392353fe4d5b9c21699b8ea7a774

- Increased width of dt slightly so Media citation title text would not wrap
- removed uneccesary col on `dt` and `dd` elements
- Added mobile styling to decrease `dt` column slightly

Used a fixed width value that makes the key value space look close enough

resolves https://github.com/laws-africa/peachjam/issues/486


longhotsummer (approved):
good trade-off

PR #515, opened 2022-10-05
Use polymorphic documents for elasticsearch

longhotsummer:
This does:

* because we now use the polymorphic object managers, we don't need to jump through hoops with elasticsearch and can just rely on inheritance
* when accessing fields specific to CoreDocument subtypes, we can rely on the fact that ES will ignore missing attributes and just return None
* use overrides to change the "doc_type" label for GenericDocument to "Document" rather than "GenericDocument"
* correctly return `author` in the serializer, not `authoring_body`

### DIFF
--- a/peachjam/models/generic_document.py
+++ b/peachjam/models/generic_document.py
@@ -29,6 +29,9 @@ class GenericDocument(CoreDocument):
     def __str__(self):
         return self.title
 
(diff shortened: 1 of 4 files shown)


nickmwangemi (approved):

actlikewill (approved):
Works well.

Issue #493, opened 2022-10-03
Search result must include a document type (eg. judgment, legislation, etc.)


longhotsummer:
* [x] ensure this in indexed by elasticsearch and returned from the search api
* [x] show it in the search result item summary

Issue #429, opened 2022-09-18
As a user of search, I would like to easily see the Author information on individual search results


NotoriousMBB:
without having to open the link to check the document
![image](https://user-images.githubusercontent.com/46931873/190914836-c5d6eba8-5609-40f1-86f2-4b46ce4085e2.png)

longhotsummer:
Blocked by #432

nickmwangemi:
@longhotsummer Could you please provide more context?

longhotsummer:
I suggest doing this:

To the right of the date, add `item.court` if it is available, in the same way we add `item.author`.

![image](https://user-images.githubusercontent.com/4178542/194037371-8da6512d-9d5a-4913-95e1-396689949ed3.png)

Look in SearchResult.vue

longhotsummer:
Merge https://github.com/laws-africa/peachjam/pull/515 first, however, to fix the serialising of the author.

longhotsummer:
Can you also do https://github.com/laws-africa/peachjam/issues/493 at the same time -- put the document type `item.doc_type` immediately to the right of the date. Both `item.date` and `item.doc_type` will never be null.

longhotsummer:
Just to call out that `author` is actually there, but #515 fixes a serialiser bug. So this issue is now about also showing the `court` information, in addition to the author :)

PR #517, opened 2022-10-05
Search result improvements

nickmwangemi:
image


peachjam/js/components/FindDocuments/SearchResult.vue
```diff
@@ -15,7 +15,11 @@
       <i>{{ item.citation }}</i>
     </div>
     <div class="text-muted">
-      {{ item.date }}
+      {{ item.date }} &nbsp; {{ item.doc_type }}
```
longhotsummer, line 18: Please can you put the doc type in a `<span class="ms-3">` rather than using the nbsp, so that there is consistent spacing between the items.

longhotsummer: I'm also not sure why we're using `strong` below 🤷  can you change both `strong`s to `span`s.

longhotsummer: less distracting formatting :)

longhotsummer:
Closes #493 
Closes #429

PR #518, opened 2022-10-05
Judgment  import resource update

actlikewill:
- Standardize the judgments bulk import resource esp. as regards to case numbers.

### DIFF
--- a/peachjam/resources.py
+++ b/peachjam/resources.py
@@ -1,5 +1,6 @@
 import logging
 import mimetypes
+import re
 
 import magic
 from cobalt import FrbrUri
@@ -17,6 +18,7 @@ from peachjam.models import (
     GenericDocument,
     Judge,
     Judgment,
+    JudgmentMediaSummaryFile,
     Locality,
     MatterType,
     SourceFile,
@@ -161,16 +163,27 @@ class JudgmentResource(BaseDocumentResource):
                 defaults={"name": row["court_obj"]["name"]},
             )
 
+        source_files = row["source_url"].split("|")
+        docx = re.compile(r".docx?$")
+        pdf = re.compile(r".pdf$")
+        for file in source_files:
+            match_docx = docx.search(file)
+            match_pdf = pdf.search(file)
+            if match_docx:
+                row["source_url"] = file
+            elif match_pdf:
+                row["source_url"] = file
+
     def after_import_row(self, row, instance, row_number=None, **kwargs):
         super().after_import_row(row, instance, row_number, **kwargs)
 
         judgment = Judgment.objects.get(pk=instance.object_id)
 
-        if "string_override" in row:
+        if row["case_string_override"]:
             CaseNumber.objects.create(
-                string_override=row["string_override"], document=judgment
+                string_override=row["case_string_override"], document=judgment
             )
-        elif "case_numbers" in row:
+        elif row["case_numbers"]:
             for c in list(map(str.strip, row["case_numbers"].split("|"))):
 
                 case_number_values = c.split("/")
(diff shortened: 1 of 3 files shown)


peachjam/resources.py
```diff
@@ -129,12 +132,26 @@ def before_import_row(self, row, **kwargs):
         DocumentNature.objects.get_or_create(name=row["nature"])
 
 
+class JudgesWidget(ManyToManyWidget):
+    def clean(self, value, row=None, *args, **kwargs):
+        judges = list(map(" ".join, list(map(str.split, value.split(self.separator)))))
```
longhotsummer, line 137: I'm not clear what this is doing. Can you add a comment? Also, list comprehensions are generally easier to read and more pythonic than using `map`.

peachjam/resources.py
```diff
@@ -146,31 +163,51 @@ def before_import_row(self, row, **kwargs):
                 defaults={"name": row["court_obj"]["name"]},
             )
 
+        source_files = row["source_url"].split("|")
+        docx = re.compile(r".docx?$")
```
longhotsummer, line 167: escape `.` because it's special

peachjam/resources.py
```diff
@@ -146,31 +163,51 @@ def before_import_row(self, row, **kwargs):
...
+        pdf = re.compile(r".pdf$")
+        for file in source_files:
+            match_docx = docx.search(file)
+            match_pdf = pdf.search(file)
+            if match_docx:
+                row["source_url"] = file
+            elif match_pdf:
+                row["source_url"] = file
+
```
longhotsummer, line 176: this doesn't make any sense -- you're setting `source_url` to the same thing each time.

peachjam/resources.py
```diff
@@ -146,31 +163,51 @@ def before_import_row(self, row, **kwargs):
...
-                    matter_type, _ = MatterType.objects.get_or_create(
-                        name=row["matter_type"]
-                    )
-                CaseNumber.objects.create(
-                    document=judgment,
-                    number=case_number[0],
-                    year=case_number[1],
-                    matter_type=MatterType.objects.get(name=row["matter_type"]),
+        elif row["case_numbers"]:
+            for c in list(map(str.strip, row["case_numbers"].split("|"))):
```
longhotsummer, line 187: Can you include a comment showing the type of input you're expecting?

longhotsummer (commented):
I think it'd be useful to start adding tests for some of the stranger edge cases here.

If full round-trip tests (eg. CSV -> objects) are too complex, then split some of the crucial code into separate methods on the object, and test them directly. eg. the media summary file code can provide a row and get back an object, which is easier to write a unit test for.

peachjam/resources.py
actlikewill: ✔️

peachjam/resources.py
actlikewill: This is supposed to strip white space around and between the judges name to avoid the duplication issue we encountered. I have changed it to a list comprehension.

peachjam/resources.py
```diff
@@ -146,31 +163,51 @@ def before_import_row(self, row, **kwargs):
                 defaults={"name": row["court_obj"]["name"]},
             )
 
+        source_files = row["source_url"].split("|")
```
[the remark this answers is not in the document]
actlikewill: changed.

peachjam/resources.py
actlikewill: I have added  a break in the loop to terminate when it encounters a .docx file. Otherwise it will accept a pdf file if it doesn't find a docx.

peachjam/resources.py
actlikewill: added.

longhotsummer (approved):

PR #535, opened 2022-10-12
Readd lost  facet translations

musangowope:
- Wrap t functions around facets titles.


peachjam/js/components/FindDocuments/index.vue
```diff
@@ -110,6 +110,19 @@
         </div>
       </div>
     </div>
+
+    <!-- DOM Hack for i18next to parse facet to locale json. i18next skips t functions in script element -->
+    <div v-if="false">
+      {{ $t('Document type') }}
```
musangowope, line 116: `i18next` parser only parses t functions that are in the `template` tag of a vue file, hence we have this here as a dom hack to make sure that these translations are include in the locale json file when running the parse command.
Without it, Document type, Author, Jurisdiction etc would be removed from the locale files if we ran the parser command

longhotsummer (approved):

PR #537, opened 2022-10-12
handle case if langs found  from language select element

musangowope:
Currently for our js translations, we get the lang key codes for the Language Select Element in the DOM. We use these lang codes to then load the json translations files.

Lawlibrary does not have a Language Select Element, hence no translation json files are loaded. 
Resulting in this:
image
because the t function has no json file to reference

To handle this edge case we should simply load the `en.json`
After the fix:
image


peachjam/js/i18n.ts
```diff
@@ -17,9 +17,14 @@ const messages: {
   [key: string] : any
 } = {};
 
-langs.forEach(key => {
-  messages[key] = loadJSONFile(`${key}/translation.json`);
-});
+// If no languages found, load en/translation.json
```
longhotsummer, line 20: Why not just set the default on line 5 to `["en"]`?

longhotsummer (approved):

peachjam/js/i18n.ts
musangowope: done

PR #538, opened 2022-10-12
Fixture Based Tests

ro6ley:
This PR:
* [x] adds tests based on sample data fixtures.
* [x] configures isolated testing for each Django app


peachjam/tests/test_views.py
```diff
@@ -0,0 +1,114 @@
...
+        response = self.client.get("/accounts/login/")
+        self.assertTemplateUsed(response, "account/login.html")
+
+    def test_homepage(self):
+        response = self.client.get("/")
+        self.assertEqual(response.status_code, 200)
+
+        # documents
+        self.assertEqual(response.context_data["documents_count"], 8)
+        self.assertEqual(len(response.context_data["recent_judgments"]), 2)
```
longhotsummer, line 17: I really advise NOT asserting against the length of the list. When it fails, it's really hard to debug what is missing/in the list, because it just says "2 != 3".

Rather, assert that a particular string is in the page, or turn the list into a string of things (eg. titles) and use `self.assertEqual(["foo", "bar", "baz"], titles)` or similar.

longhotsummer (approved):
I'm not sure if you want to  split the tests out further, but I'm happy to merge this as a starting point and build on them incrementally.

PR #558, opened 2022-10-18
Custom date widget

actlikewill:
- Adds a custom date widget for documents


actlikewill:
@longhotsummer So its seems that html validation is not working because Django is setting `novalidate` on admin html forms. I could try adding some JavaScript to at least show error messages and/or prevent form validation. 

The integer widget is also *slightly* smaller than the select widget (which hurts my ocd), I can add some custom CSS for that.
date_widget

peachjam/admin.py
```diff
@@ -340,7 +378,7 @@ class JudgmentMediaSummaryFileInline(BaseAttachmentFileInline):
 
 
 class JudgmentAdminForm(DocumentForm):
-    hearing_date = forms.DateField(widget=forms.SelectDateWidget(), required=False)
+    hearing_date = forms.DateField(widget=DateSelectorWidget())
```
longhotsummer, line 381: I think this is still optional, and the required=False is still necessary?

peachjam/admin.py
```diff
@@ -115,13 +116,50 @@ class DocumentTopicInline(admin.TabularInline):
     extra = 1
 
 
+class DateSelectorWidget(forms.MultiWidget):
+    def __init__(self, attrs=None):
+        months = [x for x in enumerate(list(calendar.month_name))][1:]
```
longhotsummer, line 121: Can you check how SelectDateWidget handles the month names, because ideally they need to be translated.

peachjam/admin.py
```diff
@@ -115,13 +116,50 @@ class DocumentTopicInline(admin.TabularInline):
...
+                    "max": 31,
+                    "class": "vIntegerField",
+                    "pattern": "(3[01]|[12][0-9]|[1-9])",
+                }
+            ),
+            forms.Select(attrs=attrs, choices=months),
+            forms.NumberInput(
+                attrs={
+                    "placeholder": "Year",
+                    "min": 1910,
```
longhotsummer, line 136:
```suggestion
                    "min": 1500,
```

peachjam/admin.py
actlikewill: Right. changed

peachjam/admin.py
actlikewill: It uses the months list from `django.utils.date` module which wraps the names in translator tags. Very useful.

longhotsummer (approved):
Great. I've never coded a `MultiWidget` before and I'm pleased to see that it's pretty simple.

PR #571, opened 2022-10-20
Optimize legislation query

nickmwangemi:


peachjam/views/generic_views.py
```diff
@@ -20,6 +20,13 @@
 )
 
 
+class BaseQuerysetPrefetchMixin:
+    model = None
+
+    def get_base_queryset(self):
+        return self.model.objects.prefetch_related("author", "nature")
```
longhotsummer, line 27: You could also just set the queryset property on the two views impacted by this; two lines of code instead of a bunch. I also like that the prefetch then is clearly linked to the model

Eg

queryset = GenericDocument.objects.prefetch_related(...)

Up to you

longhotsummer (dismissed):

peachjam/views/generic_views.py
nickmwangemi: I've opted for your suggestion.

PR #576, opened 2022-10-21
codes for DocumentNature, use nature.name for user-friendly document type

### LINKED ISSUE #551: When showing a generic document, use the nature rather than "document"
The best place to do this is when indexing into ES. This is now easy because we have polymorphic models.

longhotsummer:
Fixes #551 

* add a code field to DocumentNature, which is used when looking it up from eg. from indigo or imports
* use the name when serialising the document type, eg. into elasticsearch
* we'll make the name human-friendly (ie. "Model law" rather than "model-law") manually after migration
* show document type in metadata block

image


actlikewill (approved):
Looks good.

Issue #427, opened 2022-09-18
Consistently apply the term Regional Body across the AGP website


NotoriousMBB:
as a label for Author, e.g. the metadata section of a treaty display page
![image](https://user-images.githubusercontent.com/46931873/190914404-194a831b-51ce-41a2-a325-5a1424b2af4d.png)

e.g. filter on search results page
![image](https://user-images.githubusercontent.com/46931873/190914464-a780910f-92cf-4d46-825f-776edd4405e4.png)

longhotsummer:
## Proposal

Using different terms for different database models seems like it will come up in a few places. Without trying to solve a problem that we don't know we have yet, I suggest the following to try out a potential path that will solve the current challenge and tell us if it's a reasonable way forward.

* On the `Author` class, add a variable (not a db field, just a normal class-level variable) called something like `model_label` that defaults to `_('Author')` (not `_` for translations)
* In AfricanLII's `ready()` we do something like `Author.model_label = _('Regional body')`
* That variable is available on the class and on objects where necessary
* Use that variable in templates
* For javascript facets, pass in a JSON `labels` dictionary that provides a translated label for the `Author` model. eg. `data-labels="{'Author': 'Regional body'}".

That way the translated and adjusted string is available to templates and javascript from a single source, and still gets translated.

PR #588, opened 2022-10-24
Add label for Author model

nickmwangemi:
**Meta block:**
Screenshot from 2022-10-24 12-14-50

**Search page:**
image

resolves https://github.com/laws-africa/peachjam/issues/427


africanlii/apps.py
```diff
@@ -1,4 +1,5 @@
 from django.apps import AppConfig
+from django.utils.translation import gettext as _
```
longhotsummer, line 2: I think you need `gettext_lazy` https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#lazy-translations so that the translation isn't applied when the `_` function is called, because we don't know what language to use yet.

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -79,7 +79,7 @@ <h1 class="py-4">{{ document.title }}</h1>
 
                       {% block document-metadata-content-author %}
                         {% if document.author %}
-                          <dt>{% trans 'Author' %}</dt>
+                          <dt>{% trans document.author.model_label %}</dt>
```
longhotsummer, line 82: The label is already being translated by `_` elsewhere so it doesn't need to be double translated.

peachjam/views/generic_views.py
```diff
@@ -75,12 +75,14 @@ def get_context_data(self, **kwargs):
                 ).values_list("date__year", flat=True)
             )
         )
+        labels = {"Author": "Regional body"}
```
longhotsummer, line 78: The point of this change is that agp will use `Regional body` but other sites like lawlibrary will use `Author`. So this won't work.

peachjam/views/generic_views.py
```diff
@@ -75,12 +75,14 @@ def get_context_data(self, **kwargs):
...
 
         context["facet_data"] = {
             "years": years,
             "authors": authors,
             "courts": courts,
             "alphabet": lowercase_alphabet(),
+            "labels": labels,
```
longhotsummer, line 85: How will Musango get hold of `labels` in javascript to use the labels in the search filter interface?

longhotsummer (changes requested):
The search filters also need to use the updated label. Musango can guide you as to where he needs to read it from -- probably a `data-labels` attribute somewhere.

longhotsummer (dismissed):
great

longhotsummer (approved):

PR #590, opened 2022-10-24
Optimize db query for /admin/peachjam/genericdocument/

nickmwangemi:


peachjam/admin.py
```diff
@@ -352,6 +352,11 @@ class GenericDocumentAdmin(ImportMixin, DocumentAdmin):
     fieldsets = copy.deepcopy(DocumentAdmin.fieldsets)
     fieldsets[0][1]["fields"].extend(["author", "nature"])
 
+    def get_queryset(self, request):
+        qs = super().get_queryset(request)
+        qs = qs.prefetch_related("author", "nature", "locality")
```
longhotsummer, line 357: The admin view always shows the locality and the jurisdiction (country) in the listing view. I think you should do a prefetch_related for those fields on CoreDocumentAdmin. Then this just adds "nature" and "author", which you should also do for LegalInstrumentAdmin below.

longhotsummer (approved):

PR #592, opened 2022-10-25
Hard-code search path

longhotsummer:
If these special templates are too complex, we run the risk of them failing when they're needed, particularly the 500.html


longhotsummer:
This is causing issues with gazettes.africa when there's a 404 but we don't have the search URLs installed.

PR #604, opened 2022-10-25
Allow xframes

nickmwangemi:
image

### DIFF
--- a/peachjam/settings.py
+++ b/peachjam/settings.py
@@ -432,3 +432,6 @@ else:
             "LOCATION": "/var/tmp/django_cache",
         },
     }
+
+# Override X-Frame-Options header value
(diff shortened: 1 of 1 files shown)


peachjam/settings.py
```diff
@@ -432,3 +432,6 @@
             "LOCATION": "/var/tmp/django_cache",
         },
     }
+
+# Override X-Frame-Options header value
+X_FRAME_OPTIONS = "ALLOW"
```
longhotsummer, line 437: This will allow any site to place our content in frames, which is a bit dangerous. I think `SAMEORIGIN` is a better idea -- only our pages can embed our content.  https://docs.djangoproject.com/en/3.2/ref/clickjacking/

longhotsummer (approved):

PR #608, opened 2022-10-27
Bulk Import resource tweaks

actlikewill:
- Adds Skip Widget for skipping rows
- Adds source file widget for validating source files
- Refactors getting case numbers and source file logic into class methods
- Fixes tests

### DIFF
--- a/peachjam/download.py
+++ b/peachjam/download.py
@@ -14,6 +14,7 @@ logger = logging.getLogger(__name__)
 
 
 def download_source_file(url):
+    url = urllib.parse.quote(url, safe=":/")
     # check if the url is a Google doc url
     google_regex = re.compile(r"google")
     if google_regex.search(url):
--- a/peachjam/download.py
+++ b/peachjam/download.py
@@ -55,12 +55,8 @@ def download_file_from_google(file_id):
 
 
 def download_with_urllib(url):
-    try:
-        f = NamedTemporaryFile()
-        r = requests.get(url)
-        r.raise_for_status()
-        f.write(r.content)
-        return f
-    except requests.exceptions.HTTPError as e:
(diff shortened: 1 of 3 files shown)


longhotsummer (commented):
Can we just use `requests.get`? It handles all the correct URL escaping. This feels way too much like we're having to care about things we shouldn't have to worry about.

peachjam/download.py
```diff
@@ -57,9 +57,10 @@ def download_file_from_google(file_id):
...
     try:
         f = NamedTemporaryFile()
-        urllib.request.urlretrieve(url, f.name)
-    except urllib.error.HTTPError as e:
+        r = requests.get(url)
+        r.raise_for_status()
+        f.write(r.content)
+        return f
+    except requests.exceptions.HTTPError as e:
         logger.error(e)
```
longhotsummer, line 65: Surely downloading the file is a key part of the import? If that fails, the import shouldn't succeed, I would think?

actlikewill: @longhotsummer I have been grappling with what to do here. The issue is that the whole import fails if only one file is missing and the error is not shown to the user. Should we fail the whole import if one file is missing?

peachjam/resources.py
```diff
@@ -44,6 +51,25 @@ def clean(self, value, row=None, *args, **kwargs):
         return None
 
 
+class SourceFileWidget(CharWidget):
+    def clean(self, value, row=None, **kwargs):
+        try:
+            print(value)
```
longhotsummer, line 57:
```suggestion

```

actlikewill:
@longhotsummer Changes made.

longhotsummer (approved):

PR #610, opened 2022-10-27
Dynamic storage

longhotsummer:
This PR adds a new `DynamicStorageFileField` model field that makes it possible to dynamically determine which storage is used for a particular file. Rather than writing just the raw filename to the database for the field, it writes something like `s3:by-bucket:path/and/filename.txt`. When reading the file, it dynamically determines that this file uses S3 storage with the bucket `my-bucket` and the object location `path/and/filename.txt`. Similarly, a file could be stored as `file:path/and/filename.txt` which just uses the default file system storage.

This deployment updates the file fields in the database to indicate that we're using S3 with a single bucket.

In future, we can have files that are stored in different S3 buckets, which is the whole goal of this change. It will, for example, allow Gazettes.Africa to store gazettes in an S3 bucket, and then other peachjam installations to read those files directly from S3, while storing their own files in a different S3 bucket.

Additionally, the S3 dynamic storage supports a `silent_readonly` flag which silently ignores write operations. This is useful if we don't want to modify files that another peachjam installation is responsible for maintaining.


ro6ley (approved):
LGTM!

nickmwangemi (approved):

PR #626, opened 2022-10-31
Multiple case string overrides

actlikewill:
- adds mutilple case string override in resource


longhotsummer (dismissed):
nice -- i didn't realise we already had tests, so this is great

actlikewill:
> @longhotsummer I added one more test that does the actual import

Oops!, Let me fix

longhotsummer (approved):

PR #635, opened 2022-11-02
Update queryset for legal instrument listing page

nickmwangemi:


peachjam/views/generic_views.py
```diff
@@ -35,7 +35,7 @@ def get(self, request, *args, **kwargs):
         return super(FilteredDocumentListView, self).get(request, *args, **kwargs)
 
     def get_base_queryset(self):
-        return self.model.objects.all()
+        return self.queryset
```
longhotsummer, line 38: This will break for the views that don't have `queryset` set, but instead use `model`. Do what django's views do: `return self.queryset or self.model.objects` -- that will use queryset if specified, otherwise fall back to the model queryset.

peachjam/views/legal_instrument.py
```diff
@@ -10,9 +10,10 @@ class LegalInstrumentListView(FilteredDocumentListView):
     model = LegalInstrument
     template_name = "peachjam/legal_instrument_list.html"
     navbar_link = "legal_instruments"
+    queryset = LegalInstrument.objects.prefetch_related("author", "nature")
 
     def get_queryset(self):
-        queryset = LegalInstrument.objects.prefetch_related("author", "nature")
+        queryset = super(LegalInstrumentListView, self).get_queryset()
         return queryset.order_by("-date")
```
longhotsummer, line 17: If you move the order_by into the `queryset` attrib, then `get_queryset` can fall away completely.

longhotsummer (approved):

PR #637, opened 2022-11-02
Resource validation

actlikewill:
- Adds validation for bulk import resources

### DIFF
--- a/peachjam/resources.py
+++ b/peachjam/resources.py
@@ -38,28 +38,55 @@ from .download import download_source_file
 logger = logging.getLogger(__name__)
 
 
-class JurisdictionWidget(ForeignKeyWidget):
+class CharRequiredWidget(CharWidget):
+    def __init__(self, field):
+        self.field = field
+
     def clean(self, value, row=None, *args, **kwargs):
-        if value:
-            return self.model.objects.get(iso__iexact=value)
-        return None
+        if not value:
+            raise ValidationError(f"{self.field} is required")
+        return value
 
 
-class AuthorWidget(ForeignKeyWidget):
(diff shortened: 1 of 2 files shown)


peachjam/resources.py
```diff
@@ -38,54 +38,93 @@
 logger = logging.getLogger(__name__)
 
 
-class JurisdictionWidget(ForeignKeyWidget):
+class CharRequiredWidget(CharWidget):
+    def __init__(self, field):
```
longhotsummer, line 42: Minor: you could turn this required functionality into a mixin, since it's repeated in about 4 places.

peachjam/resources.py
```diff
@@ -38,54 +38,93 @@
...
+        attribute="locality",
+        widget=ForeignKeyRequiredWidget(Locality, name="locality", field="code"),
+    )
+    source_url = fields.Field(
+        attribute="source_url", widget=SourceFileWidget(field="source_url")
     )
-    source_url = fields.Field(attribute="source_url", widget=SourceFileWidget())
     skip = fields.Field(attribute="skip", widget=SkipRowWidget())
 
+    required_fields = None
```
longhotsummer, line 126: `set(None)` raises an exception, so perhaps set this to `[]`

longhotsummer (dismissed):

peachjam/resources.py
actlikewill: Had the same thought but the issue is the implementation is slightly different for ForeignKeyWidget vs CharWidget, and other is a CustomWidget. :grimacing:

peachjam/resources.py
actlikewill: Changed. Thanks

longhotsummer (approved):

PR #648, opened 2022-11-07
Allows null values to Judgment fields

actlikewill:
- Allows null values for judgment fields.


peachjam/models/judgment.py
```diff
@@ -51,10 +51,10 @@ def __str__(self):
 
 class Judgment(CoreDocument):
     court = models.ForeignKey(Court, on_delete=models.PROTECT, null=True)
-    judges = models.ManyToManyField(Judge, blank=True)
-    headnote_holding = models.TextField(blank=True)
-    additional_citations = models.TextField(blank=True)
-    flynote = models.TextField(blank=True)
+    judges = models.ManyToManyField(Judge, null=True, blank=True)
```
longhotsummer, line 54: I don't think `null` has any effect on ManyToManyField. blank=true should be sufficient.

actlikewill:
@longhotsummer I noticed that Django included a migration for the DynamicStorageField for files. Is that ok?

longhotsummer:
> @longhotsummer I noticed that Django included a migration for the DynamicStorageField for files. Is that ok?

Hmm, I'm not sure. Let's see what the tests do. I don't see a reason for it not to work.

longhotsummer (approved):

PR #651, opened 2022-11-07
Handle ingestor 404 errors gracefully

nickmwangemi:


longhotsummer:
This will also squash 500 or 403 or other errors which we do want to know about. We only want to hide 404 errors.

peachjam/adapters/adapters.py
```diff
@@ -106,7 +106,12 @@ def update_document(self, url):
 
         logger.info(f"Updating document ... {url}")
 
-        document = self.client_get(f"{url}.json").json()
+        try:
+            document = self.client_get(f"{url}.json").json()
+        except requests.HTTPError as e:
+            logger.error(f"An error has occurred: {e}")
+            return
+
```
actlikewill, line 114: You can try it like this:
```suggestion
            if e.response.status_code == 404:
                return
            else:
                raise e          

```

longhotsummer (approved):

PR #653, opened 2022-11-07
replace colons rathing than using slugify

longhotsummer:
otherwise slashes and dots are removed. This can safely be done multiple times.


longhotsummer:
@actlikewill i originally changed this everywhere, but after review I think slugify when dealing with the title makes sense. this is only necessary here.

actlikewill (approved):
Looks good.

PR #655, opened 2022-11-08
Set up SenLII

nickmwangemi:


longhotsummer (commented):
I suggest also adding a deployment for this in the github workflows.

longhotsummer (approved):

PR #656, opened 2022-11-08
Indigo webhook

actlikewill:
- Adds an endpoint to update a document using the indigo ingestor.

### DIFF
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -329,8 +329,16 @@ class IndigoAdapter(Adapter):
             )
 
     def delete_document(self, expression_frbr_uri):
-        document = CoreDocument.objects.filter(
-            expression_frbr_uri=expression_frbr_uri
-        ).first()
-        if document:
-            document.delete()
+        url = re.sub(r"/akn.*", expression_frbr_uri, self.url)
+        try:
+            document = self.client.get(url)
+        except requests.exceptions.HTTPError as e:
+            if e.response.status_code == 404:
+
+                document = CoreDocument.objects.filter(
+                    expression_frbr_uri=expression_frbr_uri
+                ).first()
+                if document:
+                    document.delete()
+            else:
+                raise e
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -331,7 +331,7 @@ class IndigoAdapter(Adapter):
     def delete_document(self, expression_frbr_uri):
         url = re.sub(r"/akn.*", expression_frbr_uri, self.url)
         try:
-            document = self.client.get(url)
+            document = self.client_get(url)
         except requests.exceptions.HTTPError as e:
             if e.response.status_code == 404:
 
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -10,7 +10,7 @@ from dateutil import parser
 from django.core.files import File
 from django.utils.text import slugify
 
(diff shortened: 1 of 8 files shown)


peachjam/tasks.py
```diff
@@ -28,6 +28,29 @@ def update_document(ingestor_id, document_id):
...
 
 
+@background(queue="peachjam", remove_existing_tasks=True)
+def delete_document(ingestor_id, expression_frbr_uri):
+    from peachjam.models import Ingestor
+
+    ingestor = Ingestor.objects.filter(pk=ingestor_id).first()
+
+    if not ingestor:
+        log.info(f"No ingestor with id {ingestor_id:} ")
```
longhotsummer, line 38: Typo in the expression

peachjam/adapters/adapters.py
```diff
@@ -327,3 +327,10 @@ def download_source_file(self, url, doc, title):
                     "mimetype": magic.from_file(f.name, mime=True),
                 },
             )
+
+    def delete_document(self, expression_frbr_uri):
+        document = CoreDocument.objects.filter(
```
longhotsummer, line 332: I suggest first calling the indigo api and seeing if you get a 404 for the document, and only deleting it if you do. 

The reason for that is indigo sometimes deletes a document to unpublish it and then republishes it a little while later. If for some reason the deletion background task gets delayed, it could delete a document it shouldn't delete. 

So a good rule of thumb when deleting a document when you aren't the authoritative source is to wait a bit (eg 5 mins), double check it doesn't exist, and then delete. Distributed systems are fun :)

peachjam/tasks.py
```diff
@@ -28,6 +28,29 @@ def update_document(ingestor_id, document_id):
     log.info("Ingestor is disabled, ignoring.")
 
 
+@background(queue="peachjam", remove_existing_tasks=True)
```
longhotsummer, line 31: Put a delay on this task of a few minutes (see comment above about delayed deletion).

peachjam_api/views.py
```diff
@@ -31,3 +35,26 @@ class CitationLinkViewSet(viewsets.ModelViewSet):
 class LegislationViewSet(viewsets.ModelViewSet):
     queryset = Legislation.objects.all()
     serializer_class = LegislationSerializer
+
+
+class IngestorWebhookView(APIView):
+    authentication_classes = [authentication.TokenAuthentication]
+    permission_classes = []
```
longhotsummer, line 42: Maybe coredocument perms like at least update and delete?

longhotsummer (changes requested):
Great, small suggestions.

peachjam/tasks.py
```diff
@@ -28,6 +28,29 @@ def update_document(ingestor_id, document_id):
     log.info("Ingestor is disabled, ignoring.")
 
 
+@background(queue="peachjam", schedule=(60 * 60 * 5), remove_existing_tasks=True)
```
longhotsummer, line 31: 5 hours?? Maybe `60 * 5` is better, 5 mins?

actlikewill: :laughing:

peachjam/adapters/adapters.py
actlikewill: I've added the 404 check.

peachjam/tasks.py
actlikewill: Set the delay to 5 minutes.

longhotsummer (approved):

PR #670, opened 2022-11-11
Alternative Names for Documents

### LINKED ISSUE #546: As a user searching using alternative citation formats, I want to find the relevant document
For example, some judgments have citations assigned by other publishers which are well-known, and we should support indexing and display of those.

See also #616

### LINKED ISSUE #616: As a user of a peachjam website, I would like to discover with ease documents that are usually referred to by their popular names or abbreviations
Legal instruments and other documents in the AU ecosystem are frequently referred to by their abbreviations or short names. For example:

* the African Charter on Human and Peoples' Rights is known as the Banjul Charter
* the African Charter on Democracy and Elections is referred to as the ACDEG
* the Protocol to the African Charter on Human and Peoples' Rights on the Rights of Women in Africa is better know as the Maputo Protocol.

References in resolutions, decisions and policy documents would often use these popular references instead of the full title of the document.

Between what we can do on the Indigo Platform and the search engine of the AGP website, what would be the best way to facilitate the users' discovery of these materials?

ro6ley:
This PR adds support for alternative names for documents:
* [x] New `AlternativeModel` w/ migration
* [x] Updated search document
* [x] Search by alternative titles

### Screenshots
* **Aliases in document overview:**
  Screenshot 2022-11-14 at 09 40 00

* **Alternative names in admin section:**
  Screenshot 2022-11-11 at 16 05 58

* **Searching document by alternative names:**
  image

closes #616 
closes #546


peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -77,6 +77,17 @@ <h1 class="py-4">{{ document.title }}</h1>
                         {% endif %}
                       {% endblock %}
 
+                      {% block document-metadata-content-alternative-names %}
+                        <dt>{% trans 'Alternative names' %}</dt>
+                        <dd>
+                          {% for alternative_name in document.alternative_names.all %}
```
longhotsummer, line 83: I think each should be in their own div, so they're on a separate line, since it may not be clear where one ends and other starts otherwise.

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -77,6 +77,17 @@ <h1 class="py-4">{{ document.title }}</h1>
                         {% endif %}
                       {% endblock %}
 
+                      {% block document-metadata-content-alternative-names %}
+                        <dt>{% trans 'Alternative names' %}</dt>
```
longhotsummer, line 81: Most documents won't have alternative names, so I think this should not show in the metadata block if there aren't any.

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -77,6 +77,19 @@ <h1 class="py-4">{{ document.title }}</h1>
                         {% endif %}
                       {% endblock %}
 
+                      {% block document-metadata-content-alternative-names %}
+                        {% with document.alternative_names.all as alternative_names %}
+                          {% if alternative_names %}
+                            <dt>{% trans 'Alternative names' %}</dt>
+                            <dd>
```
longhotsummer, line 84:
```suggestion
                            <dd class="text-muted">
```

longhotsummer: We should either do this in CSS for all the metadata, or stop doing it completely. But for now, let's be consistent.

longhotsummer (approved):

PR #672, opened 2022-11-14
Entity profile front end

musangowope:
Demo:
https://www.loom.com/share/3adbd045e9144599b6ca26f9ba8e5a95

Added entity profiles to the following pages:
- court detail
- author detail
- taxonomy detail

### DIFF
--- a/peachjam/templates/peachjam/_entity_profile.html
+++ b/peachjam/templates/peachjam/_entity_profile.html
@@ -2,9 +2,12 @@
 
 <div class="entity-profile">
   <div class="entity-profile__background-photo">
-    <img src="{{ entity_profile.background_photo.url }}"
-         alt="Court background image"
-    />
+    {% if entity_profile.background_photo %}
+      <img src="{{ entity_profile.background_photo.url }}"
+           alt="Court background image"
+           loading="lazy"
+      />
+    {% endif %}
   </div>
 
   <div class="entity-profile__body">
@@ -12,10 +15,11 @@
       <img src="{{ entity_profile.profile_photo.url }}"
            class="entity-profile__body__profile-image"
            alt="profile image"
+           loading="lazy"
       />
     {% endif %}
     <div class="entity-profile__body__info">
-      <h1>{{ court }}</h1>
+      <h1>{{ entity_profile_title }}</h1>
 
       {% if entity_profile.about_html %}
         <div class="mb-2">{{ entity_profile.about_html }}</div>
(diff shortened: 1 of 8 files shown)


peachjam/templates/peachjam/court_detail.html
```diff
@@ -7,13 +7,67 @@
...
+                    <strong>{% trans 'Physical address:' %}</strong>
+                  </div>
+                  <div>{{ entity_profile.address }}</div>
+                </div>
+              {% endif %}
+
+
+              {% if entity_profile.website_url %}
+                <div class="col-12 col-sm-6">
+                  <div><strong>{% trans 'Visit website:' %}</strong></div>
```
longhotsummer, line 62: I noticed that `:` crept in for advanced search. We don't use them anywhere else AFAIK, I think we shouldn't start introducing them. The bold and positioning is sufficient.

musangowope: removed

musangowope:
@longhotsummer 
is the reason why we wrap the `court_detail.html` in a `blocktrans`
https://github.com/laws-africa/peachjam/blob/fc04f25265b062f8bc41e14b6cac2cab29799465/peachjam/templates/peachjam/court_detail.html#L28
is the format for the a court title different depending on the language?

peachjam/templates/peachjam/_entity_profile.html
```diff
@@ -0,0 +1,50 @@
+{% load i18n %}
+
+<div class="entity-profile">
+  <div class="entity-profile__background-photo">
+    {% if entity_profile.background_photo %}
+      <img src="{{ entity_profile.background_photo.url }}"
+           alt="Court background image"
```
longhotsummer, line 7:
```suggestion
           alt="background image"
```

longhotsummer:
@musangowope this looks awesome.

My only concern is how/when we repeat the title above the document list.

* author: "Documents by XXX" -- makes sense
* taxonomy: "Documents by Air Pollution" -- doesn't make sense
* court AND year: "Constitution Court - 2022" in title, then "Documents by Constitution Court" -- no year :/

Is it simpler to just effectively duplicate the heading if there's an entity profile block, and drop the "Documents by"? For the court and year case, the entity title for the court wouldn't have the year, but then the list above the documents would.

musangowope:
Changes made:
- Authors still use Documents by
- taxonomy detail and court detail use the same title as their entity profile blocks and drop (Documents by)

PR #687, opened 2022-11-16
taxonomy tree improvements + toc translations

musangowope:
Translation tweaks:
- Update law-widgets so Expand and collapse button text can be customizable
- setup 18next for translations of ts/js files
- Add translations for toc text

Visual tweaks
- move the top level entry into the box as the heading, to give the list a bit more room on the left because the tree is no longer rooted at the top-level
- restrict the height of the list and scroll when it overflows
- make wrapped lines more readable across all la-table-of-contents by resetting line height to normal on la-toc-item and adding some margin below each la-toc-item (currently wrapped lines and separate items are indistinguishable). This can happen globally across the site for all uses of this component.
- Add searching (usual TOC component search) 
**Visual change:**
image


longhotsummer:
This looks good, only issue is that the taxonomy search placeholder says "Search the table of contents" -- maybe just make that "Search"?

peachjam/templates/peachjam/_taxonomy_tree_facet.html
```diff
@@ -2,9 +2,9 @@
 
 <div class="list-group mb-4">
   <div class="list-group-item bg-light">
-    <strong>{% trans 'Subject areas' %}</strong>
+    <strong>{{ first_level_taxonomy }}</strong>
   </div>
-  <div class="list-group-item">
+  <div class="list-group-item" style="max-height: 500px; overflow-y: auto">
```
longhotsummer, line 7: Elsewhere we use a `vh` max-height (eg. `25vh` on `facets-scrollable`), since the vertical size of the screen is unknown. This means we can guarantee that it never goes outside of the visible screen area (particularly on mobile). I think it would be useful to be consistent -- eg `50vh` if we think it deserves more vertical space.

I think this should be a class (`taxonomy-tree-list-group`?) rather than inline styles.

longhotsummer (changes requested):
This looks great, just a minor suggestion -- thoughts welcome.

longhotsummer (dismissed):
great!

longhotsummer (dismissed):

PR #696, opened 2022-11-22
Fix existing doc

### LINKED ISSUE #621: AttributeError: 'NoneType' object has no attribute 'doc_type'
Sentry Issue: [AGP-26](https://sentry.io/organizations/lawsafrica/issues/3624412398/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'doc_type'
  File "background_task/tasks.py", line 43, in bg_runner
    func(*args, **kwargs)
  File "peachjam/tasks.py", line 20, in update_document
    ingestor.update_document(document_id)
  File "peachjam/models/ingestors.py", line 46, in update_document
    adapter.update_document(document_id)
  File "peachjam/adapters/adapters.py", line 183, in update_document
    f"The document {expression_frbr_uri} already exists as {existing_doc.doc_type}"
```

actlikewill:
closes #621

### DIFF
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -187,7 +187,7 @@ class IndigoAdapter(Adapter):
         ).first()
         if core_doc and not existing_doc:
             raise Exception(
-                f"The document {expression_frbr_uri} already exists as {existing_doc.doc_type}"
+                f"The document {expression_frbr_uri} already exists as {core_doc.doc_type}"
                 f" but not as {model}. Delete the existing document (CoreDocument #{core_doc.pk})."
             )
 


nickmwangemi (commented):
Since we are checking that the ingested document may be in the database, but not as the right document type, what would be the implication of having it as below?

```python
  if core_doc and existing_doc:
      raise Exception(
             f"The document {expression_frbr_uri} already exists as {existing_doc.doc_type}"
             f" but not as {model}. Delete the existing document (CoreDocument #{core_doc.pk})."
            )
```

actlikewill:
@nickmwangemi The existing document may be a different doc_type. So the logic checks if the document exists as a different doc_type. If the existing_doc is found it is okay since we will perform a get or create. If the document is found in the coredocument but not in existing_doc, we know that it exists as a different doc_type and we raise the exxeption to stop the ingestion.

longhotsummer (approved):

PR #698, opened 2022-11-22
Show only applicable languages

### LINKED ISSUE #177: In document lists, show only applicable languages
In document listing views, we currently show all versions of a document. This can lead to the same document listed a few times, once for each language.

Questions:

* should we show a document only once? If so, what language do we choose?
* if there is no translated document for the current site language, do we fall back to English?
* should translated documents have translated titles?

nickmwangemi:
Demo:
https://www.loom.com/share/80098a0387f04081936e8c3f76e8ad88

Tooltip for showing doc is available in a number of language versions:
https://www.loom.com/share/fbb8cfa999f441a7a9243d9544b1aeeb

Fixes #177


peachjam/views/generic_views.py
```diff
@@ -34,8 +36,12 @@ def get(self, request, *args, **kwargs):
 
         return super(FilteredDocumentListView, self).get(request, *args, **kwargs)
 
+    def get_language(self):
+        language = get_language_from_request(self.request)
+        return Language.objects.get(iso_639_1__iexact=language).iso_639_3
+
     def get_base_queryset(self):
-        return self.queryset or self.model.objects
+        return self.model.objects.preferred_language(self.get_language())
```
longhotsummer, line 44: What about the subclasses that rely on specifying `queryset`?

peachjam/signals.py
```diff
@@ -39,3 +42,10 @@ def bg_task_finished(sender, **kwargs):
     if transaction:
         # fake an exit from the transaction context
         transaction.__exit__(None, None, None)
+
+
+@receiver(signals.post_save, sender=CoreDocument)
+def update_language(sender, instance, **kwargs):
+    if not kwargs["raw"]:
+        instance.work.languages = [instance.language.iso_639_3]
```
longhotsummer, line 50: What if the work has now has three different documents all in different languages?

longhotsummer (changes requested):
It's looking good!

Two suggestions:

1. we'll need a data migration to bulk populate the new field for the existing works
2. I think the "multi language" icon needs to be more subtle. Perhaps just the icon with a tooltip - @musangowope what do you think?

peachjam/templates/peachjam/_document_table.html
```diff
@@ -20,7 +20,15 @@
           {% else %}
             <td >{{ document.author|default_if_none:'' }}</td>
           {% endif %}
-          <td style="white-space: nowrap;">{{ document.date }}</td>
+            <td style="white-space: nowrap;">{{ document.date }}</td>
+
+            <td style="white-space: nowrap;">
+              {% if document.work.languages|length > 1 %}
+                <span class="d-inline-block" tabindex="0" data-toggle="tooltip" title="Multi-language">
```
longhotsummer, line 27: The string must be translatable - "a multiple languages available" is more descriptive, too.

peachjam/migrations/0040_backfill_work_languages.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.2.16 on 2022-12-01 11:54
+
+from django.db import migrations
+
+
+def backfill_work_languages(apps, schema_editor):
+    CoreDocument = apps.get_model("peachjam", "CoreDocument")
+    for doc in CoreDocument.objects.all():
```
longhotsummer, line 8: This will load all fields of all documents into memory. On large collections this is huge. Instead, use .iterator(100) to do only 100 at a time. 

We've learnt the hard way to batch it like this!

longhotsummer (changes requested):

longhotsummer (approved):

musangowope:
resolves https://github.com/laws-africa/peachjam/issues/177

Issue #668, opened 2022-11-10
Chore: ensure compatibility with pocketlaw


longhotsummer:
* [x] what external domains must be included in a crawl (eg. google fonts)
* [x] how do we determine what files to download / not download? (eg. gazette PDFs)
* [x] ensure we only scrape in the default language (languages are set with a POST, so this should be fine)

## outstanding
* [x] how does `wget` get all links to follow for legislation, given that the legislation list is rendered by Javascript?
* [x] integrate with search

## Listing pages

Some functionality on the document listing pages make the number of static pages we must scrape explode. All the facet options (eg. alphabetical, years, judge) will do this.

Instead, we can disable facets in pocketlaw. I suggest we do this by injecting a CSS class onto the body of the document in pocketlaw, eg `class="is-pocketlaw"` and use CSS inside peachjam to hide those features we know do not work on pocketlaw. We could even go further and include `class="is-pocketlaw is-pocketlaw-offline"` and only disable them when the page is being served offline.

longhotsummer:
## Legislation links

wget naively loads the page and follows links. Because JS is used to render the legislation tables after loading, the page doesn't actually contain links and so wget won't download them.

Options

1. a comprehensive sitemap page with all content (which is also useful for other search engines) -- we need to explicitly tell wget to download the xml and then fetch all the links, so it's not a very organic solution
2. include a simple html table of links in the legislation page which is then replaced by javascript with the richer table. This has the benefit of being a nice fallback if JS doesn't load (not a very common case) and integrates easily with search engines and with wget, at the cost of a larger page.

ro6ley:
Test lawlibrary crawl -> https://www.loom.com/share/1617dabfd6a14f64abc14b9e7139ad8e 

Notes:
- Gazettes are hosted externally, so weren't part of this crawl
- Crawl was executed recursively up to only 2 levels, hence some broken content links
- Some facets work, some don't

ro6ley:
Related PRs:
- Warc processor update to handle peachjam warcs [pocketlaw-automation #4](https://github.com/laws-africa/pocketlaw-automation/pull/4)
- Pocketlaw update to handle peachjam warcs [pocketlaw #452](https://github.com/laws-africa/pocketlaw/pull/452)

ro6ley:
[...]

PR #704, opened 2022-11-23
Tweaks for Pocketlaw

ro6ley:
This PR adds support for disabling elements when Peachjam is running on Pocketlaw.
* [x] New class for elements to be hidden `pocketlaw-hidden`
* [x] CSS rule to hide the elements
* [x] JS functionality to attach user agent details to pocketlaw elements

### Screenshots
* **Chrome user agent:**
  Screenshot 2022-11-23 at 18 24 55

* **Pocketlaw user agent:**
  Screenshot 2022-11-23 at 19 48 30

ref #668

### DIFF
--- a/peachjam/static/stylesheets/_global.scss
+++ b/peachjam/static/stylesheets/_global.scss
@@ -142,6 +142,6 @@ footer {
   }
 }
 
-[data-user-agent*='pocketlaw'] .pocketlaw-hidden {
+html[data-user-agent*='pocketlaw'] .pocketlaw-hidden {
   display: none;
 }
--- a/peachjam/static/stylesheets/_global.scss
+++ b/peachjam/static/stylesheets/_global.scss
@@ -141,3 +141,7 @@ footer {
     }
   }
 }
+
+[data-user-agent*='Pocketlaw'] {
+  display: none;
+}
--- a/peachjam/static/stylesheets/_global.scss
+++ b/peachjam/static/stylesheets/_global.scss
@@ -142,6 +142,6 @@ footer {
   }
 }
 
-[data-user-agent*='Pocketlaw'] {
+[data-user-agent*='pocketlaw'] {
   display: none;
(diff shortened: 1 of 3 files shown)


peachjam/templates/peachjam/layouts/main.html
```diff
@@ -12,6 +12,16 @@
   <!--  app-prod.js depends on pdf.js to be loaded-->
   <script defer src="{% static 'js/app-prod.js' %}"></script>
 
+ <!-- Small script to add the current user agent to all elements with the class name 'pocketlaw-hidden'. -->
+  <script type="text/javascript">
+    document.addEventListener("DOMContentLoaded", function () {
+      const pocketlawElements = document.querySelectorAll('.pocketlaw-hidden');
+      pocketlawElements.forEach(el => {
```
longhotsummer, line 19: Rather set the user agent on the html element. Then we'll use css selectors to hide/show content. 

That way the JavaScript doesn't need to know about any special selectors - literally all it does is set the user agent data attribute on html. Everything else happens in css

peachjam/static/stylesheets/_global.scss
```diff
@@ -141,3 +141,7 @@ footer {
     }
   }
 }
+
+[data-user-agent*='pocketlaw'] {
```
longhotsummer, line 145: This would be "html[data...] .pocket law-hidden { display: none;}"

(Abbreviated because on my phone)

longhotsummer (changes requested):

peachjam/static/stylesheets/_global.scss
ro6ley: Works without the html part

ro6ley: Just seen the other comment, fixed both. Now it works

longhotsummer (approved):

Issue #706, opened 2022-11-24
Legislation facet view not using translations for facet titles


longhotsummer:
![image](https://user-images.githubusercontent.com/4178542/203783985-39d5d995-6eb7-4847-b157-b4ff0dfeb9e5.png)

longhotsummer:
@musangowope this should be a quick fix, I think, and is necessary for SenLII.

PR #707, opened 2022-11-24
Senlii fine tuning

longhotsummer:
* site navigation
* data on homepage
* legislation view (now moved into liiweb)
* improved translation support
* adds a placeholder list of legislation links (which is replaced in reality when the javascript runs) which ensures that website scrapers will index all our legislation, and support scraping for peachjam

https://www.loom.com/share/f97c477a7e934b55820732c067678fcb

PR #709, opened 2022-11-24
Add translations for  legislation table facets

musangowope:
resolves https://github.com/laws-africa/peachjam/issues/706


longhotsummer (dismissed):
Looks good, but I think the template will clash with similar translation changes I made in https://github.com/laws-africa/peachjam/pull/707

longhotsummer (approved):

PR #713, opened 2022-11-28
Fix legislation list view on AfricanLII

nickmwangemi:
Fixes https://github.com/laws-africa/peachjam/issues/712


actlikewill:
@nickmwangemi I think all we need to do is add "liiweb" to the settings file for africanlii.

actlikewill (approved):

PR #715, opened 2022-11-29
Amendments update

nickmwangemi:
This PR aims to update the code to use `work_amendments` rather than `amendments` from the metadata json.

## Before - (Take note of the latest event)

image

## After - (Newer events have been added)
image

### DIFF
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -241,11 +241,8 @@ class IndigoAdapter(Adapter):
                 object_work=repealing_work,
             )
 
-        amendments = (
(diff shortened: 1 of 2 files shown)


longhotsummer (commented):
Looks good. We must force an ingestor update of all content before deploying this, because this is a new metadata field and this change assumes the field is present.

longhotsummer:
I've queued up updates for all our sites, so on Wednesday morning this can be deployed.

It will cause errors for deleted legislation which cannot be updated from Indigo, which we can delete manually.

longhotsummer (approved):

PR #716, opened 2022-11-29
Gazettes ingestor

actlikewill:
- Adds and ingestor for gazettes

### DIFF
--- a/peachjam/adapters/gazettes.py
+++ b/peachjam/adapters/gazettes.py
@@ -1,0 +1,18 @@
+import logging
+
+from peachjam.plugins import plugins
+
+from .adapters import Adapter
+
+log = logging.getLogger(__name__)
+
+
+@plugins.register("gazette-adapter")
+class GazetteAdapter(Adapter):
+    def check_for_updates(self, last_refreshed):
+        log.info("Checking for new gazettes...")
+        return
+
+    def update_document(self, document_id):
+        log.info("Updating new gazettes...")
+        return
--- a/peachjam/adapters/gazettes.py
+++ b/peachjam/adapters/gazettes.py
@@ -1,6 +1,7 @@
 import logging
 
 from countries_plus.models import Country
+from django.db import connection
 from languages_plus.models import Language
 
 from peachjam.models import CoreDocument, Gazette, Locality, SourceFile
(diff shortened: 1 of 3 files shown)


actlikewill:
@longhotsummer I am running into this issue when running this code. 
```
 raise ValueError('Cannot assign "%r": the current database router prevents this relation.' % value)
ValueError: Cannot assign "<Work: /akn/ke/officialGazette/government-gazette/1906-03-15/153 - Kenya Government Gazette dated 1906-03-15 number 153>": the current database router prevents this relation.
```

I am not sure why because I am not trying to create cross database relationships. I tried adding the minimal db router to but it is still not working.

actlikewill:
@longhotsummer I'm having trouble figuring out how to copy the source file data  using the DynamicStorageFileField. :thinking:

peachjam/adapters/gazettes.py
```diff
@@ -0,0 +1,89 @@
...
+        super().__init__(settings)
+        self.jurisdiction = self.settings["jurisdiction"]
+
+    def check_for_updates(self, last_refreshed):
+        log.info(
+            f"Checking for new gazettes from Gazettes.Africa since {last_refreshed}"
+        )
+        if last_refreshed:
+            new_gazettes = (
+                CoreDocument.objects.filter(
```
longhotsummer, line 26: `doc_type="gazette"`

peachjam/adapters/gazettes.py
```diff
@@ -0,0 +1,89 @@
...
+            }
+
+            if document.locality:
+                data["locality"] = Locality.objects.get(code=document.locality.code)
+
+            doc, new = Gazette.objects.update_or_create(
+                expression_frbr_uri=expression_frbr_uri, defaults={**data}
+            )
+
+            source_file = SourceFile(
```
longhotsummer, line 77: check for an existing source file

peachjam/adapters/gazettes.py
```diff
@@ -0,0 +1,89 @@
...
+                document=doc,
+                filename=document.source_file.filename,
+                file=document.source_file.file,
```
longhotsummer, line 80: Here's a way to get the raw file field:

```python
>>> SourceFile.objects.filter(pk=1132).values_list('file', flat=True)[0]
's3:lawsafrica-gazettes-archive:archive/sc/2022/sc-government-gazette-dated-2022-10-21-no-67.pdf'
```

And you can write it directly with `update_or_create`

peachjam/adapters/gazettes.py
actlikewill: Done

peachjam/adapters/gazettes.py
actlikewill: @longhotsummer I have made this change but when the source file is saved it the `s3:lawsafrica-gazettes-archive` prefix  gets chopped in the database off even though I am copying the value directly as it is. Also I tried editing the value directly in the database to add the prefix but I still get the file not found error using the lawlibrary AWS credentials. Changing the AWS credentials to the gazette ones works.

peachjam/adapters/gazettes.py
```diff
@@ -0,0 +1,85 @@
...
+            )
+
+            ga_source_file = (
+                SourceFile.objects.filter(document=ga_gazette)
+                .values("file", "filename", "size", "mimetype")
+                .using("gazettes_africa")
+            )
+
+            if ga_source_file:
+                SourceFile.objects.update_or_create(
```
longhotsummer, line 78: If this doesn't work then the final option is to use a raw sql query to update that field.

actlikewill: Added

peachjam/adapters/gazettes.py
actlikewill: I'm curious whether the code you showed me on gazettemachine for updating source file actually works. Does it copy the source file path with the prefix?

longhotsummer:
We need a way of testing this. ZA has 50,000+ gazettes which is crazy if it goes wrong. Instead, let's merge this and then set it up for new.ulii.org and test it with that, since uganda only has a few thousand gazettes.

actlikewill:
> 

Got it.

Issue #597, opened 2022-10-25
Some document types make more sense to order alphabetically, rather than date


longhotsummer:
For example, https://agp.africanlii.org/legal_instruments/ are more useful ordered alphabetically by title

Order `GenericDocument` by `title`.

longhotsummer:
For these document types, it may also help users to put the alpha facet above the year facet.

nickmwangemi:
@longhotsummer Order both generic documents and legal instruments?

PR #719, opened 2022-11-30
Order generic documents and legal instruments by title

nickmwangemi:
### Before - (Generic Documents, ordered by date)
before

### Before - (Legal Instruments, ordered by date)
before_legal

### After - (Generic Documents, ordered by title)

after1

### After - (Legal Instruments, ordered by title)

after2


nickmwangemi:
Fixes #597

peachjam/js/components/ListFacets.vue
```diff
@@ -99,8 +99,8 @@ export default {
         { key: 'courts', value: this.$t('Court') },
         { key: 'docTypes', value: this.$t('Document type') },
         { key: 'judges', value: this.$t('Judges') },
-        { key: 'years', value: this.$t('Year') },
-        { key: 'alphabet', value: this.$t('Alphabetical') }
+        { key: 'alphabet', value: this.$t('Alphabetical') },
```
longhotsummer, line 102: Since I don't think we can easily do this only for specific pages, I think we should leave it as-is for now.

longhotsummer (changes requested):
I don't think we should make the JS change, but the rest looks good.

longhotsummer (approved):

Issue #708, opened 2022-11-24
As a user I would like to view changed provisions of document on mobile


musangowope:
Proposed mobile design:
Similar to annotations, an icon is displayed in the gutter on the same level as its associated content. I chose a clock icon to denote change over time.
![Icon view](https://user-images.githubusercontent.com/15995210/204309861-b09f1f87-07ea-435b-a099-345ccafdc6a2.png)

on click of the clock icon, a modal is displayed, showing the what changed provision content:
![Modal](https://user-images.githubusercontent.com/15995210/204310581-54cd15b2-aad7-49a2-8947-a57475ad96dd.png)

For a side-by-side viewing, perhaps we could have a horizontal overflow content scroll?

musangowope:
@longhotsummer let me know what you think about this

longhotsummer:
🤔 

Do you mean a proper "popup" modal, or is the diff embedded exactly as it is on desktop? I think the latter is much simpler.

I like the history icons -- it should fit in with how you did the annotation markers on mobile, right?

Either horizontal scroll, or disable the side-by-side. But horizontal scroll on the "modal" content is reasonable. You'll need to set a minimum width on the columns.

PR #725, opened 2022-12-01
Doc Diff Mobile

musangowope:
- Added code to suppose viewing doc diffs in mobile

Demo:
https://www.loom.com/share/24865b914fd6419e9de137aad9d376de

resolves https://github.com/laws-africa/peachjam/issues/708
resolves https://github.com/laws-africa/peachjam/issues/727


peachjam/js/components/DocDiffs/ProvisionDiffInline.vue
```diff
@@ -88,21 +89,52 @@ export default {
...
+      this.originalElement.style.height = '0';
+      this.originalElement.style.top = '0';
+      this.originalElement.parentNode.insertBefore(this.wrapperElement, this.originalElement);
+      this.wrapperElement.append(this.originalElement, this.$el);
     }
+    window.addEventListener('resize', this.setVw);
+  },
+
+  unmounted () {
+    window.addEventListener('resize', this.setVw);
```
longhotsummer, line 130: remove?

musangowope, line 130: removeEventListeners

peachjam/js/components/DocDiffs/ProvisionDiffInline.vue
```diff
@@ -88,21 +89,52 @@ export default {
...
   mounted () {
     this.loadDiffContentsets();
     this.originalElement = document.getElementById(this.provision.id);
+    this.wrapperElement = document.createElement('div');
+    this.wrapperElement.style.position = 'relative';
     if (this.originalElement) {
-      this.originalElement.style.display = 'none';
-      this.originalElement.insertAdjacentElement('beforebegin', this.$el);
+      // Hide originalElement behind ProvisionDiffInline via absolute positioning
+      this.originalElement.style.position = 'absolute';
```
musangowope, line 119: use reLayout from gutter

peachjam/js/components/DocDiffs/ProvisionDiffInline.vue
musangowope: done

musangowope: done

peachjam/js/components/DocDiffs/ProvisionDiffInline.vue
musangowope: @longhotsummer for this issue https://github.com/laws-africa/peachjam/issues/727
using `gutter.layoutItems()` does not work

The issue is that a gutter item cannot target its anchorElement if that anchor element has a style of `display: none` as shown in this demo
https://www.loom.com/share/09691fae383448a9a18685a3fa5fdf30

This is why I use an alternative way of hiding the originalElement.
Using `visiblity: hidden` preserves the element height while hiding the element

The reason I absolutely positioned the originalElement behind the DiffInlineProvision is so that gutter items associated with the originalElement can be on the same level as the DiffIlineProvision
image

Please let me know if you would to chat through this.

longhotsummer: Ah, thanks for clarify, that makes a lot of sense. Let' go with this.

longhotsummer (approved):

PR #751, opened 2022-12-06
Show that amendments are being applied to a work

nickmwangemi:
This PR adds logic to show that a work has outstanding amendments yet to be applied.

To test locally:
- Add an extra `work_amendment` item to the list of  `work_amendments` in the metadata_json for the document. Make sure that the date for new `work_amendment` comes after the latest work amendment date. This date will be used in coming up with the relevant notice.

image


nickmwangemi:
> lgtm!
> 
> I wonder if we should open a separate issue to style unapplied amendments differently in the timeline, or add an 'Amendment not yet applied' warning at each. Currently the only difference is that there's no 'Read this version' button at those dates, if I'm not mistaken. @nickmwangemi could you add a screenshot of the History tab to the description so the others can see what I mean?

Here's the history tab:

image

longhotsummer (dismissed):
Looks good. Now is also a good time to add the suggested label, rather than doing a whole separate PR.

nickmwangemi:
Updated to show `amendment not yet applied`

image

peachjam/views/legislation.py
```diff
@@ -156,19 +186,25 @@ def get_timeline_events(self):
...
+            point_in_time_dates = [
+                point_in_time["date"] for point_in_time in points_in_time
+            ]
+            event = [
+                {
+                    "date": amendment.get("date"),
+                    "event": "amendment",
+                    "amending_title": amendment.get("amending_title"),
+                    "amending_uri": amendment.get("amending_uri"),
+                    "flag": True
```
goose-life, line 200: I think it looks good and seems to be doing the right thing, I would just query naming the flag "flag". maybe something more descriptive like `unapplied_amendment` will help us in future?

nickmwangemi: Good call

peachjam/views/legislation.py
```diff
@@ -197,7 +197,7 @@ def get_timeline_events(self):
                     "event": "amendment",
                     "amending_title": amendment.get("amending_title"),
                     "amending_uri": amendment.get("amending_uri"),
-                    "flag": True
+                    "unapplied_amendment": True
                     if amendment.get("date") not in point_in_time_dates
```
goose-life, line 201: oh, sorry, this can probably be simplified a bit: 
`"unapplied_amendment": bool(amendment.get("date") not in point_in_time_dates)`?

goose-life (dismissed):
minor comment

longhotsummer (approved):

longhotsummer (dismissed):

longhotsummer (approved):

Issue #643, opened 2022-11-03
Chore: HTML and AKN document content layout width


longhotsummer:
I would like to be able to limit the max width of the actual HTML and AKN content of a document to about 50-55em, and have auto margins on the left and right. This is in line with guidance on limiting the length of a line for readability purposes (see https://practicaltypography.com/line-length.html).

The only width limit we have right now is the 80% width to allow for the gutter. We need a way to manage both.

longhotsummer:
I think what we need here is to apply a max-width and margin-auto to either a div inside `content-and-enrichments__inner` or `la-akoma-ntoso` inside `content-and-enrichments__inner`. So the gutter and the main content still have flex, but then the "page" within the main content column sizes itself independently, within that main + gutter combination.

PR #753, opened 2022-12-07
Add max width to content-and-enrichments__inner

musangowope:
Wide screen photo:
image

resolves # https://github.com/laws-africa/peachjam/issues/643

### DIFF
--- a/peachjam/static/stylesheets/components/_document-content.scss
+++ b/peachjam/static/stylesheets/components/_document-content.scss
@@ -73,6 +73,9 @@
   }
 
   .content-and-enrichments__inner {
+    max-width: 80em;
+    margin-left: auto;
+    margin-right: auto;
     .content {
       flex: 80%;
       max-width: 80%;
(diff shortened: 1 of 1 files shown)


peachjam/static/stylesheets/components/_document-content.scss
```diff
@@ -73,6 +73,9 @@
   }
 
   .content-and-enrichments__inner {
+    max-width: 80em;
```
musangowope, line 76: I chose 80em because 55em looked to squashed

longhotsummer: Ah, this is simpler than I thought. It keeps the content with the gutter, otherwise on wide views the gutter will be far away. Nice.

Since the content is 80% of the full width, 70 * 0.8 = 56, so let's go with 70em.

Additionally, pls  can you add `font-size: 15px` to `.content.content__html` above.

musangowope: done

longhotsummer (approved):

Issue #413, opened 2022-09-15
Add useful social sharing metadata to document pages


longhotsummer:
Add basic [opengraph social metadata](https://ogp.me/) to document detail pages, as well as the standard description metadata.

Fields:

* title
* date

PR #755, opened 2022-12-07
Add meta data in for social sharing documents

musangowope:
resolves https://github.com/laws-africa/peachjam/issues/413

screenshot
image

### DIFF
--- a/peachjam/templates/peachjam/layouts/document_detail.html
+++ b/peachjam/templates/peachjam/layouts/document_detail.html
@@ -3,6 +3,16 @@
 
 {% block title %}{{ document.title }}{% endblock %}
 
+{% block head-meta %}
+  {{ block.super }}
+  <meta property="og:title" content="{{ document.title }}" />
+  <meta property="og:type" content="book" />
+  <meta property="book:release_date" content="{{ document.date|date:'Y-m-d' }}" />
+  {% if document.author %}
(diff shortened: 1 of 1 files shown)


longhotsummer (approved):
Cool. Let's see what the social networks think of "og:type=book" :)

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -3,6 +3,16 @@
 
 {% block title %}{{ document.title }}{% endblock %}
 
+{% block head-meta %}
+  {{ block.super }}
+  <meta property="og:title" content="{{ document.title }}" />
+  <meta property="og:type" content="book" />
+  <meta property="book:release_date" content="{{ document.date|date:'Y-m-d' }}" />
+  {% if document.author %}
```
musangowope, line 11: I see a document as a book. Setting the `og:type` as a `book` also allows us to set the `book:release_date`, which is what we want to achieve. Setting a date.

the property `book:release_date` accepts a date format [ISO-8601](https://www.iso.org/iso-8601-date-and-time-format.html) which accepts the format `Y-m-d`

I also set the author if author field exists(Although, it wasnt a requirement in the ticket)

longhotsummer: All good :)

PR #756, opened 2022-12-08
Add logic to show applicable legislation based on language selected

musangowope:
I used fake data to prototype this functionality:
DEMO:
https://www.loom.com/share/74e15bab30994e80a1c92a0f19106613


musangowope:
The logic of showing the most applicable document based on the preferred language of the user should be handled from the backend

PR #764, opened 2022-12-12
Showing a specific doc section based for books and journals

musangowope:
I started adding the functionality to replicate the functionality we have on `edit.indigo`. 
I.e on click of table of content item, show only the corresponding part of the document.

This document behavior will be applicable to a specific document type books and journals

DEMO:
https://www.loom.com/share/a9475156e8b442818e723c0b1cecccc2


peachjam/js/components/DocumentContent/index.ts
```diff
@@ -30,18 +30,25 @@ class DocumentContent {
   private searchApp: any;
   private navOffCanvas: OffCanvas | undefined;
   private enchrichmentsManager: EnrichmentsManager | null;
+  private storedDocument: Node | undefined;
+  private tocController: HTMLElement| null;
   constructor (root: HTMLElement) {
     this.root = root;
     this.navOffCanvas = undefined;
     this.enchrichmentsManager = null;
+    this.storedDocument = this.root.querySelector('[data-document-element]')?.cloneNode(true);
```
longhotsummer, line 39: You run this `querySelector` three times. Better to do it here in the constructor once and store the element for re-use.

peachjam/js/components/DocumentContent/index.ts
```diff
@@ -147,16 +153,38 @@ class DocumentContent {
...
+      if (id) {
+        const sectionOfFocus = this.storedDocument.querySelector(`#${id}`)?.cloneNode(true) as HTMLElement | undefined;
+        if (!sectionOfFocus) return;
+        // Delete content within document element and then append section of focus
+        documentElement.replaceChildren(sectionOfFocus);
+      } else {
+        const storedDocument = this.storedDocument.cloneNode(true);
+        // @ts-ignore
+        documentElement.replaceChildren(...this.storedDocument.children);
+        this.storedDocument = storedDocument;
```
longhotsummer, line 172: I don't think this is necessary?

peachjam/js/components/DocumentContent/index.ts
```diff
@@ -147,16 +153,38 @@ class DocumentContent {
...
+      }
+      this.enchrichmentsManager?.layoutItems();
```
longhotsummer, line 174: Is this required? The gutter watches the main container and reflows when its children change, which should be sufficient.

peachjam/js/components/DocumentContent/index.ts
musangowope: done

peachjam/js/components/DocumentContent/index.ts
musangowope: refactored

peachjam/js/components/DocumentContent/index.ts
musangowope: good catch, nope this is not required. refactored

peachjam/js/components/DocumentContent/index.ts
```diff
@@ -30,18 +30,27 @@ class DocumentContent {
...
+  private documentElement: Element | null;
   constructor (root: HTMLElement) {
     this.root = root;
     this.navOffCanvas = undefined;
     this.enchrichmentsManager = null;
+    this.documentElement = this.root.querySelector('[data-document-element]');
+    this.originalDocCloned = this.documentElement?.cloneNode(true);
+    this.tocController = this.setupTocForTab();
+
+    if (root.dataset.frbrUriSubtype === 'book' && root.getAttribute('data-display-type') === 'akn') {
```
longhotsummer, line 44: Please delegate this to an attribute on `root` -- that makes it easier for us to determine in python to en/disable this behaviour, and means this component doesn't need to know about subtypes etc.

peachjam/js/components/DocumentContent/index.ts
```diff
@@ -147,16 +154,34 @@ class DocumentContent {
...
+    this.tocController?.addEventListener('itemTitleClicked', (e) => {
+      const customEvt = e as CustomEvent;
+      const id = customEvt.detail.target.getAttribute('href').replace('#', '');
+      if (!this.documentElement) return;
+      if (!(this.originalDocCloned && this.originalDocCloned instanceof HTMLElement)) return;
+      if (id) {
+        const sectionOfFocus = this.originalDocCloned.querySelector(`#${id}`)?.cloneNode(true) as HTMLElement | undefined;
+        if (!sectionOfFocus) return;
+        // Delete content within document element and then append section of focus
+        this.documentElement.replaceChildren(sectionOfFocus);
```
longhotsummer, line 167: TIL about replaceChildren :)

peachjam/js/components/DocumentContent/index.ts
longhotsummer: eg. `data-toc-show-active-item-only` or something. If that's present, then this behaviour is turned on.

musangowope: done

longhotsummer:
Note to self: for HTML documents where we generate our own TOC based on headings, we need to "indent and group" the heading with its content (everything until the next heading).

longhotsummer (approved):

PR #768, opened 2022-12-14
fixes legislation amendments

actlikewill:
- fixes legislation amendments
-


musangowope (changes requested):
Legislation page not working
image

longhotsummer (approved):
I *think* this is fine.

musangowope (approved):

PR #777, opened 2022-12-15
Order document listings by title when alphabet filter is applied.

nickmwangemi:
Before (With alphabet filter applied, the docs are listed by date):
image

After (With alphabet filter applied, the docs are listed by title, in alphabetical order):
image

image


peachjam/views/generic_views.py
```diff
@@ -45,6 +45,8 @@ def get_base_queryset(self):
         return qs.preferred_language(self.get_language())
 
     def get_queryset(self):
+        if self.form.data.get("alphabet"):
+            return self.form.filter_queryset(self.get_base_queryset()).order_by("title")
```
longhotsummer, line 49: You're repeating code here. It's better to do what is common in both cases once -- get the queryset and filter it -- and then apply the appropriate ordering.

Something like:

```python
qs = self.form.filter_queryset(self.get_base_queryset()).order_by("-date")
if ...:
  qs = qs.order_by("title")
return qs
```

It would also make sense to me for the form itself to apply the ordering in `filter_queryset`. The form has the knowledge of the form fields, etc.

longhotsummer: See https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

longhotsummer (changes requested):
Small change, but it would make for better code.

longhotsummer (approved):
great, much simpler

Issue #494, opened 2022-10-03
As a reader I want citations linked in PDFs


longhotsummer:
We have the bulk of the plumbing here, except that pdftotext extracts slightly different looking text to pdfjs, so the extracted links don't match. Simplest is to change the citation link creation to include the context of the citation, so that the library we use to attach the links to the DOM when displaying the PDF can use its fuzzy matcher to find the correct citation text.

longhotsummer:
The major blocker here is now the fact that the text from `pdftotext` is slightly different (in terms of whitespace usually) to that generated in the dom by `pdf.js`. The solution I'd like to implement is to replace the call to `pdf_to_text` [here](https://github.com/laws-africa/peachjam/blob/a42013a9a7b4b2b736b57abdf64d93486f85f2ab/peachjam/analysis/citations.py#L42) (which calls the normal `pdftotext` with a call to a new function which calls out to nodejs to run a `pdf.js`-based text extraction.

So we need a node script that does something like (or uses) https://github.com/ffalt/pdf.js-extract to get the text from the document as though pdf.js rendered it.

The return value of the new python function should by a string of all the page text, with each page separated by ctrl-L (which pdftotext uses to signify a new page).

longhotsummer:
It looks like we'll need the DOM-sourced version of the text, and pdf.js-extract doesn't provide that. I'm looking into how to build the DOM textLayer in node and extract the text from that

longhotsummer:
Completed and deployed.

PR #782, opened 2023-01-04
Extract and display PDF citations

longhotsummer:
* show citation links on PDF documents
* use pdf.js-based text extraction so that the extracted text is the same as the rendered text

https://www.loom.com/share/c139fec826ad4cbe9bbaac8c8074e09a


longhotsummer:
Fixes #494

PR #793, opened 2023-01-09
Show static files links on Pocketlaw

nickmwangemi:
This PR aims to resolve static files links to show correctly on Pocketlaw.

Previously, PDFs on Pocketlaw, while offline, would not load due to how Django handles static files by adding hash values to the names, e.g. `https://lawlibrary.org.za/static/lib/pdfjs/pdf.worker.js` becomes `https://lawlibrary.org.za/static/lib/pdfjs/pdf.4a2a1c0fb5fb.js` and the worker.js is not available in the archive produced by Wget.

### DIFF
--- a/peachjam/views/pocketlaw.py
+++ b/peachjam/views/pocketlaw.py
@@ -1,0 +1,5 @@
+from django.views.generic import TemplateView
+
+
+class PocketLawResources(TemplateView):
+    template_name = "peachjam/pocket_law_resources.html"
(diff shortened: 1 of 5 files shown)


peachjam/views/pocketlaw.py
```diff
@@ -0,0 +1,6 @@
+from django.views.generic import TemplateView
+
+
+class PocketLawResources(TemplateView):
+    template_name = "peachjam/pocket_law_resources.html"
+    navbar_link = "pocketlaw"
```
longhotsummer, line 6: The navbar link is not needed

longhotsummer (approved):

Issue #723, opened 2022-12-01
When creating a DocumenNature when ingesting from Indigo, capitalize the nature


longhotsummer:
https://github.com/laws-africa/peachjam/blob/aaf769a589158b6f1d3ac36bfbd05686fc78a910/peachjam/adapters/adapters.py#L163

eg. rather than `activity-report`, make it `Activity report`. So replace `-` with ` ` and title case it.

Issue #779, opened 2022-12-19
Dropdown list of countries in top bar does not work


NotoriousMBB:
from a search link in Chrome

https://gazettes.africa/search/?q=%22Decreto-Lei+n.%C2%BA+1%2F2009%22&jurisdiction=Mozambique

PR #794, opened 2023-01-09
Capitalize document nature name in adapter

actlikewill:
- Capitalize document nature name in adapter

closes laws-africa/peachjam#723


longhotsummer:
Fixes #779

longhotsummer:
See the comment in https://github.com/laws-africa/peachjam/issues/723: It should be "Activity report", not "Activity Report". So you need to use `.capitalize()` not `.title()`

longhotsummer (approved):

Issue #479, opened 2022-09-28
As a content manager of a peachjam website, I would like to have mechanism to prevent some documents from be scraped by web crawlers


NotoriousMBB:
to disallow certain URLs that the content team has applied the exclusion from search policy to protect the privacy and identity of individuals.

PR #796, opened 2023-01-10
Prevent scraping

nickmwangemi:
This PR adds functionality for preventing some documents from being scraped by web crawlers and is achieved by:

* [x] Adding a new `allow_robots` field to the CoreDocument model.
* [x] Excluding URLs that have the flag set to false.
* [x] Adding a `noindex` meta tag to the document detail page.

Closes https://github.com/laws-africa/peachjam/issues/479

Demo:
https://www.loom.com/share/c8517b2aae14404f863252bec5043b5b


peachjam/admin.py
```diff
@@ -251,6 +258,7 @@ class DocumentAdmin(admin.ModelAdmin):
                     "title",
                     "date",
                     "language",
+                    "allow_robots",
```
longhotsummer, line 261: Hmm, let's put it under the Advanced section -- otherwise editors will wonder about it and click it randomly :)

It's not a common option, no need to fill up the key details section with something the editors won't worry about 99% of the time.

peachjam/models/core_document_model.py
```diff
@@ -218,6 +218,7 @@ class CoreDocument(PolymorphicModel):
 
     created_at = models.DateTimeField(_("created at"), auto_now_add=True)
     updated_at = models.DateTimeField(_("updated at"), auto_now=True)
+    allow_robots = models.BooleanField(_("allow robots"), default=True)
```
longhotsummer, line 221: Two suggestions:

1. add `db_index=True` so that it's indexed by the db. We don't want to have to go through 100,000s of rows in the db just to find the 10s that have this set to False.
2. add `help_text=_("Allow this document to be indexed by search engine robots.")`

actlikewill (dismissed):
Looks good.

peachjam/views/robots.py
```diff
@@ -0,0 +1,19 @@
+from django.views.generic import TemplateView
+
+from peachjam.models import CoreDocument
+
+
+class RobotsView(TemplateView):
+    template_name = "peachjam/robots.txt"
+    content_type = "text/plain"
+
+    def get(self, request, *args, **kwargs):
```
longhotsummer, line 10: You're only adding context, so overriding get_context_data() is the preferred way of doing this -- no need to override `get()`.

longhotsummer (changes requested):
Minor change, pls.

longhotsummer (approved):

PR #797, opened 2023-01-11
Source url field in source file

### LINKED ISSUE #790: Redirect gazettes PDFs on lawlibrary.org.za to gazettes.africa
We don't want webcrawlers to duplicate the crawls for gazettes on lawlibrary and on gazettes.africa. Instead, when asking for the source PDF for a gazette that we know lives on gazettes.africa, it should redirect to the gazettes.africa archive URL.

Example:

```
https://lawlibrary.org.za/akn/za/officialGazette/government-gazette/2022-01-03/45717/eng@2022-01-03/source ->
https://gazettes.africa/akn/za/officialGazette/government-gazette/2022-01-03/45717/eng@2022-01-03/source ->
https://archive.gazettes.africa/archive/za/2022/za-government-gazette-dated-2022-01-03-no-45717.pdf
```

actlikewill:
- adds source url to gazette ingestor
- adds redirect to source url in document source pdf view

closes #790


peachjam/views/documents.py
```diff
@@ -86,7 +86,14 @@ def render_to_response(self, context, **response_kwargs):
 class DocumentSourcePDFView(DocumentSourceView):
     def render_to_response(self, context, **response_kwargs):
         if hasattr(self.object, "source_file"):
-            file = self.object.source_file.as_pdf()
+            source_file = self.object.source_file
+            file = source_file.as_pdf()
+
+            # redirect search engine crawlers to the original source files
+            # especially for gazettes
+            if source_file.source_url:
```
longhotsummer, line 94: I think we should only do this for DocumentSourceView, since that is the view that returns the original source file, no matter its format. This view is explicitly about converting it to PDF, and we don't know if the source URL will provide a PDF.

peachjam/models/core_document_model.py
```diff
@@ -400,6 +400,9 @@ class AttachmentAbstractModel(models.Model):
     mimetype = models.CharField(_("mimetype"), max_length=1024, null=False, blank=False)
     size = models.BigIntegerField(_("size"), default=0)
     file = DynamicStorageFileField(_("file"), upload_to=file_location, max_length=1024)
+    source_url = models.URLField(
```
longhotsummer, line 403: My idea was to put this on SourceFile, not AttachmentAbstractModel. What's the benefit of having it here? What are the semantics of having a source_url for an image attachment, say?

Unless we have a good reason, I think we should limit this just to SourceFile, because we have a good reason for it there. We can always move it later if there is a reason to have it on all attachments.

peachjam/adapters/gazettes.py
```diff
@@ -78,8 +78,12 @@ def update_document(self, expression_frbr_uri):
             file_path = ga_source_file[0].pop("file")
 
             if ga_source_file:
+                source_url = (
+                    f"https://archive.gazettes.africa/{file_path.split(':')[2]}"
```
longhotsummer, line 82: I think it's better to use `https://gazettes.africa/akn/za/officialGazette/government-gazette/2022-01-03/45717/eng@2022-01-03/source` because:

1. the archive path is internal to how gazettes.africa works
2. gazettes.africa will redirect the `.../source` URL to the correct place
3. we don't have to guess the archive path URL, which is internal to gazettes.africa
4. we know exactly how to build the `.../source` URL with the FRBR URI of the gazette.

longhotsummer (changes requested):
A few changes, pls. Lmk if you want to talk about them.

peachjam/views/documents.py
actlikewill: @longhotsummer  What happens when the crawlers hit `source.pdf`? Will they not force the file to be downloaded from s3 to be converted?

longhotsummer: Good question. We only advertise a source.pdf link for non-PDF documents. So I think we're okay on this for now.

peachjam/adapters/gazettes.py
actlikewill: I've changed it to use gazettes.africa/source.

peachjam/views/documents.py
actlikewill: @longhotsummer We are also using `source.pdf` to load source files in the detail view if they are pdfs i.e that is what pdf.js is using to render the pdfs

longhotsummer (dismissed):
great

Issue #801, opened 2023-01-12
As a bulk importer I want to be able to import RTF files


longhotsummer:
In some cases we have RTF and PDF files. We want to import the RTF files in the same way we do docx files, preferring them to PDFs.

eg. see https://github.com/laws-africa/peachjam/blob/8dbc08d8cff5e1616870cbd1809c43ab7adaf8dd/peachjam/resources.py#L103

longhotsummer:
Additionally there is a bug that `source_url` may be None if there's only an `.rtf` or other non-pdf and non-docx.

We should also make the change that if there are multiple source_urls, then we should find the preferred one that is the main source document (as we currently do), and then add the others as arbitrary attachments (like we allow for summary documents).

longhotsummer:
So the additional files should be attached with a hard-coded attachment nature of "Original document"

PR #802, opened 2023-01-13
RTF files in bulk import

actlikewill:
- [x]  check for rtf files in importer
- [x] import all other files as media attachments

### DIFF
--- a/peachjam/resources.py
+++ b/peachjam/resources.py
@@ -70,6 +70,7 @@ class ForeignKeyRequiredWidget(ForeignKeyWidget):
 class SourceFileWidget(CharRequiredWidget):
     def clean(self, value, row=None, **kwargs):
         super().clean(value)
+
         source_url = self.get_source_url(value)
         try:
             with TemporaryDirectory() as dir:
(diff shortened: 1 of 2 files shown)


peachjam/resources.py
```diff
@@ -96,21 +97,35 @@ def clean(self, value, row=None, **kwargs):
...
+
+        if docx_urls:
+            return docx_urls[0]
+        if rtf_urls:
+            return rtf_urls[0]
+        if pdf_urls:
+            return pdf_urls[0]
 
-        return source_url
+        return source_files[0]
```
longhotsummer, line 128: Will source_files always have at least one element?

peachjam/resources.py
```diff
@@ -86,7 +87,7 @@ def clean(self, value, row=None, **kwargs):
                     elif mime in DOC_MIMETYPES:
```
longhotsummer, line 87: Does DOC_MIMETYPES need to be updated to include RTF?

peachjam/resources.py
actlikewill: Yes it will

peachjam/resources.py
actlikewill: I've added it. I thought that we would have to create a separate pipline for .rtf but adding the mimetype now extracts the text correctly. :100:

longhotsummer (approved):
I hope it's tested and works :)

longhotsummer:
Fixed #801

PR #803, opened 2023-01-13
Add functionality for importing multiple alternative citations

nickmwangemi:


peachjam/resources.py
```diff
@@ -359,6 +361,13 @@ def after_import_row(self, row, instance, row_number=None, **kwargs):
                 case_number.document = judgment
                 case_number.save()
 
+            # extract alternative citations
```
longhotsummer, line 364: I think we should consistently call them alternative names, because they're not always citations. it's simply in our current use case that they're citations.

Issue #611, opened 2022-10-28
As a user of the French version of the AGP site, I would like the search to recognise language-specific characters


NotoriousMBB:
[@Mariya](https://africanliiteam.slack.com/team/UB3CBRWDR)
 French accents and searching: just noting that on https://agp.africanlii.org/ when one does a search for differends one does not pickup the items with différends. This must have implications for users who don't use accents on letters when searching, I am sure this can be tweaked to pick up both?

longhotsummer:
I think we can use a similar approach to #177. We'll have to look into how we index different languages separately and then search multiple indexes concurrently.

longhotsummer:
Refs:

* https://www.elastic.co/blog/multilingual-search-using-language-identification-in-elasticsearch
* https://codarium.substack.com/p/designing-an-optimal-multi-language

PR #805, opened 2023-01-14
Multi-language search indexes

longhotsummer:
* configure language-based search indexes that use language-specific analyzers to index text fields
* configure the search view to search all indexes
* documents not in one of the supported languages are indexed into the old language-agnostic index using the default "standard" analyzer.

Index mappings:

image

image


longhotsummer:
@nickmwangemi i'm merging this into your branch so it can be deployed together, since both require re-indexing and will break search until the indexes are re-created.

longhotsummer:
Fixes #611

nickmwangemi (approved):

PR #816, opened 2023-01-16
Add gazettes to ulii site navigation

nickmwangemi:
This PR adds gazettes to the ULII site navigation.

image


longhotsummer:
Peachjam has a built-in gazettes view which is what we want to use, rather than sending the user to another site: https://github.com/laws-africa/peachjam/blob/main/peachjam/urls.py#L94

longhotsummer (approved):

PR #818, opened 2023-01-16
better handling of soffice errors. Fixes #810

### LINKED ISSUE #810: As an importer I want to see errors in preview when soffice can't extract text
Otherwise the import fails but we can't see why.

longhotsummer:
image


actlikewill:
Awesome

PR #821, opened 2023-01-17
Simplify import logic, remove duplicate code

longhotsummer:
* Fixes LAWLIBRARY-25 in sentry
* removes duplication of document nature setup


actlikewill (approved):
Works well.

PR #823, opened 2023-01-17
Redirects old lii case urls to peachjam liis

### LINKED ISSUE #792: As a user on a newly migrated website, I'd like the old case URLs to still work for a while after migration
while Google is still crawling the new migrated content.

actlikewill:
- Redirect old lii case urls to peachjam urls

closes #792

### DIFF
--- a/liiweb/views/redirect.py
+++ b/liiweb/views/redirect.py
@@ -1,0 +1,16 @@
+from django.conf import settings
+from django.views.generic import RedirectView
+
+
+class RedirectCaseURLsView(RedirectView):
+    """
+    A view to redirect legacy Lii Case URLs to new /akn/ peachjam urls.
+    We can map the court code from the old url and use the serial number
+    to get a work_frbr_uri, then redirect to the appropriate document,
+    if available
+    """
+
+    def get_redirect_url(self, court, year, number, *args, **kwargs):
+        court_code = settings.COURT_CODE_MAPPINGS.get(court)
+        frbr_uri = f"/akn/ug/judgment/{court_code.lower()}/{year}/{number}"
+        return frbr_uri
--- a/liiweb/views/redirect.py
+++ b/liiweb/views/redirect.py
@@ -10,7 +10,7 @@ class RedirectCaseURLsView(RedirectView):
     if available
     """
 
-    def get_redirect_url(self, court, year, number, *args, **kwargs):
-        court_code = settings.COURT_CODE_MAPPINGS.get(court, '')
-        frbr_uri = f"/akn/ug/judgment/{court_code.lower()}/{year}/{number}"
+    def get_redirect_url(self, country, court, year, number, *args, **kwargs):
+        court_code = settings.COURT_CODE_MAPPINGS.get(court, "")
+        frbr_uri = f"/akn/{country}/judgment/{court_code.lower()}/{year}/{number}"
         return frbr_uri
--- a/liiweb/views/redirect.py
+++ b/liiweb/views/redirect.py
@@ -11,6 +11,6 @@ class RedirectCaseURLsView(RedirectView):
     """
 
     def get_redirect_url(self, court, year, number, *args, **kwargs):
-        court_code = settings.COURT_CODE_MAPPINGS.get(court)
+        court_code = settings.COURT_CODE_MAPPINGS.get(court, '')
         frbr_uri = f"/akn/ug/judgment/{court_code.lower()}/{year}/{number}"
         return frbr_uri
--- a/ulii/urls.py
+++ b/ulii/urls.py
@@ -1,8 +1,15 @@
-from django.urls import include, path
+from django.urls import include, path, re_path
 
-from senlii import views
+from liiweb.views import RedirectCaseURLsView
+from ulii import views
 
 urlpatterns = [
     path("", views.HomePageView.as_view(), name="home_page"),
+    # This redirects old Ulii case laws to work with peachjam urls
+    re_path(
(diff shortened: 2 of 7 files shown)


ulii/views/redirect.py
```diff
@@ -0,0 +1,18 @@
...
+COURT_CODE_MAPPINGS = {}
+
+
+class RedirectCaseURLsView(RedirectView):
+    pattern_name = "document_detail"
+
+    def get_redirect_url(self, court, year, number, *args, **kwargs):
+        court_code = COURT_CODE_MAPPINGS[court]
+        expression_frbr_uri = (
+            f"/akn/ug/judgment/{court_code}/{year}/{number}/eng@{year}"
```
actlikewill, line 15: @longhotsummer We need to have the full date for this expression_frbr_uri to work. We only have year here.

actlikewill: We are also assuming the language

actlikewill: I have changed the code to search using the year and and frbr_uri_number.

longhotsummer: Peachjam accepts a work FRBR URI and will find the latest date and best language to use. So `/akn/ug/judgment/abc/2022/3` will redirect as appropriate.

eg. visit https://lawlibrary.org.za/akn/za/act/1996/constitution and you'll see it redirect.

ulii/urls.py
```diff
@@ -1,8 +1,13 @@
-from django.urls import include, path
+from django.urls import include, path, re_path
 
-from senlii import views
+from ulii import views
 
 urlpatterns = [
     path("", views.HomePageView.as_view(), name="home_page"),
+    re_path(
```
longhotsummer, line 7: Pls add a comment that these are for legacy URLs, both here and in the view. That way someone reading the code can work out why there are special handlers for it.

ulii/views/redirect.py
```diff
@@ -0,0 +1,38 @@
...
+    "high-court-execution-bailiffs-division-uganda": "UGHCEBD",
+    "election-petitions-uganda": "UGHCEP",
+    "equal-opportunities-commission": "UGEOC",
+    "center-arbitration-dispute-resolution-uganda": "UGCADER",
+    "tax-appeals-tribunal-uganda": "UGTAT",
+}
+
+
+class RedirectCaseURLsView(RedirectView):
+    def get_redirect_url(self, court, year, number, *args, **kwargs):
```
longhotsummer, line 29: I suggest not even checking that the object exists. Simply redirect to the raw URL based on the various parameters. Keeps it simple.

longhotsummer: This functionality will be used for multiple other legacy LIIs, so maybe we should put it into `liiweb` and have a way of configuring the court mappings from, say, the settings file?

actlikewill: > This functionality will be used for multiple other legacy LIIs, so maybe we should put it into `liiweb` and have a way of configuring the court mappings from, say, the settings file?

I had the same thought.

ulii/views/redirect.py
actlikewill: Awesome. I didn't know we did this. :+1:

ulii/urls.py
actlikewill: Done

ulii/views/redirect.py
actlikewill: Changed :heavy_check_mark:

liiweb/views/redirect.py
```diff
@@ -0,0 +1,16 @@
...
+class RedirectCaseURLsView(RedirectView):
+    """
+    A view to redirect legacy Lii Case URLs to new /akn/ peachjam urls.
+    We can map the court code from the old url and use the serial number
+    to get a work_frbr_uri, then redirect to the appropriate document,
+    if available
+    """
+
+    def get_redirect_url(self, court, year, number, *args, **kwargs):
+        court_code = settings.COURT_CODE_MAPPINGS.get(court)
```
longhotsummer, line 14:
```suggestion
        court_code = settings.COURT_CODE_MAPPINGS.get(court, '')
```

otherwise court_code could be None and the `.lower()` will fail

liiweb/views/redirect.py
```diff
@@ -0,0 +1,16 @@
...
+        frbr_uri = f"/akn/ug/judgment/{court_code.lower()}/{year}/{number}"
```
longhotsummer, line 15: make the `ug` a two-letter country code in the URL

ulii/urls.py
```diff
@@ -1,8 +1,15 @@
...
 
-from senlii import views
+from liiweb.views import RedirectCaseURLsView
+from ulii import views
 
 urlpatterns = [
     path("", views.HomePageView.as_view(), name="home_page"),
+    # This redirects old Ulii case laws to work with peachjam urls
+    re_path(
+        r"^ug/judgment/(?P<court>[-\w]+)/(?P<year>\d+)/(?P<number>\d+)$",
```
longhotsummer, line 10: I suggest making the `ug` a regex component `[a-z]{2}` and then all of this can move into the liiweb base, and just be generic.

liiweb/views/redirect.py
actlikewill: Done

ulii/urls.py
actlikewill: Done

ulii/urls.py
```diff
@@ -1,6 +1,6 @@
 from django.urls import include, path
 
-from senlii import views
+from ulii import views
```
longhotsummer, line 3: hmm, that's weird -- good fix though

longhotsummer (approved):

PR #826, opened 2023-01-17
Set soffice timeout

### LINKED ISSUE #780: Put a timeout on calling out to soffice
Sometimes it seems to get stuck, we have a case that was stuck for 15+ hours.

nickmwangemi:
This PR sets the soffice timeout to 30 minutes

Closes #780


actlikewill (dismissed):

nickmwangemi:
@longhotsummer How best could we test this?

longhotsummer:
A bit tricky to test, I suggest you do it manually.

1. set the time to something human, like 5 seconds
2. temporarily hardcode the command to be `["sleep", "10"]` which will sleep for 10 seconds
3. ensure it times out correctly.


I actually think this should move into https://github.com/laws-africa/docpipe/blob/e9e5f7328728b53960b6de1b9590747cb50a5548/docpipe/soffice.py#L74 and this file then just calls out to `soffice()` in docpipe instead. That way it's centralised.

nickmwangemi:
@longhotsummer I've removed the file since it wasn't in use at any point in the project.

peachjam/docx.py
```diff
@@ -6,6 +6,9 @@
 
 logger = logging.getLogger(__name__)
 
+# subprocess timeout (in seconds)
+TIMEOUT = 60 * 30
+
 
 def convert_docx_to_pdf(docx_file):
```
longhotsummer, line 13: This can also be adjusted to call https://github.com/laws-africa/docpipe/blob/main/docpipe/soffice.py#L76 so that you don't need to duplicate the timeout.

longhotsummer (dismissed):

longhotsummer:
@nickmwangemi you also should update the dependency on docpipe in requirements.txt to use the latest sha1 from docpipe's main branch.

longhotsummer (approved):

PR #828, opened 2023-01-17
Setup Zanzibarlii

nickmwangemi:
image


longhotsummer (approved):
Great!

I deleted templates/zanzibarlii/home.html since it is unused (templates/liiweb/home.html is used), and zanzibarlii.css since it is compiled from zanzibarlii.scss.

PR #832, opened 2023-01-19
Regulation in legislation list

### LINKED ISSUE #791: As a user of a peachjam website, I would like to access a list of all regulations in one place
similar to https://ulii.org/legislation-si 

The list could possibly be accessible via a tab on the /legislation page next to All legislation. The tab could be called All regulations (or statutory instruments or subsidiary legislation as the terminology may be in the concerned jurisdiction)

actlikewill:
- [x] Add Subleg variant filter  to LegislationListView
- [x] Configure url
- [x] Get subleg label from PeachjamSettings model

closes #791


peachjam/models/settings.py
```diff
@@ -47,5 +47,15 @@ def __str__(self):
         return "Settings"
 
 
+class PeachjamLabels(models.Model):
+    name = models.CharField(max_length=1024)
+    value = models.CharField(max_length=1024)
+    setting = models.ForeignKey(PeachJamSettings, on_delete=models.CASCADE)
```
actlikewill, line 53: @longhotsummer I"ve added a PeachjamLabels Model with fk to PeachjamSettings. Was this what we were going for?

liiweb/templates/liiweb/legislation_list.html
```diff
@@ -29,6 +29,16 @@ <h1 class="my-4">{% trans 'Legislation' %}</h1>
               {% trans 'All legislation' %}
             </a>
           </li>
+          <li class="nav-item">
+            <a class="nav-link {% if view.variant == 'subleg' %}active{% endif %}" href="{% url 'legislation_list_regulations' %}">
+                {% if regulation_label %}
```
longhotsummer, line 34: 2 spaces, not 4.

Can we configure black or one of our other precommit tools to enforce this?

liiweb/urls.py
```diff
@@ -15,6 +15,11 @@
         views.LegislationListView.as_view(variant="all"),
         name="legislation_list_all",
     ),
+    path(
+        "legislation/regulations",
```
longhotsummer, line 19: Let's be consistent with the internal name for this. "subsidiary legislation" or "subleg" for short is the technical term. There are multiple variants of that, such as "regulations", "SIs", etc.

So I think "legislation/subsidiary" is best and either "subsidiary" elsewhere or "subleg".

liiweb/templates/liiweb/legislation_list.html
actlikewill: I don't think black works for html. Prettier could work but it doesn't work well with django templates. It breaks up the syntax causing errors. I will try and see if there is something else that can work. In any case this part of the code has changed.

liiweb/urls.py
actlikewill: Changed.

peachjam/templates/peachjam/legislation_detail.html
```diff
@@ -7,7 +7,7 @@
   {% if child_documents %}
     <li class="nav-item" role="presentation">
       <button class="nav-link" data-bs-toggle="tab" data-bs-target="#children-tab" type="button" role="tab" aria-controls="children-tab">
-        {% trans "Subsidiary legislation" %}
+          {{ PEACHJAM_SETTINGS.subleg_label }}
```
longhotsummer, line 10:
```suggestion
        {{ PEACHJAM_SETTINGS.subleg_label }}
```

longhotsummer (approved):

PR #836, opened 2023-01-19
Search using AND within fields, but OR between fields

longhotsummer:
Before:
image

Note that before:

1.  the top match is not related to the protocol at all
2. there is no "protocol" in the document type list

After:
image

Note that afterwards:

1. the top matches are strong matches on document title
4. there are a diversity of document types, including "Protocol"

### DIFF
--- a/peachjam_search/views.py
+++ b/peachjam_search/views.py
@@ -93,12 +93,38 @@ class NestedPageQueryBackend(BaseSearchQueryBackend):
         ]
 
 
+class CrossFieldSimpleQueryStringBackend(SimpleQueryStringQueryBackend):
+    """This implements a simple_query_string query across multiple fields, using AND logic for the terms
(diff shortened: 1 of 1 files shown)


longhotsummer:
This is a pretty major improvement. We've been confused for a few weeks why more complex searches with many terms just don't seem to match well on documents for which they're the exact title. Now we know why.

actlikewill (approved):
Interesting that search could not return a document even when you type in the exact title. With this change it does.

longhotsummer:
> Interesting that search could not return a document even when you type in the exact title. With this change it does.

I am continually surprised at how difficult/unclear elasticsearch makes it to do what I think must be an incredibly common use case: full-text search across documents with titles and bodies, that works in a way that isn't surprising to the user.

PR #837, opened 2023-01-20
disables settings in context for deploy

actlikewill:
- Deploy is failing because we are referencing the pj_settings in the context processor when the subleg_column is not present

### DIFF
--- a/peachjam/context_processors.py
+++ b/peachjam/context_processors.py
@@ -1,6 +1,6 @@
 from django.conf import settings
 
-from peachjam.models.settings import pj_settings
+# from peachjam.models.settings import pj_settings
 
 
 def general(request):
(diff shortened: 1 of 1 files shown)


longhotsummer:
Hmm, this will be a problem with any change to settings. 

Another option is to temporarily disable Dokku deployment checks

actlikewill:
@longhotsummer Is there a reason we do migrations in the post deploy? It seems that it should be a predeploy task.

longhotsummer:
Neither is perfect. If we migrate in predeploy and the checks fail, then we have an updated db but the old code still running. Greg--(iPhone)On 20 Jan 2023, at 7:01 AM, actlikewill ***@***.***> wrote:﻿
@longhotsummer Is there a reason we do migrations in the post deploy? It seems that it should be a predeploy task.

PR #844, opened 2023-01-20
Optimize query for legislation

nickmwangemi:
Before:

before

After:
after

Closes https://github.com/laws-africa/peachjam/issues/845


longhotsummer (approved):
Nice!

PR #867, opened 2023-01-24
Add support for document nature listing

nickmwangemi:


peachjam/views/document_nature.py
```diff
@@ -0,0 +1,24 @@
...
+from peachjam.models import DocumentNature, GenericDocument
+from peachjam.views.generic_views import FilteredDocumentListView
+
+
+class DocumentNatureListView(FilteredDocumentListView):
+    template_name = "peachjam/document_nature_list.html"
+    model = GenericDocument
+    queryset = GenericDocument.objects.prefetch_related("author", "nature", "work")
+
+    def get_context_data(self, **kwargs):
```
longhotsummer, line 12: There's a slightly cleaner pattern to not have to fetch the nature twice from the DB.

See how we do it for the locality in lawlibrary's legislation view: https://github.com/laws-africa/peachjam/blob/main/lawlibrary/views/legislation.py#L29

1. Override `get` to set the current nature on the view once-off, right at the start. This throws the 404 if there is no such nature, and everything stops
2. `get_context_data()` and `get_queryset()` can then rely on `self.nature` knowing that it is guaranteed to be set.

longhotsummer (commented):
Nice, just a minor code suggestion.

actlikewill (approved):

PR #868, opened 2023-01-24
Patch django-background-tasks

longhotsummer:
This effectively adds
https://github.com/arteria/django-background-tasks/pull/244. Without this, all tasks are loaded from the db just to choose the most recent 5, which is very slow on large task lists (eg. re-indexing 100,000+ gazettes)


actlikewill (approved):
that linked PR has been pending for quite a while lol.

PR #869, opened 2023-01-24
Setup GhaLII, MalawiLII & ZambiaLII on peachjam

nickmwangemi:
ghalii

malawilii

zambialii


longhotsummer (commented):
Looking good.

I was expecting to see some logos and possibly new banner images, based on the existing sites.

longhotsummer:
> Looking good.
> 
> I was expecting to see some logos and possibly new banner images, based on the existing sites.

Also, some custom SCSS for each site to customise the primary brand colour?

actlikewill:
@longhotsummer We've added the logos and the banner images. We've also moved some context data for the home page to the liiweb base class. If  the primary theme colors are available we can also add them. Otherwise I think we are about ready.

nickmwangemi:
> @longhotsummer We've added the logos and the banner images. We've also moved some context data for the home page to the liiweb base class. If the primary theme colors are available we can also add them. Otherwise I think we are about ready.

We can pick the primary theme colors from the existing sites.

longhotsummer (dismissed):

PR #877, opened 2023-01-26
Guard against null court. Fixes LAWLIBRARY-21 in sentry

longhotsummer:
a ForeignKey relationship can be None


actlikewill:
https://github.com/laws-africa/peachjam/blob/2b8f8761dc7600fab4b67fbd97e7543f75dc77bf/peachjam/models/judgment.py#L165

I think this also needs to change since we are calling that method in self.clean

longhotsummer:
> https://github.com/laws-africa/peachjam/blob/2b8f8761dc7600fab4b67fbd97e7543f75dc77bf/peachjam/models/judgment.py#L165
> 
> I think this also needs to change since we are calling that method in self.clean

Good catch, updated.

actlikewill (approved):

PR #881, opened 2023-01-27
Custom login view

actlikewill:
- Adds a custom login view
- With this we can add the social apps to the context and then render the links if they are available.


peachjam/views.py
```diff
@@ -1,16 +0,0 @@
-from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
-from django.views.generic import TemplateView
-
-
-class AuthedViewMixin(LoginRequiredMixin, PermissionRequiredMixin):
```
longhotsummer, line 5: Where have these moved to?

actlikewill: `AuthedViewMixin` class is not in use anywhere that I can see. The Homepage class is in home.py

longhotsummer (approved):

nickmwangemi (approved):

Issue #853, opened 2023-01-23
Law Library - pull in gazettes


longhotsummer:
There are 60,000+ gazettes in ZA.

The blockers/performance considerations here:

1. this will drastically increase the time it takes to re-index for search (there is a bug with django-background-tasks that makes long queues even slower)
2. gazette pages will have citations extracted

We need to decide what changes we make to handle this operational impact.

longhotsummer:
This is largely done.

Requires #883

longhotsummer:
Outstanding: change "Gazettes" nav bar link to go to "/gazettes"

PR #883, opened 2023-01-29
Gazette view query improvements

longhotsummer:
* the statement `qs = self.queryset or self.model` evaluates `self.queryset` which effectively loads everything from the database :( -- instead, we check if it's None
* gazette views don't need all the faceted counts and don't need to extend from the filtered document view
* add a db index on `date` which we often filter and order by
* link gazettes into lawlibrary nav

Before:

image

After:

image


longhotsummer:
Build issue seems to be this: https://stackoverflow.com/questions/75269700/pre-commit-fails-to-install-isort-5-11-4-with-error-runtimeerror-the-poetry-co

longhotsummer:
Closes #853

actlikewill (approved):
Works well

PR #884, opened 2023-01-30
re-declare pdfjs as a node dependency

longhotsummer:
without this, we can't run pdfjs-to-text on node


actlikewill:
Was going through this before you merged. Interesting that you have the lib/pdfjs as a plain text that you then use in the citations. I was thinking that maybe you can take us through the way we extract citations. Maybe at our next code show and tell.

PR #892, opened 2023-01-31
Update duplicates in bulk import

actlikewill:
- [ ] generate frbr uri for judgments before import
- [ ] generate frbr uri for generic documents before import
- [ ] error handling

### DIFF
--- a/peachjam/resources.py
+++ b/peachjam/resources.py
@@ -262,9 +262,6 @@ class BaseDocumentResource(resources.ModelResource):
         except requests.exceptions.RequestException:
             return
 
-    def before_import_row(self, row, **kwargs):
-        logger.info(f"Importing row: {row}")
-
     def skip_row(self, instance, original, row, import_validation_errors=None):
         return row["skip"]
 
@@ -340,7 +337,6 @@ class GenericDocumentResource(BaseDocumentResource):
         "nature",
         "source_url",
         "title",
-        "work_frbr_uri",
     )
 
     class Meta(BaseDocumentResource.Meta):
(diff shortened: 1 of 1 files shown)


peachjam/resources.py
```diff
@@ -265,6 +265,22 @@ def download_attachment(url, document, nature):
     def before_import_row(self, row, **kwargs):
         logger.info(f"Importing row: {row}")
 
+        frbr_uri_data = {
```
longhotsummer, line 268: It's annoying that we have to do this before the row has been validated and cleaned :/

Bear in mind we can only generate an expression FRBR URI if the serial number override is provided; otherwise, skip this whole thing

actlikewill: @longhotsummer Yes. We need the serial number override for this to work.  It does ruin some of the validations that we had setup. I am testing to see the behavior now For the generic documents, we had the work_frbr_uri in the spreadsheet. Should I assume that this is still the case or should we try to generate it in the importer? Also are Mariya and the team still using the importer for generic documents?

peachjam/resources.py
```diff
@@ -340,26 +337,26 @@ class GenericDocumentResource(BaseDocumentResource):
...
-        row["language"] = frbr_uri.default_language
-        row["jurisdiction"] = str(frbr_uri.country).upper()
-        row["locality"] = frbr_uri.locality
-        row["frbr_uri_number"] = frbr_uri.number
-        row["frbr_uri_doctype"] = frbr_uri.doctype
-        row["frbr_uri_subtype"] = frbr_uri.subtype
-        row["frbr_uri_date"] = frbr_uri.date
+        if row.get("work_frbr_uri"):
+            frbr_uri = FrbrUri.parse(row["work_frbr_uri"])
+            row["language"] = frbr_uri.default_language
```
longhotsummer, line 349: A work URI doesn't have a language or a date.

PR #895, opened 2023-02-01
adds check for static file

actlikewill:
We can have a common template for homepage for all the liis, we need to render unique images for the banner and the logo. However images can be banner.png or banner.jpg. 
- My idea was to have a filter that checks if the image exists and then return the path to the static file tag.

- Not sure if this is the best approach. Another idea is to  put the mage names in the settings file.


longhotsummer:
Can't we simply standardise on `banner.jpg`?

PR #903, opened 2023-02-03
Taxonomy slug to include parent slug

actlikewill:
- changes taxonomy slug to include parent slug

### DIFF
--- a/peachjam/models/taxonomies.py
+++ b/peachjam/models/taxonomies.py
@@ -19,7 +19,8 @@ class Taxonomy(MP_Node):
         return self.name
 
     def save(self, *args, **kwargs):
-        self.slug = slugify(self.name)
+        parent = self.get_parent()
+        self.slug = (f"{parent.slug}-" if parent else "") + slugify(self.name)
         super().save(*args, **kwargs)
 
 
--- a/peachjam/tests/test_bulk_import.py
+++ b/peachjam/tests/test_bulk_import.py
@@ -108,7 +108,7 @@ class JudgmentBulkImportTestCase(TestCase):
         node = get(root.pk).add_child(name="Land Rights")
         get(node.pk).add_sibling(name="Environment")
 
-        data.append("land-rights|environment")
+        data.append("collections-land-rights|collections-environment")
         headers.append("taxonomy")
         dataset = tablib.Dataset(data, headers=headers, depth=0)
         result = JudgmentResource().import_data(dataset=dataset, dry_run=False)


longhotsummer:
> @longhotsummer The problem now is using the slug in bulk import is now cumbersome. We need to the full slug including the parent. Perhaps we can switch to using names instead.

I don't think this is a problem. The importer has to navigate to the page of the hierarchy they want, and then get the slug from the URL. If they were using the name they would have the same problem -- the name is not globally unique.

peachjam/models/taxonomies.py
```diff
@@ -19,7 +19,8 @@ def __str__(self):
         return self.name
 
     def save(self, *args, **kwargs):
-        self.slug = slugify(self.name)
+        parent = self.get_parent()
+        self.slug = slugify(f'{parent.slug + "-" if parent else ""}{self.name}')
```
longhotsummer, line 23: The f-string and the "+" don't really make sense together. 

```
self.slug = (f"{parent.slug}-" if parent else "") + slugify(self.name)
```

longhotsummer:
You'll need a migration to update the existing taxonomy entries.

peachjam/migrations/0047_taxonomy_slug.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.2.16 on 2023-02-06 10:24
+
+from django.db import migrations
+
+from peachjam.models import Taxonomy
+
```
actlikewill, line 6: @longhotsummer Not sure if its okay to import the model directly like this in the migration. The get_tree method doesn't work when I use `apps.get_model("peachjam", "Taxonomy")` I want the update to traverse the tree starting with the topmost level

longhotsummer: We just need a mechanism to apply this globally without having to do it manually.

We can add this (which you're right, isn't actually good practice) and do the migration, and then remove this migration.

The alternative is we do it entirely outside of the codebase and simply run a script in a shell after deploying. That's also reasonable, and probably easier?

actlikewill: I'd rather run the script externally. We'd only have to do it for agp and lawlibrary right?. Do the other liis have any taxonomies yet?

longhotsummer: I think only those two. Sounds good.

longhotsummer (approved):

Issue #885, opened 2023-01-31
As a document reader I'd like a list of citations for a document


longhotsummer:
Both incoming and outgoing citations. This will help users understand the citation network.

longhotsummer:
# Proposal

## Frontend

1. we add a new `Citations` tab to the document detail view
2. the tab has two columns
3. the first column is titled `Cited documents` and shows a list of works cited by this one (outgoing citations)
4. the second column is titled `Documents citing this one` and shows a list of works that cite this one (incoming citations)

Q: Are the cited documents ordered by date or alphabetically by title?

## Backend

We need to store citations on a per-work basis, and keep them up to date. This is complicated by the fact that there can be multiple language and dated versions of a document.

Note that citations are stored in documents, but we want to store them on a **per-work** basis. For example, I don't care if different language or historical versions of a piece of legislation cite the one I'm reading. I just care that the current work was cited by another work.

### Storing citations

For HTML documents, the citations are embedded in the html as `<a href="/akn/...">` tags. For PDF documents, the citations are stored in the database as `CitationLink` objects.

We create a new `ExtractedCitation` object:

* `citing_work`: FK to the work doing the citing
* `target_work`: FK to the work being cited (could also be `cited_work`)

### Keeping extracted citations up to date

We currently extract citations on a document (either PDF or HTML) [here](https://github.com/laws-africa/peachjam/blob/main/peachjam/models/core_document_model.py#L341). It's tempting to update that logic to also store extracted citations. However, we need to do this on a work-basis, not a document basis.

Instead, we create a new signal handler for when a document is saved or deleted. The handler updates all extracted citations for the updated work. We probably want the signal to run in the background as a background task with a short delay, to avoid delaying saves on documents with many versions (eg. legislation).

To track extracted citations for a work

1. for each document related to a work...
2. build a set of extracted work FRBR URIs, taken from all `CitationLink` objects (PDFs) and all `<a href="/akn/...">` HTML links
3. replace ALL existing `ExtractedCitation` objects where the current work is the `citing_work` with the works identified in step 2.

PR #911, opened 2023-02-06
Show a list of citations for a document

nickmwangemi:
This PR adds functionality for displaying the list of extracted citations for a document.

Closes https://github.com/laws-africa/peachjam/issues/885

image

https://www.loom.com/share/7e4a9c3291cb477ab54145fa565729de


peachjam/models/citations.py
```diff
@@ -47,3 +47,20 @@ def from_extracted_citation(cls, citation):
                 },
             ],
         )
+
+
+class ExtractedCitation(models.Model):
+    citing_work = models.ForeignKey(
+        "peachjam.Work",
+        null=False,
+        on_delete=models.PROTECT,
```
longhotsummer, line 56: Cascade deletes, otherwise we can't easily delete the work.

For both foreign keys.

peachjam/models/core_document_model.py
```diff
@@ -409,6 +459,7 @@ class AttachmentAbstractModel(models.Model):
 
     def __str__(self):
         return f"{self.filename}"
+        return self.date.year
```
longhotsummer, line 462: ?

peachjam/models/core_document_model.py
```diff
@@ -79,6 +81,54 @@ def update_languages(self):
...
+            ExtractedCitation.objects.get_or_create(
+                citing_work=self, target_work=related_work
+            )
+
+    def fetch_related_works_frbr_uris(self):
+        """Returns a list of unique work_frbr_uris,
+        taken from CitationLink objects(for PDFs) and all <a href="/akn/..."> embedded HTML links."""
+        for document in self.documents.all():
+
+            work_frbr_uris = []
```
longhotsummer, line 103: A set object will do the duplicate detection for you, and is more efficient than searching a list.

```
work_frbr_uris = set()
```

peachjam/models/core_document_model.py
```diff
@@ -79,6 +81,54 @@ def update_languages(self):
...
+
+        # delete existing extracted citations
+        ExtractedCitation.objects.filter(citing_work=self).delete()
+
+        for related_work in related_works:
+            ExtractedCitation.objects.get_or_create(
+                citing_work=self, target_work=related_work
+            )
+
+    def fetch_related_works_frbr_uris(self):
```
longhotsummer, line 98: In peachjam, a `related` has a special meaning. So perhaps `fetch_cited_works_frbr_uris` is better?

peachjam/models/core_document_model.py
```diff
@@ -79,6 +81,54 @@ def update_languages(self):
...
+        """Returns a list of unique work_frbr_uris,
+        taken from CitationLink objects(for PDFs) and all <a href="/akn/..."> embedded HTML links."""
+        for document in self.documents.all():
+
+            work_frbr_uris = []
+
+            if document.content_html:
+                root = html.fromstring(document.content_html)
+                for a in root.xpath('//a[starts-with(@href, "/akn")]'):
```
longhotsummer, line 107: Both this link or the citation link below could be expression frbr uris. For example:

`/akn/za/act/2009/1/eng@2011-01-01`

We need to track citations between works, not expressions. The work URI for the above is:

`/akn/za/act/2009/1`

So in both cases, I think we need to parse the URI and then get the work component:

```
try:
  work_frbr_uris.add(FrbrUri.parse(a.attrib["href"]).work_uri())
except ValueError:
  # ignore malformed FRBR URIs
  pass
```

longhotsummer: The `FrbrUri` object is from `cobalt.uri`

longhotsummer:
Looking great.

please add count badges to each heading in the citations tab. 

Also, the long titles wrap and make it hard to see what is a wrapped line and what is a new citation. So maybe it's better to make the lists use '<ol>'?

Can you also sanity check that the citing work and target work and prefetched by the queryset to avoid n+1 errors. I think doing it on the methods on work works well.

longhotsummer:
Pls also sort the lists by the work title. Again, doing so in the queryset method is best. 

Maybe even set the default sorting in the model meta.

longhotsummer:
Oh meta won't work because the sorting is different depending on the query (cited va target)

longhotsummer:
> Looking great.
> 
> please add count badges to each heading in the citations tab. 
> 
> Also, the long titles wrap and make it hard to see what is a wrapped line and what is a new citation. So maybe it's better to make the lists use '<ol>'?

That's meant to be 'ol'

peachjam/migrations/0048_backfill_extracted_citations.py
```diff
@@ -0,0 +1,21 @@
...
+
+from django.db import migrations
+
+
+def backfill_extracted_citations(apps, schema_editor):
+    CoreDocument = apps.get_model("peachjam", "CoreDocument")
+    for doc in CoreDocument.objects.all().iterator(chunk_size=100):
+        Work = apps.get_model("peachjam", "Work")
+        work = Work.objects.get(frbr_uri=doc.work_frbr_uri)
+        work.update_extracted_citations()
```
longhotsummer, line 11: When accessing the model like this in a migration, the model only has the basic structure. It doesn't have any custom code associated with it. So this won't work, because `update_extracted_citations` won't exist on that class.

We only need to backfill once-off for all our sites. There are a few options

1. do this manually outside of the code by running the above snippet of code manually outside of the deployment.
2. add a management command (which we only ever need once, so it's overkill)
3. duplicate the business logic here directly -- this is the simplest from an operational perspective and what I recommend

peachjam/models/citations.py
```diff
@@ -47,3 +45,36 @@ def from_extracted_citation(cls, citation):
...
+        related_name="target_work",
+        verbose_name=_("target work"),
+    )
+
+    @classmethod
+    def for_citing_works(cls, work):
+        return (
+            cls.objects.prefetch_related("citing_work", "target_work")
+            .filter(citing_work=work)
+            .order_by("citing_work__title")
```
longhotsummer, line 71: If you're filtering on citing work, then the title will always be the citing work's title. So this should be ordering by the target work, not the citing work.

peachjam/models/citations.py
```diff
@@ -47,3 +45,36 @@ def from_extracted_citation(cls, citation):
...
+        )
+
+    @classmethod
+    def for_target_works(cls, work):
+        return (
+            cls.objects.prefetch_related("citing_work", "target_work")
+            .filter(target_work=work)
+            .order_by("target_work__title")
```
longhotsummer, line 79: Order by the citing work title.

longhotsummer (changes requested):
Almost there, some changes necessary.

peachjam/tasks.py
```diff
@@ -112,3 +112,23 @@ def extract_citations(document_id):
         raise
 
     log.info("Citations extracted")
+
+
+@background(queue="peachjam", schedule=60, remove_existing_tasks=True)
+def update_extracted_citations_for_a_work(document_id):
```
longhotsummer, line 118: This is now the work id, not the document id.

peachjam/migrations/0051_backfill_extracted_citations.py
```diff
@@ -0,0 +1,72 @@
...
+        extracted_citation_model.objects.filter(citing_work=work_obj).delete()
+
+        for target_work in target_works:
+            extracted_citation_model.objects.get_or_create(
+                citing_work=work_obj, target_work=target_work
+            )
+
+
+def backfill_extracted_citations(apps, schema_editor):
+    CoreDocument = apps.get_model("peachjam", "CoreDocument")
```
longhotsummer, line 51: Minor: you're going from document to work, and therefore re-doing the work for works that have multiple documents.

This can all be driven on `Work.objects.all()` rather than `CoreDocument`.

longhotsummer (changes requested):
Still some confusion between documents and works.

longhotsummer (approved):

longhotsummer:
@nickmwangemi just a heads up that it looks like this migration will take a while to process, so the deploy will be slow.

nickmwangemi:
> @nickmwangemi just a heads up that it looks like this migration will take a while to process, so the deploy will be slow.

Affirmative. It look a while for my local migration to run.

PR #914, opened 2023-02-07
Concurrency for deployment actions

actlikewill:
Sets concurrency for deployment actions to minimize server load during deployment.


actlikewill:
@longhotsummer I've also uncommented out the deployments for sieralii and lesotholii.

longhotsummer (approved):
This will run 4 concurrently, that should be okay. Let's see.

PR #917, opened 2023-02-07
Deploy to liiweb2

actlikewill:
- Change deployment url to liiweb2


.github/workflows/deploy.yml
```diff
@@ -4,11 +4,13 @@ on:
   push:
     branches: [ main ]
 
+concurrency:
+  group: liiweb2.africanlii.org
+
```
actlikewill, line 9: @longhotsummer by adding this, it will queue up the deployments so that they don't happen concurrently. Later merges will have to wait their turn to run. This may or may not be what we want.

nickmwangemi (approved):

Issue #846, opened 2023-01-21
Chore: integrate citator for richer citation extraction


longhotsummer:
* [x] configure peachjam to call out to citator
* [x] add new citator as a service
* [x] configure patterns and lookup for new citator

longhotsummer:
I'm currently adjusting the citator to make this possible.

longhotsummer:
Done!

PR #920, opened 2023-02-09
Call citator service

longhotsummer:
* updates the citation extraction step to call the new Laws.Africa citator service, if API details are provided
* background task to update citations
* background task to re-index for search

### DIFF
--- a/peachjam/analysis/citations.py
+++ b/peachjam/analysis/citations.py
@@ -3,6 +3,9 @@ import tempfile
 from bisect import bisect_left
 
 import lxml.html
+import requests
+from django.conf import settings
+from docpipe.matchers import ExtractedCitation
 
 from peachjam.helpers import pdfjs_to_text
 from peachjam.models import CitationLink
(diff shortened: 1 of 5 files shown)


longhotsummer:
Closes #846

peachjam/analysis/citations.py
```diff
@@ -27,7 +30,9 @@ def extract_citations(self, document):
     def extract_citations_from_html(self, document):
         html = lxml.html.fromstring(document.content_html)
         for matcher in self.matchers:
-            matcher().markup_html_matches(document.expression_uri(), html)
```
actlikewill, line 30: So how was this working before? 
Now it's clear tat we are getting the citations from the response and attaching to the doc.html. What magic was this did this line doing? :smile: Was it still calling out to an external service?

actlikewill (approved):

peachjam/analysis/citations.py
longhotsummer: The other matchers are all in-process and modify `html` directly. Because we're calling an external service now, it can't modify the parsed HTML tree directly. So it can only return a new tree.

nickmwangemi (approved):

PR #923, opened 2023-02-10
Fixes doclist nature filter facet

actlikewill:
- The logic for rendering the nature facet was faulty.

### DIFF
--- a/peachjam/views/generic_views.py
+++ b/peachjam/views/generic_views.py
@@ -10,7 +10,6 @@ from peachjam.models import (
     GenericDocument,
     Judgment,
     LegalInstrument,
-    Legislation,
     Predicate,
     Relationship,
 )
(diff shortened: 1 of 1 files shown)


peachjam/views/generic_views.py
```diff
@@ -56,6 +56,16 @@ def get_context_data(self, **kwargs):
         courts = []
         natures = []
         # Initialize facet data values
+        if self.model in [GenericDocument, LegalInstrument, Legislation]:
```
longhotsummer, line 59: Can we now simply do this in all cases, since all document models have natures?

longhotsummer: Even if they're potentially None?

actlikewill: Yes we can. Changed.

longhotsummer (approved):

PR #925, opened 2023-02-10
Locality facet in document filters

### LINKED ISSUE #900: Add locality facet to the AGP /legislation page
We also want `nature`, but that will be done via #901

actlikewill:
- Adds locality facet to document filters

image

closes #900


actlikewill:
@longhotsummer This adds the facet to all list views. Is this okay or should I check only for legislation as per the ticket?

longhotsummer:
> @longhotsummer This adds the facet to all list views. Is this okay or should I check only for legislation as per the ticket?

I haven't thought through the consequences for all views, so if you can populate this facet only for AGP's legislation view (no other legislation view), let's start there.

actlikewill:
> > @longhotsummer This adds the facet to all list views. Is this okay or should I check only for legislation as per the ticket?
> 
> I haven't thought through the consequences for all views, so if you can populate this facet only for AGP's legislation view (no other legislation view), let's start there.

@longhotsummer I've subclassed the LegislationListView for Africanlii and added the localities facet.

longhotsummer (approved):

PR #926, opened 2023-02-10
Store text content in the db

longhotsummer:
* store text content in a separate model field. This means that when we re-index into ES or extract citations, we don't have to download and re-extract text from PDFs
* index mnc and case numbers into ES

Later we probably want to move content_html and toc_json onto this object, too. Those are large fields which don't need to be loaded in listing views, and only need to be loaded in detail views. Having them on a separate model is Django's recommended mechanism for avoiding expensive loads of large fields.

Once this is deployed, the first time we re-index for search it will extract the text and store it. The next time, it will simply use the existing text. The same applies to citation extraction.


actlikewill:
Awesome.

actlikewill (dismissed):
Great. I wonder if we can measure how much faster indexing is going to be.

PR #930, opened 2023-02-13
Gazettes bug fix

### LINKED ISSUE #927: Gazette adapter must support locality
Currently, the gazette adapter filters on the full jurisdiction code, making something like `AA-ECOWAS` impossible.

Instead, it should split the jurisdiction setting on hyphen, if present, and use the second part to filter on the locality code.

https://github.com/laws-africa/peachjam/blob/main/peachjam/adapters/gazettes.py#L28

nickmwangemi:
Resolves #927


peachjam/adapters/gazettes.py
```diff
@@ -23,13 +23,37 @@ def check_for_updates(self, last_refreshed):
             f"Checking for new gazettes from Gazettes.Africa since {last_refreshed}"
         )
 
-        new_gazettes = (
-            CoreDocument.objects.filter(
-                doc_type="gazette", jurisdiction=self.jurisdiction
+        if "-" not in self.jurisdiction and "*" not in self.jurisdiction:
```
longhotsummer, line 26: You've got a lot of duplication here, which you want to avoid, because it's prone to bugs and more difficult to update. Remember that a queryset can easily have extra filters added to it once it's created.

Instead:

1. split `self.jurisdiction` on "-" and get a juri code and a locality code (default locality to None if no "-")
2. create the queryset and filter the queryset on the juri code
3. check the locality code and filter the queryset accordingly, if necessary

longhotsummer (changes requested):
Don't try to do the queryset all in one go. Instead, build it up incrementally.

longhotsummer:
For another example, see here: https://github.com/laws-africa/peachjam/blob/main/peachjam/forms.py#L113

peachjam/adapters/gazettes.py
```diff
@@ -23,15 +23,34 @@ def check_for_updates(self, last_refreshed):
...
+            locality = self.jurisdiction[1]
+
+            # locality code present
+            if locality != "*":
+                queryset = queryset.filter(
+                    jurisdiction=jurisdiction, locality__code=locality
+                )
+
+            # fetch all localities for this jurisdiction
+            if locality == "*":
```
longhotsummer, line 47: You don't need this if, just make it an else.

peachjam/adapters/gazettes.py
```diff
@@ -23,15 +23,34 @@ def check_for_updates(self, last_refreshed):
...
             .non_polymorphic()
             .using("gazettes_africa")
         )
+
+        if "-" not in self.jurisdiction:
+            # locality code not specified hence None
+            queryset = queryset.filter(jurisdiction=self.jurisdiction, locality=None)
+        else:
+            self.jurisdiction.split("-")
+            jurisdiction = self.jurisdiction[0]
```
longhotsummer, line 37: You *always* want to filter on jurisdiction code, you can do that before the ifs. Then the if statement is just about locality.

longhotsummer (approved):
nice

longhotsummer:
@nickmwangemi once this is deployed, the gazette ingestor for lawlibrary will need to be updated to `ZA-*`.

PR #931, opened 2023-02-13
Legislation listing uses work year, not document year

longhotsummer:
Before:

image

After:

image

image

Ref #889


actlikewill (approved):
Works well

PR #935, opened 2023-02-15
Djlint

actlikewill:
I  tested DJlint against peachjam's templates and this is how it reformatted them. It fixed some things but I also had to fix some things by hand. Let me know what you guys think of the reformatting.

Here is a link to the rules and documentations: [https://www.djlint.com/docs/linter/](https://www.djlint.com/docs/linter/)


longhotsummer:
I think this will definitely help keep our templates consistent. It has my vote! Let's merge it only when we don't have any major open PRs (start of a sprint), because it's a pretty intrusive change.

longhotsummer (approved):

nickmwangemi (approved):

PR #936, opened 2023-02-15
provincial legislation template must use normal legislation template

longhotsummer:
Otherwise we don't get the full functionality required for pocketlaw.

With javascript disabled (required for pocketlaw):

Before:
image

After:
image


actlikewill (approved):
works

PR #939, opened 2023-02-16
Tweak gazettes on agp

nickmwangemi:
This PR tweaks how gazettes are displayed on AGP. 

Before:
image

image

image

After:
image

image

image

https://www.loom.com/share/d31dfe1ad1e140419002a9ba8b9eb179


peachjam/views/gazette.py
```diff
@@ -36,13 +37,27 @@ class GazetteListView(TemplateView):
...
+            qs = qs.filter(locality=self.locality)
+        return qs
 
     def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
+        context = super().get_context_data(locality=self.locality, **kwargs)
 
         queryset = self.get_queryset()
+        localities = list(Locality.objects.all())
+        gazettes = queryset.filter(locality__in=localities)
```
longhotsummer, line 55: Consider this in context. When viewing South Africa, this loads 60,000 gazettes and iterates over all of them, just to add 9 localities to the set. It's very inefficient. It's also going to be doing it on a page where locality is set, even though it's never used.

Here's a better query:

```
>>> str(qs.order_by().distinct("locality").values_list("locality", flat=True).query)
'SELECT DISTINCT ON ("peachjam_coredocument"."locality_id") "peachjam_coredocument"."locality_id" FROM "peachjam_gazette" INNER JOIN "peachjam_coredocument" ON ("peachjam_gazette"."coredocument_ptr_id" = "peachjam_coredocument"."id")'
>>> list(qs.order_by().distinct("locality").values_list("locality", flat=True))
[6, None]
```

So instead,

1. only calculate this when `self.locality` is None
2. use this to get distinct locality ids: `list(qs.order_by().distinct("locality").values_list("locality", flat=True))`

longhotsummer (changes requested):
Nice!

Some query changes, please.

Are the specialised LawLibraryGazette* views still necessary after this?

nickmwangemi:
> Nice!
> 
> Some query changes, please.
> 
> Are the specialised LawLibraryGazette* views still necessary after this?

The LawLibrary-specific views can fall off

lawlibrary/templates/peachjam/gazette_detail.html
```diff
@@ -6,7 +6,7 @@
     <ol class="breadcrumb">
```
longhotsummer, line 6: this file is now identical to peachjam's gazette_detail so the lawlibrary one can fall away

PR #940, opened 2023-02-16
External document model

### LINKED ISSUE #759: As an advanced user of the AGP website, I'd like to access relevant Afrobarmeter content linked to AU documents
AGP website editors should be able to manually link external documents to all AGP document content.

actlikewill:
- [x] Add external document model
- [x] Setup redirect from source_url
- [x] Add admin support with sensible defaults

image

closes #759


peachjam/models/external_document.py
```diff
@@ -0,0 +1,21 @@
...
+from .core_document_model import CoreDocument
+
+
+class ExternalDocument(CoreDocument):
+
+    frbr_uri_doctypes = ["doc"]
+
+    def clean(self):
+        self.doc_type = "external"
+        self.frbr_uri_number = slugify(self.title)
```
longhotsummer, line 14: It's better if we do this in the model admin form using `    prepopulated_fields = {"slug": ("title",)}` because then it fills itself in for the user.

Also setting the language and date defaults in the model admin form means the user doesn't have to think about it and leave it blank.

peachjam/models/external_document.py
```diff
@@ -0,0 +1,21 @@
...
+        self.frbr_uri_doctype = "doc"
```
longhotsummer, line 15: This is already enforced in the model and (I think) the admin because `frbr_uri_doctypes` is set on the model.

peachjam/models/external_document.py
```diff
@@ -0,0 +1,21 @@
...
+
+from .core_document_model import CoreDocument
+
+
+class ExternalDocument(CoreDocument):
+
+    frbr_uri_doctypes = ["doc"]
+
+    def clean(self):
+        self.doc_type = "external"
```
longhotsummer, line 13: This is already done in the model.

longhotsummer (changes requested):
Don't forget the peachjam_search aspect to prevent indexing for search.

peachjam/models/external_document.py
actlikewill: I'm setting these values here on the clean method rather than the save method to allow the  validations to pass,

longhotsummer: If you pre-populate the form, then the validation will pass automatically. That's the best place to provide defaults, I think. Doing them here means the user can't override them at all.

peachjam/models/external_document.py
actlikewill: Changed

peachjam/models/external_document.py
actlikewill: Changed

longhotsummer (approved):

PR #949, opened 2023-02-21
Add new feature issue template and add court registry field to judgment metadata section

### LINKED ISSUE #943: Create a metadata field for Registry - controlled vocabulary - in the Judgment content type
the alternative is to use locality, but this is not strictly accurate.
This is a block in Tanzania and a request in Uganda.

Update importer to include the field so migration on TanzLII can continue

nickmwangemi:
This PR adds:
- a new `adding_new_fields_to_a_document_type` issue template 
- a new `CourtRegistry` model that is manageable from the admin panel.
- adds the Court registry field to the metadata section of the judgment detail view. 
- adds the court registry as a filter facet on the court judgment listing page.
- adds as a court registry search facet in search UI.
- adds the court registry field to elasticsearch indexing.

To test locally:
- apply migrations: `python manage.py migrate`.
- create court registry objects in the admin section.
- assign a court registry to sample judgments.
- rebuild local elasticsearch index: `python manage.py search_index --rebuild`.
- test filter and search facets.

Closes #943 

Assign a court registry to a judgment:
image

Court registry field in metadata section of the judgment detail view:
image

Court registry filter facet:
image

Court registry search facet:
image


peachjam/forms.py
```diff
@@ -92,11 +92,14 @@ class BaseDocumentFilterForm(forms.Form):
...
     natures = forms.CharField(required=False)
     localities = forms.CharField(required=False)
+    registeries = forms.CharField(required=False)
 
     def __init__(self, data, *args, **kwargs):
         self.params = QueryDict(mutable=True)
         self.params.update(data)
 
+        print(self.params)
+
```
actlikewill, line 102:
```suggestion
```

peachjam/forms.py
```diff
@@ -92,6 +92,7 @@ class BaseDocumentFilterForm(forms.Form):
     judges = forms.CharField(required=False)
     natures = forms.CharField(required=False)
     localities = forms.CharField(required=False)
+    registeries = forms.CharField(required=False)
```
longhotsummer, line 95:
```suggestion
    registries = forms.CharField(required=False)
```

peachjam/models/judgment.py
```diff
@@ -72,10 +72,31 @@ def __str__(self):
         return self.name
 
 
+class CourtRegistry(models.Model):
+    court = models.ForeignKey(
+        Court,
+        on_delete=models.CASCADE,
+        null=True,
+        related_name="court_registry",
```
longhotsummer, line 80: This should be plural, eg for `court.court_registries`

peachjam_search/views.py
```diff
@@ -226,6 +227,7 @@ class DocumentSearchViewSet(BaseDocumentViewSet):
         "content": None,
         "court": None,
         "alternative_names": {"boost": 4},
+        "registry": {"boost": 4},
```
longhotsummer, line 230:
```suggestion
        "registry": None,
```

No need to boost it

peachjam/models/judgment.py
```diff
@@ -72,10 +72,31 @@ def __str__(self):
...
+    class Meta:
+        verbose_name = _("court registry")
+        verbose_name_plural = _("court registries")
+        unique_together = ("court", "name")
+
+    def __str__(self):
+        return f"{self.name} Registry - {self.court}"
+
+
 class Judgment(CoreDocument):
```
longhotsummer, line 93: When saving a judgment, if there is a registry set, then also force the court to be `self.registry.court`, so that they always align. Do it at the start of `save()`

peachjam/models/judgment.py
```diff
@@ -72,10 +72,31 @@ def __str__(self):
...
+    )
+    name = models.CharField(_("name"), max_length=1024, null=False, blank=False)
+
+    class Meta:
+        verbose_name = _("court registry")
+        verbose_name_plural = _("court registries")
+        unique_together = ("court", "name")
+
+    def __str__(self):
+        return f"{self.name} Registry - {self.court}"
```
longhotsummer, line 90:
```suggestion
        return f"{self.name} - {self.court}"
```

Not all registries are called as such. eg. in Kenya they're court stations.

peachjam/models/judgment.py
longhotsummer: Actually, I think `registries` would be better. We have `judgment.registry`, it should use similar naming.

peachjam/models/judgment.py
nickmwangemi: Note to self: So that we don't have a situation whereby a judgment is assigned a registry that doesn't belong to its court.

peachjam/views/courts.py
```diff
@@ -52,9 +52,16 @@ def get_context_data(self, **kwargs):
             .order_by("judges__name")
             .values_list("judges__name", flat=True)
             .distinct()
+            .exclude(judges=None)
```
longhotsummer, line 55: We're not actually calculating these facets correctly. In a view like https://agp.africanlii.org/legal-instruments/ we calculate the facet values after applying the filters, which helps the user drill down. In this view, we don't, and so applying one facet doesn't impact the others (which is incorrect). Now that we'll have more than one facet, it'll be more obvious.

We should be using the same pattern that is used [here](https://github.com/laws-africa/peachjam/blob/main/peachjam/views/generic_views.py#L58) where we get the base queryset, apply filters (with an exclude), and then count the results.

At the same time, we can optimise that query to return distinct values rather than doing that in python.

peachjam/models/judgment.py
```diff
@@ -72,10 +72,31 @@ def __str__(self):
         return self.name
 
 
+class CourtRegistry(models.Model):
+    court = models.ForeignKey(
```
longhotsummer, line 76: We'll need a code on this model. I suggest we make it `f'{self.court.code}-{slugify(self.name)}` which will be unique, and which we can set on save. That will be necessary for import/export support.

peachjam/models/judgment.py
```diff
@@ -72,10 +73,42 @@ def __str__(self):
...
+    class Meta:
+        verbose_name = _("court registry")
+        verbose_name_plural = _("court registries")
+        unique_together = ("court", "name")
+
+    def __str__(self):
+        return f"{self.name} - {self.court}"
+
+    def save(self, *args, **kwargs):
+        if not self.code:
```
longhotsummer, line 96: Always set the code, in case the name has changed.

peachjam/resources.py
```diff
@@ -442,6 +443,20 @@ def get_case_numbers(row):
 
         return case_numbers
 
+    def before_import_row(self, row, **kwargs):
+        super().before_import_row(row, **kwargs)
+
+        if row.get("registry"):
```
longhotsummer, line 449: The registry should be the full code, such as `"kehc-high-court-at-foo"`. So your get_or_create() should use the code as the main lookup, and the other two fields as `defaults={}`.

longhotsummer (approved):

PR #975, opened 2023-03-02
Add coredocument to admin

### LINKED ISSUE #878: As an admin I want to be able to re-index bulk documents for search, in the background
* add a django model admin command to "re-index for search in background" which simply calls `peachjam_search.tasks.search_model_saved('peachjam.CoreDocument', pk)` which sets up the background task 
* would be useful to be able to do this for all documents via the CoreDocument admin listing, which currently doesn't exist. We should add this, but not allow all editors the perms to make changes to core documents

actlikewill:
This adds the core document model to the admin. This will add all current admin bulk functionality for the core document model, including re-extract content from docx, re-extract citations and reindex documents.
- [x] Add coredocument to admin
- [ ] Remove coredocument permissions from document editors

closes #878

### DIFF
--- a/peachjam/admin.py
+++ b/peachjam/admin.py
@@ -35,6 +35,7 @@ from peachjam.models import (
     Book,
     CaseNumber,
     CitationLink,
+    CoreDocument,
     Court,
     CourtClass,
     CourtRegistry,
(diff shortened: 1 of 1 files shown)


actlikewill:
@longhotsummer I've reused the DocumentAdmin class that has reindex functionality along with the reextraction bulk actions. We need to remove coredocument permissions from document editors which I want to do when this is merged.

longhotsummer:
> @longhotsummer I've reused the DocumentAdmin class that has reindex functionality along with the reextraction bulk actions. We need to remove coredocument permissions from document editors which I want to do when this is merged.

Sounds good.

I was going to suggest that all the form fields should be readonly, because otherwise we're potentially bypassing the per-document forms. But maybe that's a useful thing for admins? No-one needs perms to edit these; even viewing should be for advanced editors only.

longhotsummer:
@actlikewill is this ready for review yet?

PR #976, opened 2023-03-03
get/set raw value on dynamic file field

longhotsummer:


actlikewill:
@longhotsummer I keep getting this error locally. @nickmwangemi doesn't seem to have it though. 
image

Also the documents show up with 0kb i.e they are not being read at all from aws. We were using sierralii production credentials

longhotsummer:
it looks like an error when trying to extract text from the PDF with our custom pdf-to-text javascript code, which runs via node.

That doesn't look like the full error, so I'm not sure what the cause is.

Can you ensure `npm install` is up to date?

longhotsummer:
Since it looks like the pdfjs issue is unrelated, can this be merged?

PR #983, opened 2023-03-06
Fix:Pagination bug skipping items

### LINKED ISSUE #929: Bug: is pagination skipping items
When filtering by letter and year, "Lebashe" shows up:

image

Without the letter filter, it looks like Lebashe falls on the pagination boundary and is lost:

image

image

These are all the 24 October documents:

image

actlikewill:
- This one was a little tricky to reproduce.
- This seems to be a problem with how Django is handling pagination and ordering. The SQL works by setting limits and offsets depending on the pagination size. For some reason, some items get lost if they have similar dates and fall across the offset boundary.
- By adding another ordering clause for `title` this problem seems to go away. 
- I still don't fully understand the cause for this, but this fix seems to work, at least locally.

Loom: [https://www.loom.com/share/bcadd61bc8904cacba55842095f9c7c6](https://www.loom.com/share/bcadd61bc8904cacba55842095f9c7c6)
fixes #929

### DIFF
--- a/peachjam/forms.py
+++ b/peachjam/forms.py
@@ -112,7 +112,7 @@ class BaseDocumentFilterForm(forms.Form):
         registries = self.params.getlist("registries")
 
         # Order by date descending initially
-        queryset = queryset.order_by("-date")
+        queryset = queryset.order_by("-date", "title")
 
         if years and exclude != "years":
             queryset = queryset.filter(date__year__in=years)


peachjam/forms.py
```diff
@@ -112,7 +112,7 @@ def filter_queryset(self, queryset, exclude=None):
         registries = self.params.getlist("registries")
 
         # Order by date descending initially
-        queryset = queryset.order_by("-date")
+        queryset = queryset.order_by("-date", "title")
```
longhotsummer, line 115: Nice catch -- my guess is that ordering of rows with the same date is not guaranteed, and so can change. By ordering by date and then title, we guarantee a consistent ordering.

longhotsummer: https://stackoverflow.com/questions/67343360/is-order-by-stable-for-several-rows-with-the-same-key-values

longhotsummer (approved):

PR #1005, opened 2023-03-13
Walk results pages in indigo adapter

### LINKED ISSUE #1004: Walk paginated list of works from indigo in indigo adapter
https://github.com/laws-africa/peachjam/blob/main/peachjam/adapters/adapters.py#L87

Keep loading the `next` url until it is null, appending the `results` from each page into a big list.

image

actlikewill:
closes #1004


longhotsummer (commented):
cool

PR #1007, opened 2023-03-13
Faceting on registries is breaking search

longhotsummer:
We need to re-build the search index with updated mappings to allow faceting on registry, because it must be created in the mapping as a keyword field.

Without this, search results for most queries are coming back empty or very limited.

To easily re-index all documents, we need to delete and re-creating the mapping, and then run background tasks to re-index for search. That's easiest once https://github.com/laws-africa/peachjam/pull/975 is merged.

### DIFF
--- a/peachjam_search/views.py
+++ b/peachjam_search/views.py
@@ -216,7 +216,7 @@ class DocumentSearchViewSet(BaseDocumentViewSet):
         "nature": "nature",
         "year": "year",
         "judges": "judges",
-        "registry": "registry",
+        # "registry": "registry",
         "attorneys": "attorneys",
     }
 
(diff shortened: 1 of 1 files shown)


longhotsummer:
I actually think we need to build on the existing reindex management command to queue up background refresh tasks, rather than doing it through entirely separate places. It needs to be easy to do this.

actlikewill (approved):

actlikewill:
> I actually think we need to build on the existing reindex management command to queue up background refresh tasks, rather than doing it through entirely separate places. It needs to be easy to do this.

Were else do we do this apart from the reindex command?

actlikewill:
Oh, we want the reindex command to also rebuild the mapping like when we run `python manage.py search_index --rebuild`

Issue #1025, opened 2023-03-16
Add a menu item for Speeches & Law Reform Reports on ZambiaLII


NotoriousMBB:
To display the Speech nature of generic documents
https://new.zambialii.org/admin/peachjam/genericdocument/

![image](https://user-images.githubusercontent.com/46931873/225558123-1fe58b9d-b151-4f34-9988-99cff528a7ee.png)

longhotsummer:
@nickmwangemi this should be two separate menu items:

1. one for "Speeches" - /doc/speech (documents exist)
2. one for "Law Reform Reports" - /doc/law-reform-report (document's don't seem to exist yet)

The document natures are here: https://new.zambialii.org/admin/peachjam/documentnature/

PR #1030, opened 2023-03-16
ZambiaLII Tweaks

nickmwangemi:
Social Media:
image

image

Right-hand sidebar logos:

image

Speeches & Law Reform Reports:
image

image

Increase file size upload limit:
image


agp.conf
longhotsummer, line 1: @nickmwangemi this isn't agp specific, so let's call this peachjam.conf instead

niiroobiro: @nickmwangemi Law reform Reports is separate from Speeches. These are 2 different items on that menu.

longhotsummer:
@nickmwangemi I've clarified the menu items issue in https://github.com/laws-africa/peachjam/issues/1025

zambialii/templates/liiweb/home.html
```diff
@@ -47,24 +47,101 @@ <h4 class="mb-4">
...
+    </div>
+    <div class="col-md-6 mb-3">
+      <img height="26"
+           src="{% static 'images/indigo-logo.jpg' %}"
+           alt="indigo-logo"/>
+    </div>
+  </div>
+{% endblock %}
+{% block social-media %}
+  <ul class="nav nav-tabs about-tabs d-flex justify-content-center justify-content-lg-start"
```
longhotsummer, line 82: This should definitely be a re-usable include template at this point which you can pass parameters into to determine what twitter and facebook profiles to use.

Otherwise we're duplicating lots of code all over the place.

longhotsummer (changes requested):

liiweb/templates/liiweb/_social_media.html
```diff
@@ -0,0 +1,66 @@
...
+       id="twitter-tab-content"
+       role="tabpanel"
+       aria-labelledby="twitter-tab">
+    <div class="d-flex justify-content-center justify-content-lg-start">
+      <a class="twitter-timeline"
+         data-width="248"
+         data-height="400"
+         href="https://twitter.com/{{ twitter }}?ref_src=twsrc%5Etfw">
+        {% blocktrans %} Tweets by {{ tweets_by }} {% endblocktrans %}
+      </a>
```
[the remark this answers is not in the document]
nickmwangemi: No. Have a better approach in mind?

actlikewill (approved):

PR #1053, opened 2023-03-28
Show judges in document detail page for all LIIs

nickmwangemi:
This PR adds the judges field to the metadata section of the document detail page.

image


lawlibrary/templates/peachjam/layouts/document_detail.html
longhotsummer, line 1: I think best to delete this whole file since it's not doing anything?

longhotsummer (approved):

PR #1054, opened 2023-03-28
Fix "amendment not yet applied" notice

nickmwangemi:
Before:
image

After:
image


peachjam/templates/peachjam/_timeline_events.html
```diff
@@ -11,7 +11,7 @@ <h5 class="mb-0">
               <span class="badge rounded-pill bg-primary">{% trans 'this version' %}</span>
             {% endif %}
             {% for event in history_item.events %}
-              {% if event.unapplied_amendment %}
+              {% if event.unapplied_amendment and history_item.date > latest_amendment_date %}
```
longhotsummer, line 14: Determining if it's an unapplied amendment is a business logic issue and should live in the python code, wherever `unapplied_amendment` is calculated.

longhotsummer (changes requested):
The logic isn't quite right, yet.

peachjam/templates/peachjam/_timeline_events.html
goose-life: I agree that it should be after the latest expression date. Apologies if my write-up was unclear on that, Nick. I think I used 'document' where I should have used 'expression'. 

And yes, looks like `point_in_time_dates` has dates for actual expressions as opposed to just timeline dates.

I would probably change https://github.com/laws-africa/peachjam/blob/main/peachjam/views/legislation.py#L202 to include that extra condition.

nickmwangemi:
> The logic isn't quite right, yet.

Noted. Working on this.

goose-life (approved):
nice!

peachjam/views/legislation.py
```diff
@@ -192,18 +192,30 @@ def get_timeline_events(self):
...
+
+                event.append(
+                    {
+                        "date": event_date,
+                        "event": "amendment",
+                        "amending_title": amendment.get("amending_title"),
+                        "amending_uri": amendment.get("amending_uri"),
+                        "unapplied_amendment": bool(
+                            amendment.get("date") not in point_in_time_dates
+                            and converted_event_date > converted_latest_expression_date
```
longhotsummer, line 215: Pro-tip: you don't need to parse the dates. They're in ISO8601 format `YYYY-MM-DD`. The textual comparison of the dates is the same as the date-value comparison. You already make use of this with `max(point_in_time_dates)` above. That's why ISO8601 is so cool.

So you can just compare the string values directly.

So you can actually go back to the original list comprehension and just add `and amendment.get("date") > latest_expression_date`

longhotsummer (changes requested):
No need to do any date parsing. I think using the existing list comprehension is fine, it just needs the extra condition.

longhotsummer (approved):

PR #1061, opened 2023-03-29
Set default langugages 

### LINKED ISSUE #1057: As a user on TanzLII site, I want to be able to be able to choose between Swahili and English.
The TanzLII site should default to Swahili.
We need users to select between English and Swahili for TanzLII. This will ensure that the site is displayed in the correct language and that users can navigate and interact with it effectively.

actlikewill:
- Set default languages

closes #1057


longhotsummer (changes requested):
great - pls also delete the effectively blank _set_language.html files in the other lii apps.

you'll also then need to set `LANGUAGES` for the other lii apps, otherwise they will all have language selectors for all languages.

the languages for each app are in https://docs.google.com/spreadsheets/d/1aneRQdmNGCTU1laVSXQ9IoSBjmAKr3j6kwUv7ScaaTg/edit?pli=1#gid=0

longhotsummer (dismissed):

PR #1062, opened 2023-03-29
fix history events for stub docs without points in time

longhotsummer:


peachjam/views/legislation.py
```diff
@@ -187,28 +180,34 @@ def get_timeline_events(self):
...
-                    "amending_title": amendment.get("amending_title"),
-                    "amending_uri": amendment.get("amending_uri"),
-                    "unapplied_amendment": bool(
-                        amendment.get("date") not in point_in_time_dates
-                        and amendment.get("date") > latest_expression_date
-                    ),
-                }
-                for amendment in amendments
-            ]
+            latest_expression_date = (
```
longhotsummer, line 190: this is the key change -- if there are no points in time, then the latest date is the date of the document we're looking at (which will be a stub)

nickmwangemi (approved):

PR #1066, opened 2023-03-30
Use configured label for author field in search and listing views

nickmwangemi:
AGP

**search facet**
image

**filter facet on listing view**
image

LawLibrary
image


longhotsummer (approved):
nice

PR #1071, opened 2023-03-31
citation popups

longhotsummer:
Add wikipedia-style popups for citations in document body, using the `la-decorate-external-refs` widget from law-widgets.

1. for Akoma Ntoso, this works out the box
2. for PDF and HTML, we need to first add some attributes to the `<a>` tags so that the law-widgets picks them up

https://www.loom.com/share/021f935d5f6e4080b5a44d52aee3fc70


actlikewill (approved):
The size and information contained in the popup is not so much that it is distracting, which was my main concern. This works well.

nickmwangemi (approved):

Issue #583, opened 2022-10-24
As an advanced search user I want to be able to express boolean term queries


longhotsummer:
Boolean query strings which allows for "foo AND (bar OR baz) AND NOT boom" style queries.

We use QS's simple query string syntax which allows this: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-simple-query-string-query.html

We can help the user build a query string like this, rather than expecting them to understand it. Google has a simple but effective advanced search page which we could learn from: https://www.google.co.za/advanced_search

longhotsummer:
# Proposal

We adjust the "Advanced search" dialog on the search page as follows:

1. make it wider to take up the full width
2. one row per field (including "Any field" at the top)
3. each field (except the dates field) supports advanced boolean queries
4. date field remains as-is

So each field looks something like this:

```
Title

All these words: .....
This exact word or phrase: ....
Any of these words: ...
None of these words: ...
```

## Building the query string:

When building the query, use this logic. Each step tokenises the input, manipulates it, and appends the output to the complete query string.

Tokenize using: `tokens = str.match(/([^\s"]+|"[^"]*")+/g)`

* all: tokenize and join with ` AND `
* exact: wrap each token in quotes, if they're not already wrapped; join with ` AND `
* any: tokenize and join with ` OR ` and wrap the whole thing in brackets
* none: tokenize and join with ` OR `, then change to `NOT ( ... )`

Some examples:

* all: `foo bar` -> `foo AND bar`
* exact: `"foo bar" baz" -> `"foo bar" AND "baz"`
* any: `foo bar` -> `(foo OR bar)`
* none: `foo bar` -> `NOT (foo OR bar)`

PR #1075, opened 2023-04-03
As an advanced search user I want to be able to express boolean term queries

Sandravaphilips:
- Issue https://github.com/laws-africa/peachjam/issues/583
- This is what I have now: https://www.loom.com/share/401417f27b7a4b688dd45e63fb4ef7d3

What I've done:
- Search works as expected for every field.
- Search and input fields persist on page refresh.
- Increased the total width to max-width

Things yet to do:
- Add input descriptions, so visitors know how to use the advanced search.
- Add the fields for global search.
- Clean up code and add comments where necessary 
- _disableDate_ needs to use _formatFieldValues_. 
- Also need to remove the spaces between search inputs to prevent +- for example, when serializing state in the URL


peachjam/js/components/FindDocuments/AdvancedSearchFields.vue
```diff
@@ -0,0 +1,112 @@
...
+          type="text"
+          class="form-control"
+          :aria-describedby="$t(formTitle)"
+          @input="e => updateSubfields(e, 'any')"
+        >
+      </div>
+    </div>
+    <div class="row mt-3">
+      <div class="col-sm-2">
+        <label :for="`${inputName}-none`">{{ $t("None of these words:") }}:</label>
```
longhotsummer, line 59:
```suggestion
        <label :for="`${inputName}-none`">{{ $t("None of these words") }}:</label>
```

peachjam/js/components/FindDocuments/AdvancedSearchFields.vue
```diff
@@ -0,0 +1,112 @@
+<template>
+  <div class="">
+    <div class="form-title">
```
longhotsummer, line 3: Prefer an h* tag, rather than custom styling, so it's consistent across the site. Perhaps `h4`?

peachjam/js/components/FindDocuments/index.vue
```diff
@@ -475,7 +498,20 @@ export default {
         if (params.has('date_to')) this.advancedFields.date.date_to = params.get('date_to');
         advancedSearchFields.forEach(key => {
           if (!params.has(key)) return;
-          this.advancedFields[key] = params.get(key);
+          const fieldValue = params.get(key);
+          const splitValue = fieldValue.match(/[^\s"']+|['"][^'"]*["']+/g);
+
+          splitValue.forEach((value, index) => {
```
longhotsummer, line 504: This isn't quite the right place to do this. The inputs the user provided to the advanced search boxes don't change. Instead, when generating the query string to send to the server, we transform the advanced search fields.

In the case where the user has entered boolean terms for the "all fields" field (not yet in your UI), then the `search` param would be updated to produce, for example, `foo -"bar baz"`.

If the user has entered boolean terms for the "title" field, then the `search__title` param would be updated appropriately, etc.

peachjam/js/components/FindDocuments/index.vue
```diff
@@ -518,9 +545,38 @@ export default {
...
+        if (fieldKey === 'all') {
+          splitValue = splitValue.join(' ');
+        } else if (fieldKey === 'exact') {
+          splitValue = splitValue.map(value => {
+            if (value.startsWith('"') || value.startsWith("'")) return value;
+            else return `"${value}"`;
+          });
+          splitValue = splitValue.join(' ');
+        } else if (fieldKey === 'any') {
+          splitValue = splitValue.join('|');
```
longhotsummer, line 564: I think this must all be wrapped in quotes, since the default operator is AND.

So: `foo (bar | baz)` rather than `foo bar | baz` which is actually equivalent to `(foo AND bar) or baz` because of the boolean precedence of AND and OR.

longhotsummer: In other words, if the user enters:

```
all: foo
any: bar baz
```

the search string should be: `foo (bar | baz)`

Sandravaphilips:
Hi Greg. 
This is what I have now:
https://www.loom.com/share/e3118c4e395e42aab665396f8b2a4574

Things to note:
- I didn't expand the main fields, because it looked like too much was happening primarily because we're using two columns. Felt it'll be better to put them beneath.
- The URL is back to using the formatted query. I hope that's okay since users can now input the same query in the fields.
- On refresh, only the main fields are populated. If the user inputs a value in the additional fields, it overrides the output of the main one. Populating additional fields on mount will mean destructing the formatted string back to the normal state.

longhotsummer (approved):

PR #1089, opened 2023-04-05
add a custom tag to background tasks

longhotsummer:
this allows us to filter bg tasks when applying alert rules.

As per https://docs.sentry.io/platforms/python/enriching-events/tags/

Without this, I cannot see a way to allow us to differentiate between http requests and background tasks when setting up alert rules.


actlikewill (approved):
Great.

nickmwangemi (approved):

PR #1090, opened 2023-04-06
Handle docs with empty work amendments

nickmwangemi:


peachjam/views/legislation.py
```diff
@@ -54,13 +54,13 @@ def get_notices(self):
...
-        work_amendments_dates = [
-            work_amendment["date"] for work_amendment in work_amendments
-        ]
 
-        latest_amendment_date = max(work_amendments_dates)
+        if work_amendments:
+            work_amendments_dates = [
+                work_amendment["date"] for work_amendment in work_amendments
+            ]
+            latest_amendment_date = max(work_amendments_dates)
```
longhotsummer, line 63: `latest_amendment_date` now won't be declared for use in the if statement below. It should default to None if there are no amendments.

longhotsummer (approved):

PR #1091, opened 2023-04-11
Redirect www to non-www

longhotsummer:
eg. www.lawlibrary.org.za redirects to lawlibrary.org.za, so that we don't have duplicate websites.

### DIFF
--- a/peachjam/middleware.py
+++ b/peachjam/middleware.py
@@ -1,0 +1,20 @@
+from django.shortcuts import redirect
+
+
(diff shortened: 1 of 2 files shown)


actlikewill:
@longhotsummer What is the importance of doing this? Is it an analytics thing?

actlikewill (approved):

longhotsummer:
Yes, for analytics, and so to avoid confusing search engines and diluting our traffic.

PR #1092, opened 2023-04-11
Document deletion in the ingestor

### LINKED ISSUE #304: Document ingestors need a way to identify documents deleted at their source that also need to be deleted locally
## Proposal

- We need to add a mechanism to compare what we have in the local database to what we have at the remote ingestor source and keep them in sync. 
#### Indigo Adapter
- At the point where gather the `points in time` urls for updating, we can add a check for existing related works using the Work model and compare them with the list of urls in the point in time object. 
- If we have existing documents for the Work that are not in the point in time object, we can delete them

#### Gazettes Adapter

- For gazettes, we can do a straight comparison for values which are in the current database that do not exist in the Gazettes.africa database for the current jurisdiction. We delete the ones that we have in the database that do not exist in Gazettes.africa database.

actlikewill:
- Adds document deletion in the ingestor.

closes #304


actlikewill:
@longhotsummer Turns out we already had a document deletion method in the adapter that we used with the indigo webhook

https://github.com/laws-africa/peachjam/blob/337e81a0367c449b4c8a115820a723db53cd4504/peachjam/adapters/adapters.py#L352-L365

peachjam/adapters/adapters.py
```diff
@@ -92,10 +96,23 @@ def get_doc_list(self):
...
 
-    def get_updated_documents(self, last_refreshed):
+    def check_for_deleted(self, docs):
+        uris = {d["expression_frbr_uri"] for d in docs}
+        for doc in docs:
+            for pit in doc["points_in_time"]:
+                for expr in pit["expressions"]:
+                    uris.add(expr["expression_frbr_uri"])
+
+        deleted = Legislation.objects.exclude(
```
longhotsummer, line 106: Just remember this must be scoped to only the place (country + locality) that the ingestor is applicable to. eg. AfricanLII we have four separate ingestors (EAC, ECOWAS, SADC, AU). They should only delete "their" legislation.

longhotsummer:
I think we should update the Indigo adapter so we give it an API URL (eg. `https://api.laws.africa/v2` and a place code and it builds the full URL when it needs it. That way we don't need to do the weird `re.sub` to delete the document.

So, the updated and delete calls both get an expression FRBR URI, and the adapter generates a full url with `api_url + expression_frbr_uri`. The call to `get_doc_list()` would build up `api_url + '/' + place_code + '.json'`

peachjam/adapters/adapters.py
actlikewill: I've added the jurisdiction and locality as settings which we can use here.

peachjam/adapters/adapters.py
```diff
@@ -76,26 +76,52 @@ def __init__(self, settings):
...
+    def check_for_deleted(self, docs):
+        uris = {d["expression_frbr_uri"] for d in docs}
+        for doc in docs:
+            for pit in doc["points_in_time"]:
+                for expr in pit["expressions"]:
+                    uris.add(expr["expression_frbr_uri"])
+
+        qs = Legislation.objects.filter(jurisdiction__iso=self.jurisdiction.upper())
+        if self.locality:
+            qs = qs.filter(locality__code=self.locality)
```
longhotsummer, line 114: if there is no locality, you must explicitly filter on `locality=None`.

peachjam/adapters/gazettes.py
```diff
@@ -29,31 +29,42 @@ def check_for_updates(self, last_refreshed):
...
+            updated_qs = updated_qs.filter(jurisdiction=self.jurisdiction.split("-")[0])
 
             # locality code present e.g. "ZA-gp"
             if self.jurisdiction.split("-")[1] != "*":
-                queryset = queryset.filter(
-                    locality__code=self.jurisdiction.split("-")[1]
-                )
+                locality_code = self.jurisdiction.split("-")[1]
+                updated_qs = updated_qs.filter(locality__code=locality_code)
+                deleted_qs = deleted_qs.filter(locality_code=locality_code)
```
longhotsummer, line 53: is the single underscore in `locality_code` correct?

peachjam/adapters/adapters.py
actlikewill: Added. :heavy_check_mark:

peachjam/adapters/gazettes.py
actlikewill: Added :heavy_check_mark:

peachjam/adapters/adapters.py
longhotsummer: did you git push?

actlikewill: sorry, check now

longhotsummer (approved):
Cool, let's try it. Hopefully it doesn't delete everything accidentally!

longhotsummer:
> > Cool, let's try it. Hopefully it doesn't delete everything accidentally!
> 
> lol now you've made me nervous 😅

Anything that automatically could potentially delete everything in a database should definitely make you nervous!

actlikewill:
> > > Cool, let's try it. Hopefully it doesn't delete everything accidentally!
> > 
> > 
> > lol now you've made me nervous sweat_smile
> 
> Anything that automatically could potentially delete everything in a database should definitely make you nervous!

right. let me just double check everything before merging

actlikewill:
@longhotsummer I want to wait for the pdf issue on indigo to resolve to merge this in. I want to test it properly.

longhotsummer:
ok. The PDF issue is resolved.

PR #1095, opened 2023-04-12
Fix frbr uri parse in bulk import

### LINKED ISSUE #1094: Skipping rows on Bulk Imports for Generic documents is not working properly.
Image

actlikewill:
- The generic document bulk importer fails when you skip a row . 
- This adds a check for the work_frbr_uri before parsing the value.
- It also adds a validator to make sure the value exists.

closes #1094

### DIFF
--- a/peachjam/resources.py
+++ b/peachjam/resources.py
@@ -335,6 +335,11 @@ class GenericDocumentResource(BaseDocumentResource):
         attribute="nature",
         widget=DocumentNatureWidget(DocumentNature, field="code"),
     )
+    work_frbr_uri = fields.Field(
+        column_name="work_frbr_uri",
+        attribute="work_fbr_uri",
+        widget=CharRequiredWidget(field="work_frbr_uri"),
+    )
 
     required_fields = (
         "author",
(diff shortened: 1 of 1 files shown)


longhotsummer:
Does this still allow the field to be blank?

actlikewill:
> Does this still allow the field to be blank?

No, we we are not allowing the field to be blank for generic documents.

longhotsummer:
> > Does this still allow the field to be blank?
> 
> No, we we are not allowing the field to be blank for generic documents.

Does it allow row skipping when the field is blank?

Row validation happens before the row is skipped, IIRC, which is annoying.

actlikewill:
> > > Does this still allow the field to be blank?
> > 
> > 
> > No, we we are not allowing the field to be blank for generic documents.
> 
> Does it allow row skipping when the field is blank?
> 
> Row validation happens before the row is skipped, IIRC, which is annoying.

Very annoying. It does allow skipping if the field is blank. The errors may come from the other values which we pull from from the frbr_uri e.g language and country

longhotsummer (approved):

PR #1105, opened 2023-04-14
Search across multiple indexes

### LINKED ISSUE #498: AfricanLII: search across multiple LIIs
# Implementation

* [x] updated search backend
* [ ] updated search frontend
* [x] resolver

# Overview

As a user, I want to be able to search on africanlii.org and find materials from the AU and all our in-country LII partner websites.

For example:

1. if I search on AfricanLII for "constitution" it should find all the constitutions we have across the network
2. if I search for "Makwanyane" it should find "S v Makwanyane" on LawLibrary
3. etc.

# Proposal

## Frontend

I don't think we need to substantially change the search experience, at least as a first iteration. The search interface already supports a jurisdiction facet, and includes the jurisdiction in the search results. I suggest we tweak only the following:

1. adjust wording on the search page to indicate that it searches across multiple websites
2. show a flag icon alongside the country name in the search result, to help visually distinguish the different countries. Our other sites are all single-country and so this hasn't been an issue until now.
  a. what is the flag for AfricanLII or AU materials?
3. when the user clicks on a result, it simply goes to the FRBR URI on the africanlii website.
  a. If the content is actually on the website (eg. AU materials) then the user sees the document, as usual.
  b. if the content is NOT on the website, then an updated resolver (see below) redirects to the appropriate LII website
4. advanced search should have a list of countries to search, supporting multiple choices
5. update search wording on the homepage (eg. "search through legal databases for 15 countries")

## Backend

The `DocumentSearchViewSet` must be configured to search all the various LII elasticsearch indexes. Currently, the list of indexes to search is determined by the language analyzers we support. That can be retained, but the list of base indexes should be configurable through `settings.py`.

The search index currently doesn't index anything that explicitly indicates what website the search result is on (apart from the index name). Rather than embedding that into the search index, I think we just support a more advanced resolver, which we need in any case. The resolver will be common to all LII websites.

### Resolver

The idea behind the resolver is that any peachjam site can determine which of the other (official) peachjam sites are responsible for a document. It can do that based on the country (and/or locality) of the FRBR URI.

For example:

1. I'm on lawlibrary.org.za and click on a link to a Zimbabwean act (eg. `/akn/zw/act/2009/1`)
2. the user is sent to `lawlibrary.org.za/akn/zw/act/2009/1`
3. lawlibrary first looks for the document in its database, but doesn't find it
7. it then makes a guess as to which LII website has it, based on the `zw` country code.
8. since lawlibrary is not authoritative for `zw`, it redirects the user to `zimlii.org/akn/zw/act/2009/1`
9. zimlii.org follows the same process. Notably, if it doesn't have the document it knows that it is authoritative for `zw` and shows a 404 page, rather than redirecting somewhere else.

In the general AfricanLII case, it will redirect most traffic to the authoritative site.

### Resolver configuration

For this to work, the resolver needs to know:

1. a list of the official partner websites (the authorities) and their country codes (perhaps also locality codes, if necessary)
2. which country code the local website is authoritative for (or, alternatively, which partner website it actually is)

I suggest making (1) a hardcoded list of authorities for the resolver, common to all LIIs. For (2) each resolver has a name which corresponds to the name provided for `APP_NAME` in each of the LIIs, so it's easy to determine the local LII.

actlikewill:
- Adds search across multiple indexes

### Demo

https://www.loom.com/share/25b8fb3b4fc94c039d82cc618dc921b6

closes #498

### DIFF
--- a/peachjam/views/documents.py
+++ b/peachjam/views/documents.py
@@ -1,3 +1,5 @@
+from cobalt import FrbrUri
+from django.conf import settings
 from django.http import Http404, HttpResponse
 from django.shortcuts import get_object_or_404, redirect
 from django.utils.decorators import method_decorator
@@ -9,6 +11,96 @@ from peachjam.models import CoreDocument
 from peachjam.registry import registry
 
 
+class RedirectResolver:
+
+    RESOLVER_MAPPINGS = {
+        "africanlii": {
+            "country_code": "aa",
+            "domain": "agp.africanlii.org",
+        },
+        "ghalii": {
+            "country_code": "gh",
+            "domain": "new.ghalii.org",
+        },
+        "lawlibrary": {
+            "country_code": "za",
+            "domain": "lawlibrary.org.za",
+        },
+        "lesotholii": {
+            "country_code": "ls",
+            "domain": "new.lesotholii.org",
+        },
+        "malawilii": {
+            "country_code": "mw",
+            "domain": "malawilii.org",
+        },
+        "namiblii": {
+            "country_code": "na",
+            "domain": "namiblii.org",
+        },
+        "senlii": {
+            "country_code": "sn",
+            "domain": "senlii.org",
+        },
+        "seylii": {
+            "country_code": "sc",
+            "domain": "new.seylii.org",
+        },
+        "sierralii": {
+            "country_code": "sl",
+            "domain": "new.sierralii.org",
+        },
+        "tanzlii": {
+            "country_code": "tz",
+            "domain": "new.tanzlii.org",
+        },
+        "tcilii": {
+            "country_code": "tc",
+            "domain": "new.tcilii.org",
+        },
+        "ulii": {
+            "country_code": "ug",
+            "domain": "new.ulii.org",
+        },
+        "zambialii": {
+            "country_code": "zm",
+            "domain": "new.zambialii.org",
+        },
+        "zanzibarlii": {
+            "place_code": "tz-znz",
+            "domain": "zanzibarlii.org",
+        },
+        "zimlii": {
+            "country_code": "zw",
+            "domain": "zimlii.org",
+        },
+    }
+
+    def __init__(self, app_name):
+        self.current_authority = self.RESOLVER_MAPPINGS[app_name.lower()]
+
+    def get_domain_for_frbr_uri(self, parsed_uri):
+        best_domain = self.get_best_domain(parsed_uri)
+        if best_domain != self.current_authority["domain"]:
+            return best_domain
+        return None
+
+    def get_best_domain(self, parsed_uri):
+        country_code = parsed_uri.country
+        place_code = parsed_uri.place
+
+        if country_code != place_code:
+            for key, mapping in self.RESOLVER_MAPPINGS.items():
+                if mapping.get("place_code") == place_code:
+                    return mapping.get("domain")
+
+        # if no domain matching with place code is found use country code
+        for key, mapping in self.RESOLVER_MAPPINGS.items():
+            if mapping.get("country_code") == country_code:
+                return mapping.get("domain")
+        return None
+
+
 class DocumentDetailViewResolver(View):
     """Resolver view that returns detail views for documents based on their doc_type."""
 
@@ -18,9 +110,21 @@ class DocumentDetailViewResolver(View):
             return redirect("document_detail", frbr_uri=kwargs["frbr_uri"][:-1])
 
         frbr_uri = add_slash(kwargs["frbr_uri"])
+
+        try:
+            # return 404 early if frbr_uri is invalid
+            parsed_frbr_uri = FrbrUri.parse(frbr_uri)
+        except ValueError:
+            raise Http404()
+
         obj, exact = CoreDocument.objects.best_for_frbr_uri(frbr_uri, get_language())
 
         if not obj:
+            resolver = RedirectResolver(settings.PEACHJAM["APP_NAME"])
+            domain = resolver.get_domain_for_frbr_uri(parsed_frbr_uri)
+            if domain:
+                url = f"https://{domain}{frbr_uri}"
+                return redirect(url)
             raise Http404()
 
         if not exact:
--- a/africanlii/settings.py
+++ b/africanlii/settings.py
@@ -7,3 +7,20 @@ ROOT_URLCONF = "africanlii.urls"
 JAZZMIN_SETTINGS["site_title"] = "Africanlii"  # noqa
 JAZZMIN_SETTINGS["site_header"] = "Africanlii"  # noqa
 JAZZMIN_SETTINGS["site_brand"] = "Agp.africanlli.org"  # noqa
+
+EXTRA_SEARCH_INDEXES = [
+    "ghalii",
+    "lawlibrary",
+    "lesotholii",
+    "malawilii",
+    "namiblii",
+    "senlii",
+    "seylii",
+    "sierralii",
+    "tanzlii",
+    "tcilii",
+    "ulii",
+    "zambialii",
+    "zanzibarlii",
+    "zimlii",
+]
--- a/peachjam/settings.py
+++ b/peachjam/settings.py
@@ -257,6 +257,8 @@ ELASTICSEARCH_DSL_SIGNAL_PROCESSOR = (
     "peachjam_search.tasks.BackgroundTaskSearchProcessor"
 )
 
+EXTRA_SEARCH_INDEXES = []
+
 REST_FRAMEWORK = {
     "DEFAULT_AUTHENTICATION_CLASSES": [
         "rest_framework.authentication.SessionAuthentication",
--- a/peachjam/tests/test_resolver.py
+++ b/peachjam/tests/test_resolver.py
@@ -1,0 +1,73 @@
+from cobalt import FrbrUri
+from django.test import TestCase
+
+from peachjam.views.documents import RedirectResolver
+
+urls = [
+    "/akn/zm/judgment/zmsc/2021/7/eng@2021-01-19",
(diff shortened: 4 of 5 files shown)


peachjam/views/documents.py
```diff
@@ -9,6 +11,55 @@
...
+            locality = parsed_uri.locality
+
+            # special case for tanzlii and zanzibarlii. All others are authoritative for their countries
+            if (
+                country_code == "tz"
+                and locality == "znz"
+                and self.current_authority["domain"]
+                != self.RESOLVER_MAPPINGS["zanzibarlii"]["domain"]
+            ):
+                return f"http://{self.RESOLVER_MAPPINGS['zanzibarlii']['domain']}{frbr_uri}"
```
longhotsummer, line 52:
```suggestion
                return f"https://{self.RESOLVER_MAPPINGS['zanzibarlii']['domain']}{frbr_uri}"
```

peachjam/views/documents.py
```diff
@@ -9,6 +11,55 @@
...
+
+            if country_code != self.current_authority["code"]:
+                for k, v in self.RESOLVER_MAPPINGS.items():
+                    if country_code == v["code"]:
+                        return f"http://{v['domain']}{frbr_uri}"
+            return None
+        except ValueError:
```
longhotsummer, line 59: It's better to fail loudly than silently. I think this exception block is too broad -- if the code assumes something is in the mappings settings, it's better to fail when it's not than to silently do nothing.

peachjam/views/documents.py
longhotsummer, line 52: You're building the final URL in two places, which means you're duplicating code.

Rather, split this into two steps:

1. identify which authority to use (if any) -- handle any authority-lookup edge cases here
2. based on that authority, generate a URL

peachjam/views/documents.py
```diff
@@ -9,6 +11,55 @@
 from peachjam.registry import registry
 
 
+class RedirectResolver:
+
+    RESOLVER_MAPPINGS = {
+        "africanlii": {"code": "aa", "domain": "agp.africanlii.org"},
```
longhotsummer, line 17: `code` is ambiguous. In FRBR URI terms, we have three:

1. country code (eg. `tz`)
2. locality code (eg. `znz`) which must be combined with a country code to be unique
3. place code -- the combination of the two (eg. `tz-znz`)

We need to be able to juggle those. For example, LawLibrary is authoritative for all South African content, so it covers `za` and all localities (`za-*`). However, TanzLII is authoritative for `tz` and `tz-*` except `tz-znz`.

Matching on a full place code  should be considered a "tighter" match than matching on just the country code. So here's a suggestion that will make it easier to handler the edge cases like ZanzibarLII:

1. change `code` to `country_code` for all of these settings, except ZanzibarLII, where change it to `place_code`
2. first check for an exact match on `place_code`; if found, use that authority
3. then check for a match on `country_code`; if found, use that authority (this will be the common case)
4. nothing matches, so give up

peachjam/views/documents.py
longhotsummer: I'd say splitting this method into two separate methods would make that easier.

peachjam/views/documents.py
```diff
@@ -9,6 +11,55 @@
...
+    }
+
+    def __init__(self):
+        self.current_authority = self.RESOLVER_MAPPINGS[
+            settings.PEACHJAM["APP_NAME"].lower()
+        ]
+
+    def get_url_for_frbr_uri(self, frbr_uri):
+        try:
+            parsed_uri = FrbrUri.parse(frbr_uri)
```
longhotsummer, line 41: I think the caller should do the parsing and provide an FrbrUri object. This object doesn't have enough context to know what to do if the FRBR URI isn't valid, whereas the caller has a better idea (eg. raise a 404 or throw an exception).

longhotsummer (changes requested):
Nice! I've left some feedback that will help remove TanzLII as a hard-coded edge case.

peachjam/views/documents.py
```diff
@@ -9,6 +11,48 @@
...
+        "tcilii": {"country_code": "tc", "domain": "new.tcilii.org"},
+        "ulii": {"country_code": "ug", "domain": "new.ulii.org"},
+        "zambialii": {"country_code": "zm", "domain": "new.zambialii.org"},
+        "zanzibarlii": {"place_code": "tz-znz", "domain": "new.zanzibarlii.org"},
+        "zimlii": {"country_code": "zw", "domain": "zimlii.org"},
+    }
+
+    def __init__(self):
+        mapping_code = self.RESOLVER_MAPPINGS[settings.PEACHJAM["APP_NAME"].lower()]
+        self.current_authority = mapping_code.get("place_code") or mapping_code.get(
```
longhotsummer, line 36: I think it'll be simpler if the current authority is either the key in the dict, or the object from the dict. That way you can compare an authority with the current authority directly, without having to re-compare place codes.

So the code comparison is done ONLY when looking for the authority that matches an FRBR URI. Once you have that authority, you compare it with the current one.

Otherwise you're juggling codes in three places, instead of one.

peachjam/views/documents.py
```diff
@@ -9,6 +11,48 @@
...
+        "malawilii": {"country_code": "mw", "domain": "malawilii.org"},
+        "namiblii": {"country_code": "na", "domain": "namiblii.org"},
+        "senlii": {"country_code": "sn", "domain": "senlii.org"},
+        "seylii": {"country_code": "sc", "domain": "new.seylii.org"},
+        "sierralii": {"country_code": "sl", "domain": "new.sierralii.org"},
+        "tanzlii": {"country_code": "tz", "domain": "new.tanzlii.org"},
+        "tcilii": {"country_code": "tc", "domain": "new.tcilii.org"},
+        "ulii": {"country_code": "ug", "domain": "new.ulii.org"},
+        "zambialii": {"country_code": "zm", "domain": "new.zambialii.org"},
+        "zanzibarlii": {"place_code": "tz-znz", "domain": "new.zanzibarlii.org"},
```
longhotsummer, line 30:
```suggestion
        "zanzibarlii": {"place_code": "tz-znz", "domain": "zanzibarlii.org"},
```

peachjam/views/documents.py
```diff
@@ -9,6 +11,48 @@
...
+
+    def get_url_for_frbr_uri(self, parsed_uri):
+        country_code = parsed_uri.country
+        place_code = parsed_uri.place
+        domain = None
+        if (
+            self.current_authority != place_code
+            or self.current_authority != country_code
+        ):
+            for keys, values in self.RESOLVER_MAPPINGS.items():
```
longhotsummer, line 48: I think you need a test for this method, because this logic is buggy :)

actlikewill: @longhotsummer This logic is kinda tricky to get right. Mostly, because of the special case of zanzibarlii and tanzlii. I want zanzibarlii to be authoritative for `/akn/tz-znz` documents, while tanzlii should be authoritative for `/akn/tz-*` documents. The code checks if the request has a locality code and uses that to check for an authoritative domain. If there is none, it falls back to the country code. What happens is when on you are on zanzibarlii and you request for `akn/tz/doc/`, (without the locality) it still thinks that zanzibarlii is authoritative for that document whereas I want it to to go to tanzlii. i want this to work right and still preserve the same behavior for `akn/aa-*/`  and `akn/za-*`. The logic should still be extendable if we were, say,  to deploy a site for `aa-eac` or `za-mp`.

actlikewill: Let me add some failing tests

peachjam/views/documents.py
actlikewill: I've split the method up into two, one for getting best match and one for comparing with current authority

peachjam/views/documents.py
actlikewill: @longhotsummer  as we change the domains we will need to update the resolver mappings

peachjam/views/documents.py
actlikewill: tests added

peachjam/views/documents.py
actlikewill: added

peachjam/views/documents.py
```diff
@@ -9,6 +11,98 @@
...
+
+    def get_best_authority(self, parsed_uri):
+        country_code = parsed_uri.country
+        place_code = parsed_uri.place
+
+        domain = None
+        if country_code != place_code:
+            for key, mapping in self.RESOLVER_MAPPINGS.items():
+                if mapping.get("place_code") == place_code:
+                    domain = mapping.get("domain")
```
longhotsummer, line 96: I think you can simply `return mapping.get("domain")` here (or `return mapping`), no need to keep looping.

Ditto for line 102 below

actlikewill: right, changed

peachjam/views/documents.py
```diff
@@ -9,6 +11,96 @@
...
+        "zimlii": {
+            "country_code": "zw",
+            "domain": "zimlii.org",
+        },
+    }
+
+    def __init__(self, app_name):
+        self.current_authority = self.RESOLVER_MAPPINGS[app_name.lower()]
+
+    def get_url_for_frbr_uri(self, parsed_uri):
```
longhotsummer, line 82: Minor: this method name implies it will return a URL but it actually returns a domain, and `get_best_authority` implies that method will return an authority but actually returns a domain.

Some variable and method renaming would make things more consistent.

longhotsummer (approved):
super. This can be merged. There are still the UX changes to make.

peachjam/views/documents.py
actlikewill: lol, always happy to renaime variables

actlikewill: done

PR #1110, opened 2023-04-19
Force default language for ZanzibarLII and TanzLII

### LINKED ISSUE #1086: As a reader visiting TanzLII for the first time, I want the site to default to Swahili
About 10% of users have Swahili in their browser configuration. This isn't enough for us to provide a good experience on TanzLII for Swahili speakers.

Instead, the site must default to Swahili. Non-Swahili speakers can then change to English. I'm not sure how we force a language on Django unless the user has a language cookie set. Do we need tweaked middleware?

Additionally, we need to make the option to change languages clearer. I think we can improve the dropdown label, as well as put something bigger on the homepage.

How do other international sites handle this? What's the norm?

nickmwangemi:
This PR adds a custom middleware that forces the default site language to be used, unless another one is set via sessions or cookies.

This has been applied for ZanzibarLII and TanzLII.

Resolves #1086

### DIFF
--- a/peachjam/middleware.py
+++ b/peachjam/middleware.py
@@ -1,4 +1,5 @@
 from django.shortcuts import redirect
+from django.utils.deprecation import MiddlewareMixin
 
 
 class RedirectWWWMiddleware:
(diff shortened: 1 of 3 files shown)


longhotsummer (approved):
very simple, nice!

PR #1125, opened 2023-04-24
Faster pdf loading

### LINKED ISSUE #1067: Improve the load time of PDFs on PeachJam
Loading is very slow compared to LIIweb v2 and loading pauses if the user navigates away from the window. In effect, users sit and wait for a document to load, which is not a good experience

longhotsummer:
* simplify PDF loading code
* rather than waiting for all pages to render before adding them to the page, we hide the progress bar once the file is downloaded and display the pages immediately

Fixes #1067


actlikewill (approved):
Works well

PR #1127, opened 2023-04-24
Search ui for global search

### LINKED ISSUE #498: AfricanLII: search across multiple LIIs

actlikewill:
- Updates search ui

image

image

closes #498


peachjam_search/templates/peachjam_search/search.html
```diff
@@ -6,6 +6,11 @@
 {% block page-content %}
   <div>
     <div class="container pt-lg-5 pt-2 pb-5">
+      <p class="lead text-center">
+        {% blocktrans trimmed %}
+        Search regional and national African legal information.
```
longhotsummer, line 11: This will need to be done in the africanlii app so that other places don't pick it up.

actlikewill: @longhotsummer This view is served by the peachjam_search app not the main peachjam app. Whats the best way to override it?

peachjam/js/locale/sw/translation.json
```diff
@@ -50,6 +50,7 @@
   "Search documents by flynote": "Tafuta nyaraka kwa flynote",
   "Search documents by headnote holding": "Tafuta hati kwa kutumia muhtasari",
   "Search documents by judges": "Tafuta nyaraka kwa kutumia majaji",
+  "Search documents by jurisdiction": "Tafuta nyaraka kwa kutumia mamlaka za kisheria",
   "Search documents by title": "Tafuta hati kwa kutumia kichwa cha habari",
```
actlikewill, line 54: @longhotsummer I added this translation for swahili. The other locales do not have the translation for this string. i.e french and portugese

peachjam_search/views.py
```diff
@@ -174,7 +175,10 @@ class SearchView(TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context["labels"] = {"author": Author.model_label}
+        context["labels"] = {
+            "author": Author.model_label,
+            "searchPlaceholder": _(f"Search {settings.PEACHJAM['APP_NAME']}"),
```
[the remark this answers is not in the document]
actlikewill: Cool. Changed.

peachjam_search/templates/peachjam_search/search.html
longhotsummer: Create a file called `africanlii/templates/peachjam_search/search.html` and it will override this one, like normal. You'll need to setup a block in the base template to override it easily in the new template.

peachjam/js/components/FindDocuments/AdvancedSearch.vue
```diff
@@ -94,6 +94,29 @@
               >
             </div>
           </div>
+          <div class="col-12 mb-3">
```
longhotsummer, line 97: There's a pending change to substantially alter how advanced search works. I suggest you don't add this now, and it can be added later, once that PR is merged.

peachjam_search/templates/peachjam_search/search.html
actlikewill: :sweat_smile:  I forgot that you can simply subclass a template, I was thinking I had to pull the search app into africanlii

peachjam/js/components/FindDocuments/AdvancedSearch.vue
actlikewill: @longhotsummer I've removed what I added in advanced search.

longhotsummer (approved):

PR #1128, opened 2023-04-25
Add ordering mechanism for courts

nickmwangemi:
This PR adds a mechanism for ordering courts and court classes in a hierarchical manner.

Before:
image

After:
image


peachjam/views/judgment.py
```diff
@@ -15,7 +15,7 @@ def get_context_data(self, **kwargs):
 
         court_classes = (
             CourtClass.objects.select_related("courts")
-            .order_by("name", "courts__name")
+            .order_by("courts__order", "name", "courts__name")
```
longhotsummer, line 18: I don't think this makes sense. Once the court classes also have an `order` field added, then this can be adjusted to take that into account.

nickmwangemi:
> Great. I'm assuming you have tested multiple court classes and that it all works as expected.

Yes. If a number of court classes have the same ordering, it defaults to the name, in alphabetical order.

PR #1133, opened 2023-04-26
Use SVG flag icons rather than emoji

longhotsummer:
This is because Windows doesn't have a built-in font for emoji flags :(

See https://nolanlawson.com/2022/04/08/the-struggle-of-using-native-emoji-on-the-web/

image


actlikewill:
@longhotsummer Does the css need to be explicitly imported somewhere?

actlikewill (approved):

PR #1136, opened 2023-04-30
Adds 1x1 flags

actlikewill:
- Adds the missing flags  folder
- The deploys are failing because the flag.min.css file is references these files. 
- I tried editing the file to remove the references they are too many. It is easier just adding the folder.


longhotsummer:
Thanks

PR #1140, opened 2023-05-02
Fix ranker with renamed field

actlikewill:
- Update the ranker with the new ranking field.


actlikewill:
@longhotsummer I had not updated the ranker with the previous PR,

peachjam/migrations/0070_alter_work_ranking.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("peachjam", "0069_rename_rank_work_ranking"),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
longhotsummer, line 13: What is different here to the previous migration?

longhotsummer (approved):

peachjam/migrations/0070_alter_work_ranking.py
longhotsummer: oh the label

PR #1141, opened 2023-05-03
Use gthread not gevent

longhotsummer:
* use gthread not gevent
* reduce workers to 2 and use 8 threads in each worker
* this will mean that long-running requests (eg. PDF conversion) don't block other requests

See:

* https://dev.to/lsena/gunicorn-worker-types-how-to-choose-the-right-one-4n2c 
* https://luis-sena.medium.com/gunicorn-vs-python-gil-221e673d692

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ENV PYTHONUNBUFFERED=1
 RUN apt-get update && apt-get install -y libreoffice poppler-utils
 
 # Production-only dependencies
-RUN pip install psycopg2==2.9.3 gevent==21.12.0 gunicorn==20.1.0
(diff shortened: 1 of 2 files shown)


actlikewill:
@longhotsummer I'm curious, how were you able to tell that requests were being blocked?

actlikewill (approved):

longhotsummer:
In Indigo, we could see that long-running requests just blocked all other requests. When the long running one completed, suddenly the others flooded through. In peachjam sites we haven't directly seen it, but the risk is there.

It's also based on my understanding of how the worker processes work.

Basically, gthread is more widely used and is simpler.

PR #1152, opened 2023-05-04
Rank feature search query

### LINKED ISSUE #1073: Chore: use pagerank to boost highly cited documents in search
The idea is that [PageRank](https://en.wikipedia.org/wiki/PageRank) will help boost both commonly-cited documents and "important" documents higher in search. In particular, searching for "constitution" should provide the principal act (which is highly cited) rather than the commencements or amendment acts.

## Proposal

[Neo4J](https://neo4j.com/) is a widely used graph database that supports PageRank (and ArticleRank, a version better suited to our use case). The details then become: how do we get data into the graph database, calculate a rank for works, and get the rank out and into ElasticSearch?

### Concepts

First, some concepts:

PageRank works on a graph. A (computer science) graph is made up of nodes (eg. documents) and edges (the relationships between the nodes) -- [more detail in the neo4j docs](https://neo4j.com/docs/getting-started/current/get-started-with-neo4j/graph-database/).

Conceptually, we want to rank all the Works in a LII database based on citation and other relationships between them. Note that our citations and relationships are between works and not documents. All the historical and language versions of a work have the same ranking.

The relationships between our works are citations for now. Later, we can explore bringing in relationships such as amendments and those defined manually by admins. The citation relationship is directional (like a link in a webpage): a citing work cites a target work. The citing work lends weight to the target work.

PageRank produces a ranking of each work. This is a floating point number normalised to between 0 and 1. A higher number means, roughly, that the work is more important.

[ArticleRank](https://neo4j.com/docs/graph-data-science/current/algorithms/article-rank/) is designed for citation networks and is very similar to PageRank, except that it doesn't bias against works that have a lot of outgoing citations. 

### When to rank

Either periodically (eg. daily) or when works or relationships are updated, we need to push information into the neo4j database, run the pagerank algorithm to generate rankings. We probably want to do this when the relevant objects have changed, particularly for large databases (100k+ gazettes and judgments, for example), rather than unnecessarily. 

### Calculating rankings

The process of ranking is as follows:

1. clear out the existing graph in neo4j
2. create all the works in neo4j
3. add all the relationships (citation links) to neo4j, based on our existing ExtractedCitation model
4. run the ArticleRank algorithm on the neo4j graph, which updates the `rank` attribute of each node
5. update the `rank` attribute on the `Work` model in the LII database (a new field)
6. push the updated rankings into ElasticSearch, using the work's ranking for each of its documents. We probably want to find a way to update just this one field, and do so in bulk, rather than re-indexing the entire dataset (which is slow).

[neomodel](https://neomodel.readthedocs.io/en/latest/) is a high-level library for declaring and interacting with Neo4j objects. 

### Using rankings

ElasticSearch has a particular type of field for boosting search results, called [`rank_feature`](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-rank-feature-query.html). When a search is executed, we tell ElasticSearch to boost results based on the `rank_feature`.

### Frontend

There are no frontend changes required.

### Testing and deployment

We need a production-sized database to test rankings effectively, because we need real-world citation network data. However, we don't yet know how much weight to give to those rankings during searching.

We first need to calculate rankings and push them into ElasticSearch (but not yet use them). Then, we can use our development environment to search the production ElasticSearch and work out the best weights for the rankings.

Adding the new field will require updating ElasticSearch field mappings, and therefore a re-index of all content (this is an hour or two for our biggest websites). 

Rather than updates triggering ranking, we can start with a management command to update rankings manually. Once we have the process refined, we can setup signals to trigger updates.

actlikewill:
- adds rank feature to search query

closes #1073


longhotsummer:
Suggested next steps:

1. when indexing into ES, ensure all documents have at least a very small ranking value (eg. 0.0000001)
2. put the ranking query into the MUST clauses
3. add the rank boost to the site settings

peachjam_search/views.py
```diff
@@ -55,6 +55,17 @@ def construct_search(self, request, view, search_backend):
         ]
 
 
+class RankFeatureBackend(BaseSearchQueryBackend):
+    @classmethod
+    def construct_search(cls, request, view, search_backend):
+        rank = Q(
```
longhotsummer, line 61: Can you simply not return anything if the setting's boost value isn't set? Because even with zero, this will break on indexes that don't yet have the ranking feature.

peachjam_search/views.py
```diff
@@ -134,14 +146,20 @@ class SearchFilterBackend(CompoundSearchFilterBackend):
...
-    should_backends = [
-        # Search each field individually using SimpleQueryString which allows quotes, +foo, -bar etc.
-        CrossFieldSimpleQueryStringBackend,
-        # Customised search on PDF page content
-        NestedPageQueryBackend,
-    ]
+        self.must_backends = [MultiFieldSearchQueryBackend()]
+
+        if pj_settings().pagerank_boost_value:
+            self.must_backends += [RankFeatureBackend()]
```
longhotsummer, line 155: Ah, I see. I think hard-code in the backends, and then the backend can return an empty list instead? Less fiddly than doing this here.

longhotsummer: That way, the backend is the only thing that cares about its config. Otherwise both this and the backend must care.

peachjam/graph/ranker.py
```diff
@@ -73,11 +80,74 @@ def calculate_ranks(self):
...
         for neo_work in NeoWork.nodes.all():
-            work = works[neo_work.frbr_uri]
-            work.ranking = neo_work.ranking
-            work.save(update_fields=["ranking"])
+            w = works[neo_work.frbr_uri]
+            # only update those that have changed
+            if w["work"].ranking != neo_work.ranking:
+                w["work"].ranking = neo_work.ranking
+                w["work"].save(update_fields=["ranking"])
+                w["work"]["updated"] = True
```
longhotsummer, line 91: This is ... odd. Rather than using a dict, maybe just use a list?

```python
updated = []
...
w = works[neo_work.frbr_uri]
if neo_work.ranking != w.ranking:
  w.ranking = neo_work.ranking
  w.save(updated_fields=["ranking"])
  updated.add(w)

self.update_index([w.frbr_uri for w in updated])
```

peachjam_search/views.py
actlikewill: changed

peachjam_search/views.py
actlikewill: changed :heavy_check_mark:

peachjam/graph/ranker.py
actlikewill: changed :heavy_check_mark:

longhotsummer (approved):
Awesome, let's get it out and re-rank all the lawlibrary docs.

PR #1172, opened 2023-05-10
Add mailchimp config for TCILII newsletter

nickmwangemi:
Adds the mailchimp form action code necessary for newsletter to work


nickmwangemi:
@longhotsummer Were we aiming to have the newsletter form different from what we have by any chance?

longhotsummer:
> @longhotsummer Were we aiming to have the newsletter form different from what we have by any chance?

No I don't think so. They should all follow the same pattern.

longhotsummer (approved):
Looks good, assuming you have tried it :)

PR #1184, opened 2023-05-11
Imports and exports

longhotsummer:
Ensure that an exported xlsx can be re-imported.

* support alternative names import and export
* ValueError must be raised rather than ValidationError
* Not all fields are required
* Handle empty values for optional fields

### DIFF
--- a/peachjam/resources.py
+++ b/peachjam/resources.py
@@ -185,6 +185,28 @@ class TaxonomiesWidget(CharWidget):
         return value
 
 
+class IncomingManyToOneField(fields.Field):
+    """Handles many-to-one relationships on the incoming side."""
+
+    def save(self, obj, data, is_m2m=False, **kwargs):
+        if not self.readonly:
+            attrs = self.attribute.split("__")
+            for attr in attrs[:-1]:
+                obj = getattr(obj, attr, None)
+            cleaned = self.clean(data, **kwargs)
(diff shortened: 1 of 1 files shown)


peachjam/resources.py
```diff
@@ -356,21 +368,12 @@ class GenericDocumentResource(BaseDocumentResource):
         attribute="nature",
         widget=DocumentNatureWidget(DocumentNature, field="code"),
     )
-    work_frbr_uri = fields.Field(
-        column_name="work_frbr_uri",
-        attribute="work_fbr_uri",
-        widget=CharRequiredWidget(field="work_frbr_uri"),
-    )
 
     required_fields = (
```
longhotsummer, line 372: I don't think this is actually doing anything.

actlikewill:
@longhotsummer The case names, judges and some other fields are being lost with reimport because we handle them in a special way. I can take a stab at adding the export methods that will format them correctly. For case names, we will have to change the way we handle them in import for it to work correctly.

longhotsummer:
> @longhotsummer The case names, judges and some other fields are being lost with reimport because we handle them in a special way. I can take a stab at adding the export methods that will format them correctly. For case names, we will have to change the way we handle them in import for it to work correctly.

Ok, please go for it. Create a new branch off of this one.

actlikewill (approved):

PR #1187, opened 2023-05-12
Associate generic documents and legal instruments with multiple authors

### LINKED ISSUE #1137: As an editor I want to be able to associate multiple authors with a document
This will unblock briefing papers #1035

nickmwangemi:
Admin panel
image

Document metadata display
image

## Questions guiding implementation
1. What document type(s) does the new field apply to?
    - It applies for the generic document type
2. Is the new field related to any other fields, such as court or jurisdiction?
    - No.
3. What countries does the new field apply to?
    - Applies to all jurisdictions.
4. Is the field required, or optional?
    - The field is optional.
5. Must the field be included in the document metadata display?
    - Yes, it'll be included with an `Authors` label.
6. Must the field be included as a search facet?
    - No.
7. Must the field be included as a document listing facet? On which listing pages?
    - No.
8. Does it impact the MNC for judgments?
   - No.
9. Must the field be included in the import/export functionality?
   - Yes

## Implementation aspects:
- [x] Added to modelling?
- [x] Added to admin interface?
- [x] Added to document detail page?

Resolves #1137


peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -12,7 +12,20 @@
   <meta property="og:type" content="book" />
   <meta property="book:release_date"
         content="{{ document.date|date:'Y-m-d' }}"/>
-  {% if document.author %}<meta property="book:author" content="{{ document.author }}" />{% endif %}
```
longhotsummer, line 15: I suggest we just drop this field, we're not actually doing the correct thing any way: https://ogp.me/#type_profile

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -174,6 +187,21 @@ <h1>{{ document.title }}</h1>
                         {% endwith %}
                       {% endblock %}
                       {% block document-metadata-content-author %}
+                        {% if document.doc_type == "generic_document" %}
+                          {% with document.authors.all as authors %}
+                            {% if authors %}
+                              <dt>
+                                {% trans 'Authors' %}
```
longhotsummer, line 194: You need to fold in the logic from line 179 below.

longhotsummer (commented):
Looking good. @actlikewill pls sanity check the import/export functionality.

@nickmwangemi does the peachjam_search.documents.SearchableDocument need to be updated?

peachjam_search/documents.py
```diff
@@ -53,7 +53,7 @@ class SearchableDocument(Document):
     order_outcome = fields.KeywordField(attr="order_outcome.name")
 
     # GenericDocument, LegalInstrument
-    author = fields.KeywordField(attr="author.name")
+    authors = fields.KeywordField(attr="authors.name")
```
longhotsummer, line 56: This won't work because `document.authors.name` doesn't work. look instead at how we handle it for alternative names, see `prepare_alternative_names`

peachjam/migrations/0075_backfill_authors.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 3.2.18 on 2023-05-16 11:30
+
+from django.db import migrations
+
+
+def backfill_authors(apps, schema_editor):
+    CoreDocument = apps.get_model("peachjam", "CoreDocument")
+    for doc in CoreDocument.objects.all().iterator(chunk_size=100):
```
longhotsummer, line 8:
```suggestion
    for doc in CoreDocument.objects.only('author').all().iterator(chunk_size=100):
```

otherwise the entire content is loaded which is slow.

peachjam_search/documents.py
```diff
@@ -53,7 +53,7 @@ class SearchableDocument(Document):
     order_outcome = fields.KeywordField(attr="order_outcome.name")
 
     # GenericDocument, LegalInstrument
-    author = fields.KeywordField(attr="author.name")
+    authors = fields.TextField()
```
longhotsummer, line 56:
```suggestion
    authors = fields.KeywordField()
```

it must remain a keyword field so we can calculate aggregates. It can safely have multiple values.

peachjam/migrations/0075_backfill_authors.py
longhotsummer: You will need to explicitly do this for `GenericDocument` and `LegalInstrument`, since 'author' is not a field on `CoreDocument`

peachjam/migrations/0075_backfill_authors.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 3.2.18 on 2023-05-16 11:30
+
+from django.db import migrations
+
+
+def backfill_authors(apps, schema_editor):
+    CoreDocument = apps.get_model("peachjam", "CoreDocument")
```
longhotsummer, line 7: You need to do this separately for `LegalInstrument` and `GenericDocument` because `CoreDocument` doesn't have the field `author`.

peachjam/migrations/0075_backfill_authors.py
```diff
@@ -0,0 +1,20 @@
...
+    for doc in CoreDocument.objects.only('author').all().iterator(chunk_size=100):
```
longhotsummer, line 8: Apply a filter `.exclude(author=None)` otherwise you're going update 100,000+ gazettes unnecessarily.

peachjam/migrations/0075_backfill_authors.py
```diff
@@ -0,0 +1,20 @@
...
+        if doc.author not in doc.authors:
```
longhotsummer, line 9: in this forward migration, you know that `doc.authors` is empty. Can you just use `doc.authors.set([author])`?

peachjam/forms.py
```diff
@@ -107,7 +107,7 @@ def filter_queryset(self, queryset, exclude=None):
         alphabet = self.cleaned_data.get("alphabet")
         authors = self.params.getlist("authors")
         courts = self.params.getlist("courts")
-        doc_type = self.params.getlist("doc_type")
+        doc_types = self.params.getlist("docTypes")
```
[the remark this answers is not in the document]
longhotsummer: The vue.js code that sends the param uses `doc_type`, so we must use that.

longhotsummer (changes requested):
The vue.js for handling search results must also change to use `authors`. https://github.com/laws-africa/peachjam/blob/4ac5e318bfc55e930e073a16fc3f55dbce07f315/peachjam/js/components/FindDocuments/SearchResult.vue#L26

longhotsummer:
> The vue.js for handling search results must also change to use `authors`.
> 
> https://github.com/laws-africa/peachjam/blob/4ac5e318bfc55e930e073a16fc3f55dbce07f315/peachjam/js/components/FindDocuments/SearchResult.vue#L26

For this, in that file create a Vue method something like this:

```js
methods: {
  ...
  authors () {
    if (item.authors) {
      return Array.isArray(item.authors) ? ', '.join(item.authors) : item.authors;
    }
    return '';
  }
}
```

and then replace the `{{ item.author }}` in the span in the HTML with `authors`.

peachjam/js/components/FindDocuments/SearchResult.vue
```diff
@@ -24,9 +24,9 @@
         class="me-3"
       >{{ item.court }}</span>
       <span
-        v-if="item.author"
+        v-if="item.authors"
         class="me-3"
-      >{{ item.author }}</span>
+      >{{ item.authors }}</span>
```
longhotsummer, line 29:
```suggestion
      >{{ authors(item) }}</span>
```

longhotsummer (approved):

PR #1191, opened 2023-05-13
Ensure super().get_base_queryset is used

longhotsummer:
Otherwise we don't get the document language filtering behaviour

Before:

image

### DIFF
--- a/peachjam/views/authors.py
+++ b/peachjam/views/authors.py
@@ -1,7 +1,7 @@
 from django.db.models import Q
 from django.shortcuts import get_object_or_404
 
-from peachjam.models import Author, CoreDocument
+from peachjam.models import Author
 from peachjam.views.generic_views import FilteredDocumentListView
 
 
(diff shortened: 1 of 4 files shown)


peachjam/views/authors.py
```diff
@@ -10,9 +10,13 @@ class AuthorDetailView(FilteredDocumentListView):
     navbar_link = "author_detail"
 
     def get_base_queryset(self):
-        return CoreDocument.objects.prefetch_related("nature", "work").filter(
```
longhotsummer, line 13: This was the major issue that prompted this PR. The rest is just cleaning up.

peachjam/views/generic_views.py
```diff
@@ -27,6 +27,7 @@ class DocumentListView(ListView):
     context_object_name = "documents"
     paginate_by = 50
     model = CoreDocument
+    queryset = CoreDocument.objects.prefetch_related("nature", "work")
```
longhotsummer, line 30: Nature is now common to all document types, so we can move this here.

PR #1205, opened 2023-05-16
fix redirects

longhotsummer:
* old seylii judgments use `court-appeal` not `court-of-appeal`: eg https://old2.seylii.org/sc/judgment/court-appeal/2023/8
* remove `new.` for sites that have been migrated


nickmwangemi (dismissed):

nickmwangemi:
The changes broke some tests

PR #1229, opened 2023-05-19
Configurable social login

### LINKED ISSUE #1227: Allow login for Gmail emails
We need to allow login for gmail emails

- To the Settings model we can add a field called allowed_login_domains
- This can be a string and it will accept a comma separated list of domains that will be allowed to login using Google login.
- Default value can be "laws.africa,africanlii.org"
- This value is what will be used in the social account adapter as the `ALLOWED_DOMAINS`
- To disable account sign up using Social auth, we can override the `is_open_for_signup` method and return False.

actlikewill:
- allows social login domains to be configurable from admin

closes #1227


actlikewill:
NB: deploying this might cause the deploy to fail but maybe not the setting value is not is a template

peachjam/auth.py
```diff
@@ -6,16 +6,22 @@
...
+
 
 class SocialAccountAdapter(DefaultSocialAccountAdapter):
     """Adapter that permits logins only from specific domains."""
 
-    ALLOWED_DOMAINS = ["laws.africa", "africanlii.org"]
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.allowed_domains = [
+            s.strip() for s in pj_settings().allowed_login_domains.split(",")
```
longhotsummer, line 18: I think this is easier to manage and read if it's space-separated

peachjam/auth.py
```diff
@@ -6,16 +6,22 @@
...
+        ]
 
     def pre_social_login(self, request, sociallogin):
         domain = sociallogin.email_addresses[0].email.split("@", 1)[1]
 
-        if domain not in self.ALLOWED_DOMAINS:
+        if domain not in self.allowed_domains:
```
longhotsummer, line 24: we should skip this check if no domains are configured

longhotsummer:
> NB: deploying this might cause the deploy to fail but maybe not the setting value is not is a template

I think it will. Good time to move the migration into the pre_deploy step in app.json

peachjam/auth.py
actlikewill: I have set some default values `laws.africa` and `africanlii.org` so this is always present

longhotsummer: What if we want to allow logins for any domain?

peachjam/auth.py
```diff
@@ -6,16 +6,22 @@
...
+
 
 class SocialAccountAdapter(DefaultSocialAccountAdapter):
     """Adapter that permits logins only from specific domains."""
 
-    ALLOWED_DOMAINS = ["laws.africa", "africanlii.org"]
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.allowed_domains = [
+            s.strip() for s in pj_settings().allowed_login_domains.split(" ")
```
longhotsummer, line 18: If you use `.split()` you don't need to worry about weird whitespace or strip the results, and an empty string works as expected.

```python
>>> "  foo bar   baz boom  ".split(" ")
['', '', 'foo', 'bar', '', '', 'baz', 'boom', '', '']
>>> "  foo bar   baz boom  ".split()
['foo', 'bar', 'baz', 'boom']
>>> "".split()
[]
>>> "".split(" ")
['']
>>> " ".split(" ")
['', '']
>>> " ".split()
[]
```

actlikewill: awesome. thanks

longhotsummer (dismissed):
great

longhotsummer (approved):

longhotsummer (approved):

PR #1230, opened 2023-05-19
Nesting taxonomies in admin

actlikewill:
- shows nesting of taxonomies in admin

image

### DIFF
--- a/peachjam/admin.py
+++ b/peachjam/admin.py
@@ -152,7 +152,21 @@ class SourceFileInline(BaseAttachmentFileInline):
     readonly_fields = (*BaseAttachmentFileInline.readonly_fields, "source_url")
 
 
+class TopicChoiceField(forms.ModelChoiceField):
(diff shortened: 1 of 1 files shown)


peachjam/admin.py
```diff
@@ -152,7 +152,21 @@ class SourceFileInline(BaseAttachmentFileInline):
     readonly_fields = (*BaseAttachmentFileInline.readonly_fields, "source_url")
 
 
+class TopicChoiceField(forms.ModelChoiceField):
+    def label_from_instance(self, obj):
+        return f"{'-'*(obj.depth-1)}{obj.name}"
```
longhotsummer, line 157:
```suggestion
        return f"{'-'*(obj.depth-1)} {obj.name}"
```

I think it's a touch easier to read with a space between the dashes and the name.

longhotsummer (approved):
nice!

nickmwangemi (approved):

PR #1233, opened 2023-05-20
Use igraph rather than neo4j

longhotsummer:
* use igraph (which has simple dependencies, unlike networkx which requires numpy and scipy) for calculating pagerank
* eliminate need for external graph database
* clean up top-level dependencies in pyproject.toml

We only need to track our top-level dependencies. Transitive dependencies don't need to be versioned at the top level. This makes it easier for us to maintain our dependencies and understand version conflicts.

```
$ APP_NAME=lawlibrary DJANGO_SETTINGS_MODULE=lawlibrary.settings python manage.py rank_works --force
2023-05-20 07:46:29 INFO ranker none 79526 4309222784 Creating graph
2023-05-20 07:46:29 INFO ranker none 79526 4309222784 Created graph
2023-05-20 07:46:29 INFO ranker none 79526 4309222784 Running pagerank
2023-05-20 07:46:29 INFO ranker none 79526 4309222784 Finished pagerank
2023-05-20 07:46:29 INFO ranker none 79526 4309222784 Updated db rankings for 11 works
2023-05-20 07:46:29 INFO ranker none 79526 4309222784 updating index with 11 documents
/Users/greg/coding/laws.africa/peachjam/.env/lib/python3.8/site-packages/elasticsearch/connection/base.py:200: ElasticsearchWarning: Elasticsearch built-in security features are not enabled. Without authentication, your cluster could be accessible to anyone. See https://www.elastic.co/guide/en/elasticsearch/reference/7.16/security-minimal-setup.html to enable security.
  warnings.warn(message, category=ElasticsearchWarning)
2023-05-20 07:46:29 INFO ranker none 79526 4309222784 Updated index
```


peachjam/graph/ranker.py
```diff
@@ -102,7 +71,7 @@ def bulk_update(self, works):
             f"updating index with {len(docs)} documents",
         )
         searchable_doc = SearchableDocument()
-        actions = [
+        actions = (
```
longhotsummer, line 74: this makes it an generator, so it'll generate items only when the ES helper consumes them.

PR #1234, opened 2023-05-22
Fixes for article admin

longhotsummer:
* strip html from article summaries when importing
* fix display of import/export fields (jazzmin re-implements django-import-export's templates and uses the wrong (old?) variable)
* taxonomies can be blank
* ensure only published articles are externally visible
* publish/unpublish bulk operations
* filters for article admin view

image

image


actlikewill (dismissed):
This is good to go.

peachjam/models/article.py
```diff
@@ -11,6 +11,8 @@
 
 def file_location(instance, filename):
     filename = os.path.basename(filename)
+    print(instance.id)
+    print(instance.pk)
```
longhotsummer, line 15:
```suggestion
```

PR #1266, opened 2023-06-01
Quick search

### LINKED ISSUE #967: As a user I want to be able to quickly do a full-text search from a listing page
On Gazettes.Africa we can search from the country and year pages, which is useful and the search is automatically faceted to the correct country (and year, if set):

image

It would be useful to have this on the judgment listing pages (court detail pages), and most of the other listing pages.

On the court page, I suggest below the year navigation. Otherwise, just above the facets and listing.

The only pages we don't currently support this would be the taxonomy pages, because we can't facet on taxonomies yet.

actlikewill:
- adds quick search

image

closes #967

### DIFF
--- a/peachjam/templates/peachjam/_quick_search.html
+++ b/peachjam/templates/peachjam/_quick_search.html
@@ -1,0 +1,12 @@
+<form action="{% url 'search:search' %}" method="get" class="d-flex mb-3">
+  <input type="text"
+         class="form-control flex-grow-1"
+         autofocus
+         name="q"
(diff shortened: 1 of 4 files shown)


actlikewill:
@nickmwangemi I have added this feature to a few pages, i.e court detail, document nature and author. I'm not sure that Legislation pages can benefit from quick search, faceting on doctype. I'm okay to merge this as it is and we can decide whether to add it to legislation pages later.

nickmwangemi (approved):
LGTM

PR #1278, opened 2023-06-07
Citation re-extraction

### LINKED ISSUE #1156: As an admin I want the system to re-extract citations when new documents are uploaded.
* it's expensive, so if I upload 3 documents over the course of an hour (say), it should only do it after they're all uploaded
* do it intelligently: if I upload a document at date D, then only documents newer than D could have cited D, so only re-check those.

Without this, citation extraction is not happening correctly and users are missing out, in particular when old judgments are uploaded for the first time, or when alternative citations are added to existing judgments.

actlikewill:
- [x] Add CitationProcessing Model
- [x] Add signal to update CitationProcessing on doc save
- [x] Add Background task

closes #1156


actlikewill:
:fearful: We have an infinite loop.

The extraction process also calls model_save which updates the date, which is what kicks off the rextraction process.
We need to reset the CitationProcessor date when the extraction is done somehow.

peachjam/models/core_document_model.py
```diff
@@ -499,10 +499,14 @@ def pre_save(self):
...
             self.work.save()
 
-    def save(self, *args, **kwargs):
+    def save(self, *args, is_extracting_citations=False, **kwargs):
         # give ourselves and subclasses a chance to pre-populate derived fields before saving, in case full_clean() has
         # not yet been called
         self.pre_save()
+
+        # we use this flag in the re_extract_citations signal to prevent an infinite loop
+        self.is_extracting_citations = is_extracting_citations
```
actlikewill, line 508: @nickmwangemi I managed to fix the infinite loop by adding this flag. I call it when saving after extracting citations and check for it in the signal.

peachjam/models/citations.py
```diff
@@ -78,3 +84,36 @@ def for_target_works(cls, work):
...
+    def update_processing_date(self, date):
+        self.processing_date = date
+        self.save()
+
+    def re_extract_citations(self):
+        """Extract citations for all documents dated on or after the processing date."""
+        from peachjam.models import CoreDocument
+        from peachjam.tasks import extract_citations
+
+        later_documents = CoreDocument.objects.filter(
```
longhotsummer, line 101: This could load 100,000 documents, including all their content. Use `only("pk")`. You may also want to use `.iterator()` on in the for loop.

peachjam/models/citations.py
```diff
@@ -78,3 +84,36 @@ def for_target_works(cls, work):
...
+
+class CitationProcessing(SingletonModel):
+    processing_date = models.DateField(_("processing date"), null=True, blank=True)
+
+    def update_processing_date(self, date):
+        self.processing_date = date
+        self.save()
+
+    def re_extract_citations(self):
+        """Extract citations for all documents dated on or after the processing date."""
```
longhotsummer, line 97:
```suggestion
        """Queues up background tasks to re-extract citations for all documents dated on or after the processing date. This is to handle the case where an older document has just been added to the system, and newer documents may therefore cite it."""
```

peachjam/models/citations.py
```diff
@@ -78,3 +84,36 @@ def for_target_works(cls, work):
...
+        from peachjam.models import CoreDocument
+        from peachjam.tasks import extract_citations
+
+        later_documents = CoreDocument.objects.filter(
+            date__gte=self.processing_date
+        ).order_by("date")
+        log.info("Re-extracting citations for %s documents", later_documents.count())
```
longhotsummer, line 104: Also include the date in the log message, which will help us debug it if necessary.

peachjam/models/citations.py
```diff
@@ -78,3 +84,36 @@ def for_target_works(cls, work):
...
+        self.save()
+
+    def re_extract_citations(self):
+        """Extract citations for all documents dated on or after the processing date."""
+        from peachjam.models import CoreDocument
+        from peachjam.tasks import extract_citations
+
+        later_documents = CoreDocument.objects.filter(
+            date__gte=self.processing_date
+        ).order_by("date")
```
longhotsummer, line 103: Guard against `self.processing_date` being None.

peachjam/models/core_document_model.py
longhotsummer: Hmm, good point. Perhaps there's a cleaner way...

peachjam/signals.py
```diff
@@ -75,3 +75,14 @@ def doc_deleted_update_extracted_citations(sender, instance, **kwargs):
     """Update language list on related work after a subclass of CoreDocument is deleted."""
     if isinstance(instance, CoreDocument):
         update_extracted_citations_for_a_work(instance.work_id)
+
+
+@receiver(signals.post_save)
```
longhotsummer, line 80: What if we approach this a bit differently to avoid the circular logic issue.

Instead of queuing this up every time a document is saved, what if we explicitly queue it up in the admin form when a document is created/edited in the admin area, or if it's created/updated via an ingestor?

My motivation is that lots of other background logic (including citation extraction) saves a document, but that doesn't always mean we need to re-extract citations of future documents. We only actually want to do re-extract citations if the document is brand new, or has substantially changed. Let's assume that editing in the admin area means "substantially changed", and not background task updates.

Or, slightly differently, we use the signal handler to only handle the case where a document is created (ie. check for the creation flag in the handler), and then in the admin form we handle the "document changed" case.

That would allow us to remove the (rather odd) explicit flag. The flag is awkward because we're decorating the document to communicate to the signal handler.

actlikewill: @longhotsummer what if in the signal we check whether the document is new, or check only if the date has changed since that is the value we are most interested in?

longhotsummer: If we're going to restrict it to certain fields, that must be done in the Admin area because then we have access to the form which can tell us what fields changed. We'd need to ensure that the fields we monitor are those that are checked during citation matching. I think these are: alternative names, citation, title (and date).

peachjam/admin.py
```diff
@@ -413,8 +414,19 @@ class NewForm(self.new_document_form_mixin, form):
...
     def save_model(self, request, obj, form, change):
+        from peachjam.models import citations_processor
+
         if not change:
             obj.created_by = request.user
+
+        # if date, title, or content changed, re-extract citations or if new document
+        if (
+            not change
+            or ["date", "title", "alternative_name", "citation"] in form.changed_data
```
actlikewill, line 425: @longhotsummer It seems we can't  get changed_data for `alternative_names` here but we can get it in the save_related method. Should we add this logic there as well?

longhotsummer: Responding to changes in `alternative_names` is essential, so we have to do it somehow.

peachjam/admin.py
```diff
@@ -413,13 +414,30 @@ class NewForm(self.new_document_form_mixin, form):
...
+
+        alternative_names_formset = [
+            formset for formset in formsets if formset.model == AlternativeName
+        ][0]
+
+        if (
+            not change
+            or ["date", "title", "citation"] in form.changed_data
+            or alternative_names_formset.has_changed()
+        ):
```
actlikewill, line 436: @longhotsummer Here is what I went with for the alternative names.

actlikewill: @longhotsummer  I moved the logic to save_related where we can get the formsets.

peachjam/admin.py
```diff
@@ -413,13 +414,30 @@ class NewForm(self.new_document_form_mixin, form):
...
         if not change:
             obj.created_by = request.user
+
         super().save_model(request, obj, form, change)
 
     def save_related(self, request, form, formsets, change):
-        super().save_related(request, form, formsets, change)
         # after saving related models, also save this model again so that it can update fields based on related changes
+        # if date, title, or content changed, re-extract citations or if new document
+        from peachjam.models import citations_processor
```
longhotsummer, line 426: import this at the top level

longhotsummer (changes requested):
looks good, just the top-level import change pls; local imports get messy quickly

peachjam/admin.py
actlikewill: Changed :+1:

longhotsummer (approved):

PR #1293, opened 2023-06-13
Court registry drill-down

nickmwangemi:
This PR:
- adds registry detail pages that enabled a used to drill-down a court by its registries and further by years. This essentially adds the 2 new URLS below:
    * `/judgments/<court-code>/<registry-code>`
    * `/judgments/<court-code>/<registry-code>/<year>`
 - removes the unused code registry facet code

https://www.loom.com/share/7eb447a3f89a4e7cae795520e88d6968?sid=d9f8791c-06f8-46e8-8fb1-70d406b8eacf


peachjam/views/courts.py
```diff
@@ -30,18 +30,15 @@ def get_queryset(self):
...
 
-        years = sorted(
-            list(
-                self.model.objects.filter(court=self.court)
-                .order_by()
-                .values_list("date__year", flat=True)
-                .distinct()
-            ),
-            reverse=True,
+        years = self.model.objects.filter(court=self.court).dates(
```
longhotsummer, line 33: 🤯

peachjam/views/courts.py
```diff
@@ -67,14 +64,26 @@ def get_context_data(self, **kwargs):
...
+        registries = [
             {
-                registry
-                for registry in self.form.filter_queryset(
-                    self.get_base_queryset(), exclude="registries"
-                ).values_list("registry__name", flat=True)
-                if registry
+                "name": registry["name"],
+                "code": registry["code"],
+                "url": reverse(
```
longhotsummer, line 71: If you define `get_absolute_url()` on the Registry class, then you can simply use registry objects here, rather than copying the name and code into a new dictionary just so that you can include a url with the information. So the registry object knows how to build its own URL.

peachjam/views/courts.py
```diff
@@ -108,3 +116,58 @@ def get_context_data(self, **kwargs):
...
+            context["year"] = self.kwargs["year"]
+            context["formatted_registry_name"] = (
+                self.registry.name + " - " + str(self.kwargs["year"])
+            )
+
+        return context
+
+
+class CourtRegistryYearView(CourtRegistryDetailView):
+    queryset = Judgment.objects.prefetch_related("judges")
```
longhotsummer, line 172: this can fall away, because it's identical to how it's defined in the superclass `CourtRegistryDetailView`.

peachjam/views/courts.py
```diff
@@ -108,3 +116,58 @@ def get_context_data(self, **kwargs):
 
 class CourtYearView(CourtDetailView):
     queryset = Judgment.objects.prefetch_related("judges")
```
longhotsummer, line 118: This can fall away, it's identify to how it's defined in CourtDetailView.

peachjam/templates/peachjam/_registries.html
```diff
@@ -0,0 +1,19 @@
+{% load i18n %}
+{% if registries %}
+  <h4 class="my-4">Registries</h4>
+  {% for group in registry_groups %}
```
longhotsummer, line 4: a `.col-*` must always be included in a `.row` -- add `<div class="row">` above the for loop. that will also create the columns that you're missing.

peachjam/templates/peachjam/_registries.html
```diff
@@ -0,0 +1,19 @@
+{% load i18n %}
+{% if registries %}
+  <h4 class="my-4">Registries</h4>
```
longhotsummer, line 3:
```suggestion
  <h4>Registries</h4>
```

otherwise the title is too far separated from the list.

peachjam/templates/peachjam/court_registry_detail.html
longhotsummer, line 1: A lot of this is duplicated from `court_detail.html`. If you extend `court_detail.html`, you can avoid duplication and override only what is different.

peachjam/views/courts.py
```diff
@@ -108,3 +116,58 @@ def get_context_data(self, **kwargs):
...
+        self.court = get_object_or_404(Court, code=self.kwargs["code"])
+        self.registry = get_object_or_404(
+            CourtRegistry, code=self.kwargs["registry_code"]
+        )
+        return super().get_queryset()
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["registry"] = self.registry
+        context["formatted_registry_name"] = self.registry.name
```
longhotsummer, line 141: If you keep this as `formatted_court_name`, then it'll get used in the appropriate place in the template without you having to change anything. This is the reason we have `formatted_court_name` -- so that it can have extra data added if necessary, like the year or a registry.

peachjam/views/courts.py
```diff
@@ -108,3 +116,58 @@ def get_context_data(self, **kwargs):
...
+            }
+            for y in years
+        )
+        context["all_years_url"] = reverse(
+            "court_registry", args=[self.court.code, self.registry.code]
+        )
+
+        if "year" in self.kwargs:
+            context["year"] = self.kwargs["year"]
+            context["formatted_registry_name"] = (
```
longhotsummer, line 164: Do this at the same point as when you first build up `formatted_registry_name` above, rather than separately here.

longhotsummer (changes requested):
Looking great. Some reformatting to reduce duplication of code, mostly.

peachjam/views/courts.py
nickmwangemi: Yeah, blew my mind too!

nickmwangemi:
image

peachjam/templates/peachjam/_registries.html
```diff
@@ -0,0 +1,21 @@
+{% load i18n %}
+{% if registries %}
+  <h4>Registries</h4>
```
longhotsummer, line 3: This needs to be marked as translateable

longhotsummer (approved):

longhotsummer (approved):

PR #1297, opened 2023-06-14
Model translations

actlikewill:
- Model Translations


peachjam/js/components/FindDocuments/index.vue
```diff
@@ -268,14 +268,14 @@ export default {
       },
       {
         title: this.$t('Court'),
-        name: 'court',
+        name: JSON.parse(document.querySelector('#data-labels').textContent).court,
```
longhotsummer, line 271: Aren't these the names used in the form submission? I don't think they should be translated, that's what `title` above is for, and the translation is done (currently) in Javascript.

longhotsummer: ah, because we want to facet on the different language-based fields... I see.

actlikewill:
@longhotsummer I added the logic to get the translatable facets at runtime and it works well.l

peachjam_search/views.py
```diff
@@ -318,6 +318,45 @@ def __init__(self, *args, **kwargs):
         self.index = get_search_indexes(self.document._index._name)
         self.search = self.search.index(self.index)
 
+    def get_translatable_fields(self, request):
+        # get language from request
+        language = request.LANGUAGE_CODE
```
longhotsummer, line 323: shouldn't this be using `get_language_from_request` like we do elsewhere?

peachjam_search/views.py
```diff
@@ -318,6 +318,45 @@ def __init__(self, *args, **kwargs):
...
+        # suffix for translatable fields e.g. court_en, court_fr etc
+        suffix = ""
+        if language in [lang for lang, _ in settings.LANGUAGES]:
+            suffix = "_" + language
+
+        translatable_facets = {
+            "court": {"field": f"court{suffix}", "options": {"size": 100}},
```
longhotsummer, line 330: The configuration is now duplicated.

Instead, adjust the existing configuration -- only the `field` field needs to change.

longhotsummer: Perhaps a hard-coded list of the translatable fields (eg. `["court", "nature", ...]`) which you then use to introspect and adjust the `facets` and `filters` fields.

peachjam_search/views.py
```diff
@@ -318,6 +318,45 @@ def __init__(self, *args, **kwargs):
...
+
+        return translatable_facets, translatable_filters
+
+    def list(self, request, *args, **kwargs):
+        # get translatable facets and filters
+        translatable_facets, translatable_filters = self.get_translatable_fields(
+            request
+        )
+        # add translatable facets
+        self.faceted_search_fields.update(translatable_facets)
```
longhotsummer, line 354: This currently will update the field as it is on the original class. The next request will use the updated fields, which isn't correct. Instead, deep-copy `self.faceted_search_fields` and `self.filter_fields` and then update the copy.

longhotsummer (changes requested):
Looking good, some changes required.

Do you need to change the `Hit` object that comes back to ensure, say, the `court` field is actually `court_fr`?

actlikewill:
> Looking good, some changes required.
> 
> Do you need to change the `Hit` object that comes back to ensure, say, the `court` field is actually `court_fr`?

After checking, I don't think we need to touch the hit objects since we are changing the "inner" field that elasticsearch is using to do the lookup. The result object comes with the variable assigned to the correct field. Unless there is something I'm missing?

longhotsummer:
> > Looking good, some changes required.
> > Do you need to change the `Hit` object that comes back to ensure, say, the `court` field is actually `court_fr`?
> 
> After checking, I don't think we need to touch the hit objects since we are changing the "inner" field that elasticsearch is using to do the lookup. The result object comes with the variable assigned to the correct field. Unless there is something I'm missing?

Help me rest easy: can you show me an example of the JSON that is returned to the client, that has an appropriately translated value for `court` or other translated field?

actlikewill:
@longhotsummer Adding serializer method fields seems to do the trick

image

longhotsummer:
Nice, this is looking good.

peachjam_search/views.py
```diff
@@ -318,6 +321,32 @@ def __init__(self, *args, **kwargs):
...
+
+        for field in translatable_fields:
+            self.filter_fields[field] = self.filter_fields[field] + suffix
+            self.faceted_search_fields[field]["field"] = (
+                self.faceted_search_fields[field]["field"] + suffix
+            )
+
+    def list(self, request, *args, **kwargs):
+        # TODO: uncomment when we have reindexd the data
+        # self.get_translatable_fields(request)
```
actlikewill, line 347: @longhotsummer I've commented out this as well

longhotsummer (approved):

PR #1300, opened 2023-06-16
Tweaks

longhotsummer:
Small improvements discovered during demos over the last few days.

* unstyled lists have spacing between items, so that wrapped lines on narrow displays are easier to see 
image

* make copy buttons outline-secondary, so they're more subtle and align with the colouring of the language and date dropdowns 
image

* use the "multiple languages available" icon in the document detail view, so that there's a similarity between document listings and document detail. Also use bootstrap tooltips to show the textual description for the icon 
image

* add eswatinilii to africanlii.org homepage
image

* scroll tabbed navs on mobile https://www.loom.com/share/752e20b6ebbe4e448ea8c5b930a92cd5

* use iconography for document notices
image
image
image


actlikewill:
@longhotsummer while we are here, should we make these cards all the same height? 

image

longhotsummer:
> @longhotsummer while we are here, should we make these cards all the same height?
> 
> image

I'd rather do that later, because I don't want to have to work out how to get all the images etc. to look correct on the  different viewport sizes (ie widths) AND heights.

actlikewill (approved):

PR #1307, opened 2023-06-19
Add eSwatiniLII resolver mapping

nickmwangemi:


peachjam/views/documents.py
```diff
@@ -74,6 +74,10 @@ class RedirectResolver:
             "country_code": "zw",
             "domain": "zimlii.org",
         },
+        "eswatinilii": {
```
longhotsummer, line 77: I suggest keeping the list sorted alphabetically.

longhotsummer (approved):

PR #1312, opened 2023-06-20
Import html

actlikewill:
- adds html to import


longhotsummer (approved):
Are there any weird interactions with the source file field?

actlikewill:
> Are there any weird interactions with the source file field?

@longhotsummer If source url is present it overrides the content html.

PR #1325, opened 2023-06-22
correctly extract links from html. fixes #1322

### LINKED ISSUE #1322: Bug: we're not correctly indexing extracted citations in non-AKN HTML
See https://github.com/laws-africa/peachjam/blob/eb920acab9051b8f1be535654c164f8f4e3862e4/peachjam/models/core_document_model.py#L606

We handle AKN HTML well, but not generic HTML.

eg. in https://lawlibrary.org.za/akn/za/judgment/zasca/2023/103/eng@2023-06-22

image

Those are not being indexed for listing in the Citation tab.

longhotsummer:


actlikewill (approved):
Works well.

PR #1342, opened 2023-06-27
Shortened citation lists

nickmwangemi:
Before:
image

After:
image


peachjam/views/generic_views.py
```diff
@@ -165,16 +167,44 @@ def get_context_data(self, **kwargs):
...
+
         return context
 
+    def fetch_docs(self, works):
+        docs = sorted(
+            list(
+                CoreDocument.objects.prefetch_related("work")
+                .filter(work__in=works)
+                .distinct("work_frbr_uri")
+                .order_by("work_frbr_uri")
```
nickmwangemi, line 191: .order_by("work_frbr_uri", "-date")

peachjam/templates/peachjam/_citations_list.html
```diff
@@ -0,0 +1,34 @@
...
+  {{ item.doc_type }} <span class="badge bg-secondary">{{ item.docs|length }}</span>
+  <ol>
+    {% if item.docs|length < 11 %}
+      {% for doc in item.docs|slice:":10" %}
+        <li>
+          <a href="{% url 'document_detail' frbr_uri=doc.work.frbr_uri|strip_first_character %}">{{ doc.title }}</a>
+        </li>
+      {% endfor %}
+    {% endif %}
+    {% if item.docs|length > 10 %}
```
nickmwangemi, line 12: change to {% else %}

peachjam/templates/peachjam/_citations_list.html
```diff
@@ -0,0 +1,33 @@
...
+        <li>
+          <a href="{% url 'document_detail' frbr_uri=doc.work.frbr_uri|strip_first_character %}">{{ doc.title }}</a>
+        </li>
+      {% endfor %}
+    {% else %}
+      {% for doc in item.docs|slice:":10" %}
+        <li>
+          <a href="{% url 'document_detail' frbr_uri=doc.work.frbr_uri|strip_first_character %}">{{ doc.title }}</a>
+        </li>
+      {% endfor %}
```
longhotsummer, line 16: The two for loops are identical. You always want to show the first 10 items, no matter if there are less than 10 or more than 10. So the first ten can be shown, always, outside of an if statement. Then the exceptional case is when there are more than 10.

peachjam/templates/peachjam/_citations_list.html
```diff
@@ -0,0 +1,27 @@
...
+  {{ item.doc_type }} <span class="badge bg-secondary">{{ item.docs|length }}</span>
+  <ol>
+    {% for doc in item.docs|slice:":10" %}
+      <li>
+        <a href="{% url 'document_detail' frbr_uri=doc.work.frbr_uri|strip_first_character %}">{{ doc.title }}</a>
+      </li>
+    {% endfor %}
+    {% if item.docs|length > 10 %}
+      {% for doc in item.docs|slice:"10:" %}
+        <li class="collapse {{ item.doc_type }}">
```
longhotsummer, line 12: If `{{ item.doc_type }}` is "Concept note", then the class will be `class="collapse Concept note"` which may cause it to have odd styling.

Instead, use the outermost for-loop's counter, which is unique for the group:

```class="collapse grp-{{ forloop.parentloop.counter }}"```

longhotsummer: You'll also need to pass in something to the template so that the class  can differentiate between the same groups in the incoming and outgoing citation lists.

longhotsummer (approved):
I've made some small tweaks. Nice, let's ship it.

PR #1343, opened 2023-06-27
Text search

### LINKED ISSUE #967: As a user I want to be able to quickly do a full-text search from a listing page

actlikewill:
- Adds text search for listing pages
- Adds text search for gazettes

closes #967 

https://www.loom.com/share/72e6e54142ca474daf17028b9b802ab3?sid=13a3a7e8-cea1-4d14-ba19-cf0cacc65f0e

### DIFF
--- a/peachjam/templates/peachjam/_quick_search.html
+++ b/peachjam/templates/peachjam/_quick_search.html
@@ -3,11 +3,13 @@
          class="form-control flex-grow-1"
          autofocus
          name="q"
-         placeholder="Search {{ court.name }} {{ author.name }} {{ nature.name }} {{ registry.name }} {{ year }}"/>
+         placeholder="Search {{ court.name }} {{ author.name }} {{ nature.name }} {{ registry.name }} {{ locality.name }} {{ year }} "/>
   <input type="submit" class="btn btn-outline-primary ms-2" value="Search"/>
   {% if year %}<input type="hidden" name="year" value="{{ year }}"/>{% endif %}
   {% if court %}<input type="hidden" name="court" value="{{ court }}"/>{% endif %}
   {% if author %}<input type="hidden" name="authors" value="{{ author }}"/>{% endif %}
   {% if nature %}<input type="hidden" name="nature" value="{{ nature }}"/>{% endif %}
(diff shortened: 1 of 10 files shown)


peachjam/templates/peachjam/_gazette_search.html
longhotsummer, line 1: Why do gazettes need their own special one? Can we unify it into the generic one?

longhotsummer (commented):
Great. Just wondering about the need for the gazette-specific one

peachjam/templates/peachjam/_gazette_search.html
```diff
@@ -0,0 +1,11 @@
+<form action="{% url 'search:search' %}" method="get" class="d-flex mb-3">
+  <input type="text"
+         class="form-control flex-grow-1"
+         autofocus
+         name="q"
+         placeholder="Search {{ locality.name }} {{ year }} Gazettes"/>
+  <input type="submit" class="btn btn-outline-primary ms-2" value="Search"/>
+  {% if year %}<input type="hidden" name="year" value="{{ year }}"/>{% endif %}
+  {% if locality %}<input type="hidden" name="locality" value="{{ locality }}"/>{% endif %}
+  <input type="hidden" name="doc_type" value="Gazette"/>
```
actlikewill, line 10: @longhotsummer The gazette specific one facets on doc_type=Gazette. I could try adding doc_type facet to the generic one and adding the doc_type context to the listing pages.

actlikewill: @longhotsummer Let me add that

actlikewill: Changed :+1:

longhotsummer (approved):

PR #1345, opened 2023-06-28
Remove twitter embed and add social media section

nickmwangemi:
image


liiweb/templates/peachjam/_footer.html
```diff
@@ -13,6 +13,39 @@ <h2 class="mb-2">{% trans 'Subscribe' %}</h2>
...
+        <div class="col-12 col-md-6 mb-2">
+          <h5 class="mb-2">{% trans 'Follow us on social media' %}</h5>
+          <ul class="list-inline">
+            {% if PEACHJAM_SETTINGS.facebook_link %}
+              <li class="list-inline-item">
+                <a href="{{ PEACHJAM_SETTINGS.facebook_link }}"
+                   target="_blank"
+                   rel="noreferrer">
+                  <img src="{% static 'images/facebook.svg' %}"
+                       alt="AfricanLII Facebook"
```
longhotsummer, line 27:
```suggestion
                       alt="Facebook"
```

liiweb/templates/peachjam/_footer.html
```diff
@@ -13,6 +13,39 @@ <h2 class="mb-2">{% trans 'Subscribe' %}</h2>
...
+                </a>
+              </li>
+            {% endif %}
+            {% if PEACHJAM_SETTINGS.twitter_link %}
+              <li class="list-inline-item">
+                <a href="{{ PEACHJAM_SETTINGS.twitter_link }}"
+                   target="_blank"
+                   rel="noreferrer">
+                  <img src="{% static 'images/twitter.svg' %}"
+                       alt="AfricanLII Twitter"
```
longhotsummer, line 39:
```suggestion
                       alt="Twitter"
```

liiweb/templates/peachjam/_social_media.html
```diff
@@ -0,0 +1,30 @@
+{% load i18n static %}
+<div class="col-12 col-md mb-2">
```
longhotsummer, line 2: Rule of thumb: a `.col` must always be a direct child of a `.row` - so if you can't guarantee that, then there's something up.

So, this line is contextual - it depends on how this file is included. This is only necessary when it's included in the footer. So remove it from here, and just have the h5 onwards.

liiweb/templates/peachjam/_footer.html
```diff
@@ -13,6 +13,11 @@ <h2 class="mb-2">{% trans 'Subscribe' %}</h2>
     <div class="col-12 col-md-6 mb-2">
       {% block newsletter-form-code %}{% endblock %}
     </div>
+    {% if PEACHJAM_SETTINGS.facebook_link or PEACHJAM_SETTINGS.twitter_link %}
```
longhotsummer, line 16: If you move this into the above `div`, then you don't need to worry about cols and rows and stuff.

longhotsummer (approved):
Super.

PR #1352, opened 2023-06-30
Article attachments

### LINKED ISSUE #1277: As a SierraLII user, I want to be able to attach pdf files of the articles I am uploading to ensure they can be downloaded.

actlikewill:
- Adds attachments to articles

image

closes #1277


peachjam/models/core_document_model.py
```diff
@@ -755,6 +755,16 @@ def extension(self):
         return os.path.splitext(self.filename)[1].replace(".", "")
 
 
+class ArticleAttachment(AttachmentAbstractModel):
```
longhotsummer, line 758: Is it not possible to put this in article.py?

actlikewill: @longhotsummer unfortunately no, we get circular import errors. Is there a good way to handle circular imports in our modules? it keeps coming up.

longhotsummer: Hmm, depends on the context. Suggestions:

* put abstract models into a separate file so that they can be imported everywhere
* use strings to refer to other models (like you have done below)
* local imports

longhotsummer (approved):

Issue #1161, opened 2023-05-08
Main issue for search improvements


longhotsummer:
Meta issue for discussing search improvements.

Please add examples of searches where users didn't get what they had hoped for. Include:

* what the user searched for
* what they were hoping to find
* any other relevant information

longhotsummer:
Searching with both the title and the citation doesn't work well, because no field contains both of these values. eg. airports company act 44 of 1993

Fixed in: https://github.com/laws-africa/peachjam/pull/1228

longhotsummer:
A user emailed just with the term: "evidence law uganda". On ulii.org, searching for "evidence law" has really poor results. Searching for "evidence act" or just "evidence" has better legislation results.

longhotsummer:
Some thoughts:

* Use minimum should match on phrase fields to allow OR but most those that have most results https://www.elastic.co/guide/en/elasticsearch/guide/current/match-multi-word.html -- NB this could make a huge difference

* index full text fields into multiple indexes that use stemming and shingles. Combine those to boost better results - https://www.elastic.co/guide/en/elasticsearch/guide/current/most-fields.html
* Shingles help to find words close together https://www.elastic.co/guide/en/elasticsearch/guide/current/_closer_is_better.html
* see https://medium.com/unacademy-engineering/making-search-relevant-using-elasticsearch-a7b546c9a72a

longhotsummer:
changing to OR rather than AND, and using `minimum_should_match=70%` fixes these searches:

* evidence law (finds "evidence act")
* constitutive act of the AU (finds "constitutive act of the african union")

longhotsummer:
When searching for legislation, if a TOC title is a strong match, show the TOC breakdown and/or show the TOC breakdown for general legislation text matches.

For example, "small-scale mining licenses" is a TOC entry in https://sierralii.gov.sl/akn/sl/act/2023/16/eng@2023-05-12 and should be very prominent for https://sierralii.gov.sl/search/?q=small-scale+mining+licence&doc_type=Legislation

longhotsummer:
https://github.com/laws-africa/peachjam/pull/1647 boosts titles more, and prevents double-scoring for documents without a citation. This makes a search for "protection of information act" have POPIA at the top, as it should be.

longhotsummer:
See https://github.com/laws-africa/peachjam/issues/1655 for a case where missing a space in the title of a document results in poor search results.

longhotsummer:
[...]

PR #1360, opened 2023-07-04
Search improvements

longhotsummer:
* use minimum_should_match (see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html and https://github.com/laws-africa/peachjam/issues/1161)
* index `case_name` for citator searches
* import/export allows created_at

### DIFF
--- a/peachjam/resources.py
+++ b/peachjam/resources.py
@@ -250,7 +250,6 @@ class BaseDocumentResource(resources.ModelResource):
 
     class Meta:
         exclude = (
-            "created_at",
             "updated_at",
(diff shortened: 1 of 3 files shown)


longhotsummer:
changing to OR rather than AND, and using minimum_should_match=70% fixes these searches:

evidence law (finds "evidence act")
constitutive act of the AU (finds "constitutive act of the african union")

actlikewill (approved):
Great. Search results seem to be better with this.

PR #1361, opened 2023-07-06
Resize Images

### LINKED ISSUE #1323: To reach optimal site performance, we should implement properly sized images

actlikewill:
- resizes images

closes #1323


actlikewill:
@nickmwangemi Perhaps if you have a large screen you can test these images and see if they render okay.

longhotsummer:
Looks good!

liiweb/static/images/banner-small.jpg
longhotsummer, line 1: where does this get used?

longhotsummer: oh never mind, I see -- clever

longhotsummer (approved):

liiweb/static/images/banner-small.jpg
actlikewill: @longhotsummer The tool that chrome supplies to lint images doesn't like different aspect ratios within a set of `srcset` images. They say that you should not use `srcset` for "art direction". But a small banner image becomes very pixelated, even on moblie, so I only cropped the width. Lighthouse is happy with it.

PR #1365, opened 2023-07-07
test for translated field; meta programming

longhotsummer:
* define prepare_... methods dynamically
* prepare method checks for existing of attribute before calling it, because different sites may have different languages configured

### DIFF
--- a/peachjam_search/documents.py
+++ b/peachjam_search/documents.py
@@ -55,10 +55,10 @@ class SearchableDocument(Document):
 
     # Judgment
     court = fields.KeywordField(attr="court.name")
-    court_en = fields.KeywordField(attr="court.name_en")
-    court_sw = fields.KeywordField(attr="court.name_sw")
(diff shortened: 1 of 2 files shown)


peachjam_search/documents.py
```diff
@@ -55,10 +55,10 @@ class SearchableDocument(Document):
 
     # Judgment
     court = fields.KeywordField(attr="court.name")
-    court_en = fields.KeywordField(attr="court.name_en")
-    court_sw = fields.KeywordField(attr="court.name_sw")
-    court_fr = fields.KeywordField(attr="court.name_fr")
-    court_pt = fields.KeywordField(attr="court.name_pt")
+    court_en = fields.KeywordField()
```
longhotsummer, line 58: Because we use `prepare_...` I don't think the `attr` does anything.

actlikewill (approved):
Works okay.

PR #1374, opened 2023-07-17
Setup Open By-laws

nickmwangemi:
https://www.loom.com/share/ab70210054ea405d85e737d5ef0d7e19?sid=0603acc4-d4c4-496c-b0f5-ab589d9fe002


open_by_laws/urls.py
```diff
@@ -0,0 +1,23 @@
+from django.urls import include, path
+
+from open_by_laws.views import HomePageView, MunicipalByLawsView
+
+urlpatterns = [
+    path("", HomePageView.as_view(), name="home_page"),
```
longhotsummer, line 6: We'll need a URL to redirect from the old place codes, such as `za-xxx/yyy` to `bylaws/xxx`

longhotsummer: We'll also need to redirect the by-law detail page so that we don't lose old links:

eg:

```
/za-cpt/act/by-law/2016/air-quality-management/eng/
->
/akn/za-cpt/act/by-law/2016/air-quality-management/
```

I suggest: `/za-<str>/act/by-law/<path>` 

* if the path ends in `/xxx/` then strip that
* add `/akn/` at the start and redirect

lawlibrary/views/legislation.py
```diff
@@ -10,27 +9,18 @@ def get_queryset(self):
         return super().get_queryset().filter(locality=None)
 
 
-class ProvincialLegislationView(TemplateView):
-    template_name = "lawlibrary/provincial_legislation.html"
+class LocalityLegislationView(BaseLocalityLegislationView):
     navbar_link = "legislation/provincial"
```
longhotsummer, line 13: All of the core logic of this view can move into `BaseLocalityLegislationView`. The only thing that really needs to live here is to customise `page_heading` to `Provincial Legislation`.

lawlibrary/views/legislation.py
```diff
@@ -10,27 +9,18 @@ def get_queryset(self):
...
 
     def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
         codes = "mp ec nc kzn gp wc lp nw fs".split()
         provinces = Locality.objects.filter(code__in=codes)
```
longhotsummer, line 19: Simply load all localities. Lawlibrary only has provinces as localities any way.

lawlibrary/views/legislation.py
```diff
@@ -10,27 +9,18 @@ def get_queryset(self):
...
-        groups = provinces[:5], provinces[5:]
-        return super().get_context_data(province_groups=groups, **kwargs)
+        context["province_groups"] = provinces[:5], provinces[5:]
```
longhotsummer, line 20: When moving to the base class, change this to break the list of provinces into two groups based on the length of the locality list (which is hard-coded here).

lawlibrary/views/legislation.py
```diff
@@ -10,27 +9,18 @@ def get_queryset(self):
...
+        context["page_heading"] = "Provincial Legislation"
+        return context
 
 
-class ProvincialLegislationListView(BaseLegislationListView):
-    template_name = "lawlibrary/provincial_legislation_list.html"
+class LocalityLegislationListView(BaseLocalityLegislationListView):
     navbar_link = "legislation/provincial"
```
longhotsummer, line 26: Move this `navbar_link` into the superclass (eg. as `legislation/locality`), and this class can fall away completely.

longhotsummer:
It's looking good. As you're doing the refactoring to move the provincial logic into liiweb, consider that any use of teh word "provincial" should be replaced with locality.

liiweb/views/legislation.py
```diff
@@ -66,3 +67,33 @@ def add_children(self, queryset):
...
         return qs
+
+
+class LocalityLegislationView(TemplateView):
+    template_name = "liiweb/locality_legislation.html"
+    navbar_link = "legislation/locality"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        provinces = Locality.objects.all()
```
longhotsummer, line 78: `provincies` -> `localities`

lawlibrary/urls.py
```diff
@@ -1,14 +1,15 @@
...
+from liiweb.views import LocalityLegislationListView
 
 urlpatterns = [
     path("", views.HomePageView.as_view(), name="home_page"),
     path("legislation/", views.LegislationListView.as_view(), name="legislation_list"),
     path(
-        "legislation/provincial",
-        views.ProvincialLegislationView.as_view(),
-        name="provincial_legislation",
+        "legislation/locality",
```
longhotsummer, line 10: I think we should keep the existing URL of `legislation/provincial` because it's semantically meaningful.

lawlibrary/views/legislation.py
```diff
@@ -1,36 +1,14 @@
...
-        return super().get(*args, **kwargs)
-
-    def get_queryset(self):
-        return super().get_queryset().filter(locality=self.locality)
-
+class LocalityLegislationView(BaseLocalityLegislationView):
     def get_context_data(self, **kwargs):
-        return super().get_context_data(locality=self.locality, **kwargs)
+        return super().get_context_data(
+            locality_legislation_title="Locality Legislation", **kwargs
```
longhotsummer, line 13:
```suggestion
            locality_legislation_title="Provincial Legislation", **kwargs
```

nickmwangemi:
OBL
image
image


LawLibrary
image
image

longhotsummer:
In the "Cape Town Legislation Page" on lawlibrary the breadcrumb should be "Provincial Legislation", not "Locality Legislation"

longhotsummer (approved):

PR #1384, opened 2023-07-20
Place official municipal partners in separate section

nickmwangemi:
image

Builds-on #1374


open_by_laws/views/home.py
```diff
@@ -7,6 +7,16 @@ class HomePageView(LiiWebPageView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        codes = "cpt eth jhb wc033 wc011 ec443 wc013 wc012 wc015 wc041 wc023".split()
-        context["municipalities"] = Locality.objects.filter(code__in=codes)
+
+        codes = "cpt eth jhb wc015 wc041 wc023".split()
```
longhotsummer, line 11: This will be stored locally as Locality objects, but the others below won't be, because we don't want them showing up in the database.

So for the ones below, please just put together fake localities like:

```python
[{
  "name": "Bergrivier", "code": "za-wc013", "url": "https://bergrivier.openbylaws.org.za",
  ...
}]

File: open_by_laws/templates/open_by_laws/_municipalities.html [hunk 2]
```diff
@@ -0,0 +1,12 @@
+{% load static %}
+<div class="row">
+  {% for item in object_list %}
+    <div class="col-md-4">
+      <a class="placard {{ item.code }}"
+         href="{% url 'municipal_by_laws' item.code %}"
```
longhotsummer, line 6: Does this need to check for `item.url` and use that if available?

longhotsummer:
I've merged this into my microsites branch.

PR #1385, opened 2023-07-20
Add missing static files

nickmwangemi:


longhotsummer:
fixed through a push to master.

PR #1388, opened 2023-07-21
Use through model on judgment.judges to preserve ordering

longhotsummer:
This way, the order in which the judges are added to the decision in the admin area is preserved when presented to the user. We don't have to explicitly add an ordering field -- the editor must just enter the judges in the correct order.

We need to tweak some of the older migrations to work with theh existing data, since nothing is changing under the hood. Only Django's view of the database and how it works with it is changing.

https://www.loom.com/share/1bb619637d2844a7a53ed55e87ce7120

image

image


actlikewill (approved):
Works well. Curious what effect changing the older migration has. Why do we need to do that?

longhotsummer:
> Works well. Curious what effect changing the older migration has. Why do we need to do that?

Otherwise django thinks the table is somehow different and creates a migration that, when run, clashes with reality. I tried `managed = False` without luck. So instead, I'm just pretending that it created the table like that right at the beginning, and it's happy -- it doesn't think anything has changed.

nickmwangemi (approved):

PR #1404, opened 2023-07-25
TypeError: Cannot read properties of undefined (reading 'Tab')

Sandravaphilips:
Issue https://github.com/laws-africa/peachjam/issues/1334

We're already importing the Tab Sass and js modules. I found the problem was bootstrap being undefined when setting up tabs for document content. To fix it, I checked if it's been loaded first before setting up the tabs

### DIFF
--- a/peachjam/js/components/DocumentContent/index.ts
+++ b/peachjam/js/components/DocumentContent/index.ts
@@ -7,11 +7,12 @@ import { createTocController, generateHtmlTocItems, wrapTocItems } from '../../u
 import EnrichmentsManager from './enrichments-manager';
 import i18next from 'i18next';
 
+const bootstrap = (window as { [key: string]: any }).bootstrap;
 class OffCanvas {
   protected offCanvas: any;
   body: HTMLElement | null;
   constructor (element: HTMLElement) {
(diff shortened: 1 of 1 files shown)


peachjam/js/components/DocumentContent/index.ts
```diff
@@ -7,11 +7,12 @@ import { createTocController, generateHtmlTocItems, wrapTocItems } from '../../u
 import EnrichmentsManager from './enrichments-manager';
 import i18next from 'i18next';
 
+const bootstrap = (window as { [key: string]: any }).bootstrap;
```
longhotsummer, line 10: When does the boostrap JS get imported? Because this line will be run the moment it's encountered in the JS, so the bootstrap code must be run before it -- do we know that will always happen?

Sandravaphilips: So, bootstrap sometimes isn't loaded before the previous line `(window as { [key: string]: any }).bootstrap.Tab(tocTabTriggerEl)` runs. That's why Bootstrap returns undefined. This just prevents the Tab line from running before Bootstrap loads.

But, ideally, Bootstrap should load before that line (works locally too). Or would you rather I move it to the setupTabs method?

longhotsummer:
Resolved by #1405

Issue #1318, opened 2023-06-22
Reduce unused Javascript to decrease bytes consumed by network activity as a measure to improve performance scores.


niiroobiro:
![image](https://github.com/laws-africa/peachjam/assets/4178542/2aa29029-4a62-4100-b3ac-c5b371977b01)

longhotsummer:
See https://github.com/laws-africa/peachjam/issues/1324 for reducing unused CSS.

longhotsummer:
A bigger target is to reduced unused JS.

I'm going to change this issue to be about that.

longhotsummer:
done

PR #1405, opened 2023-07-25
Reduce unused Javascript to decrease bytes consumed by network activity as a measure to improve performance scores.

Sandravaphilips:
Issue https://github.com/laws-africa/peachjam/issues/1318

There were only a few modules not being used, so I'm unsure if it'll significantly reduce bytes. I tried looking for a way to use a CDN for this project, but it was taking too long to find a working method. 

For now, the files are still present in Bootstrap; they just aren't imported.


longhotsummer:
@Sandravaphilips how easy is it to explicitly import the bootstrap JS into our typescript/javascript code, since we use webpack? Then it  would be included in app.js and not imported using a separate `<script>` tag.

See the docs here: https://getbootstrap.com/docs/5.0/customize/optimize/#lean-javascript

Sandravaphilips:
@longhotsummer is this what you meant? If it is, then it'll also solve the Tab issue of  #1334

longhotsummer (approved):
yup, spot-on

longhotsummer:
Pls don't merge this quite yet.

longhotsummer:
@Sandravaphilips rather than importing the js from the bundled js, can you see if it's possible to do that from an npm-installed version of bootstrap via package.json? Then we don't need to host the bootstrap JS in our repo at all.

longhotsummer:
I've done this for the SCSS, it should be straight-forward to do it for the javascript now too. Thanks.

Sandravaphilips:
@longhotsummer It's done now.

longhotsummer:
I had to add @types/bootstrap for it to compile, but this looks good to me.

Sandravaphilips:
Oh, I forgot I installed that. I usually avoid pushing changes to lock.json, so I only push the files I worked on. I'll merge now

longhotsummer (approved):

PR #1406, opened 2023-07-25
Add labels for documents

nickmwangemi:
- Create `Label` model
- Add field `labels` to `CoreDocument` model
- Apply labels when `CoreDcoument` instance is saved
- `apply_labels` method on the document object, which can be called by the admin view

image

image

image

image


peachjam/models/generic_document.py
```diff
@@ -85,6 +86,13 @@ class Meta(CoreDocument.Meta):
     def __str__(self):
         return self.title
 
+    def apply_labels(self):
+        if self.repealed:
```
longhotsummer, line 90: don't forget to call `super().apply_labels()`

longhotsummer: you also need to handle the case where the label is there incorrectly, and should be removed.

peachjam/models/core_document_model.py
```diff
@@ -37,6 +38,21 @@
 from peachjam.storage import DynamicStorageFileField
 
 
+class Label(models.Model):
+    name = models.CharField(
```
longhotsummer, line 42: don't forget that this must be translateable.

peachjam/templates/peachjam/_labels.html
```diff
@@ -0,0 +1,8 @@
+<div class="d-flex align-items-center">
+  {% if labels %}
+    {% for label in labels %}
+      {% if label.code == "repealed" %}<span class="badge rounded-pill bg-danger">Repealed</span>{% endif %}
```
longhotsummer, line 4: "Repealed" should be `label.name`

peachjam/templates/peachjam/_labels.html
```diff
@@ -0,0 +1,8 @@
+<div class="d-flex align-items-center">
+  {% if labels %}
```
longhotsummer, line 2: Move the `if` outside the `div` -- otherwise there's no point to the if block. We don't want to render an empty div, though.

peachjam/models/generic_document.py
```diff
@@ -85,6 +86,23 @@ class Meta(CoreDocument.Meta):
...
         return self.title
 
+    def apply_labels(self):
+        # label to indicate that this legislation is repealed
+        label, _ = Label.objects.get_or_create(
+            code="repealed", defaults={"name": "Repealed", "code": "repealed"}
+        )
+        # apply label if repealed
+        if self.metadata_json.get("repeal"):
+            self.repealed = True
```
longhotsummer, line 96: This method doesn't need to be responsible for setting this field. Firstly because applying labels should be driven by this field (so setting it doesn't make sense) and secondly because this is done by the indigo ingestor.

So drop this line, and rely on `self.repealed` rather than `metadata_json`.

See adapters.py line 215 for where it is set.

peachjam/models/core_document_model.py
```diff
@@ -407,6 +424,21 @@ class Meta:
     def __str__(self):
         return f"{self.doc_type} - {self.title}"
 
+    def apply_labels(self):
+        """Apply labels to this document based on its properties."""
+        # label showing that a document is cited/reported in law reports, hence "more important"
+        label, _ = Label.objects.get_or_create(
```
longhotsummer, line 430: The concept of "reported" only applies to judgments, so this should all move to the judgment object.

So `CoreDocument.apply_labels()` is a no-op and doesn't do anything.

peachjam/models/core_document_model.py
```diff
@@ -37,6 +38,21 @@
 from peachjam.storage import DynamicStorageFileField
 
 
+class Label(models.Model):
+    name = models.CharField(
+        _("name"), max_length=1024, unique=True, null=False, blank=False
+    )
+    code = models.SlugField(_("code"), max_length=1024, unique=True)
```
longhotsummer, line 45: Don't forget a Meta with a default sorting on `name`.

peachjam/admin.py
```diff
@@ -492,14 +499,25 @@ def reextract_content(self, request, queryset):
...
             search_model_saved(doc._meta.label, doc.pk)
         self.message_user(request, f"Queued tasks to re-index for {count} documents.")
 
     reindex_for_search.short_description = "Re-index for search (background)"
 
+    def apply_labels(self, request, queryset):
+        """Set up a background task to apply labels to documents."""
+        count = queryset.count()
+        for doc in queryset:
+            apply_labels(doc.pk)
```
longhotsummer, line 514: i suggest just applying the labels directly -- it's not a risky or long-running call

peachjam/models/core_document_model.py
```diff
@@ -508,6 +540,8 @@ def save(self, *args, **kwargs):
         # give ourselves and subclasses a chance to pre-populate derived fields before saving, in case full_clean() has
         # not yet been called
         self.pre_save()
+        # apply labels
+        self.apply_labels()
```
longhotsummer, line 544: move after the next line so that self is guaranteed to have a pk

peachjam/admin.py
```diff
@@ -492,14 +498,23 @@ def reextract_content(self, request, queryset):
...
+        """Set up a background task to re-index documents for search."""
         count = queryset.count()
         for doc in queryset:
             search_model_saved(doc._meta.label, doc.pk)
         self.message_user(request, f"Queued tasks to re-index for {count} documents.")
 
     reindex_for_search.short_description = "Re-index for search (background)"
 
+    def apply_labels(self, request, queryset):
+        """Set up a background task to apply labels to documents."""
```
longhotsummer, line 510:
```suggestion
```

peachjam/js/components/FindDocuments/SearchResult.vue
```diff
@@ -74,6 +88,13 @@ export default {
       default: false,
     }
   },
+  data (){
+    const documentLabels = JSON.parse(document.querySelector('#data-labels').textContent).documentLabels;
+    console.log(documentLabels);
```
longhotsummer, line 93:
```suggestion
```

peachjam/js/components/FindDocuments/SearchResult.vue
```diff
@@ -74,6 +88,13 @@ export default {
       default: false,
     }
   },
+  data (){
+    const documentLabels = JSON.parse(document.querySelector('#data-labels').textContent).documentLabels;
```
longhotsummer, line 92: this will now be done once for every rendered search result, which isn't great. instead, do it at the top level of the main FindDocuments/index.vue component, and then push the documentLabels as state into these components.

peachjam/models/core_document_model.py
```diff
@@ -505,10 +530,14 @@ def pre_save(self):
...
-        # give ourselves and subclasses a chance to pre-populate derived fields before saving, in case full_clean() has
-        # not yet been called
-        self.pre_save()
-        return super().save(*args, **kwargs)
+        try:
+            # give ourselves and subclasses a chance to pre-populate derived fields before saving,
+            # in case full_clean() has not yet been called
+            self.pre_save()
+            return super().save(*args, **kwargs)
+        finally:
```
longhotsummer, line 538: i don't think you want the try-finally block -- if pre_save() or save() fails, it won't help to call `apply_labels()`. Instead, stash the result of `super().save()` (assuming there is one) and then return it after calling `apply_labels`.

peachjam/models/judgment.py
```diff
@@ -284,6 +284,21 @@ def assign_title(self):
...
+            code="reported",
+            defaults={"name": "Reported", "code": "reported"},
+        )
+
+        # if the judgment has alternative_names, apply the "reported" label
+        if self.alternative_names.exists():
+            self.labels.add(label.pk)
+        # if the judgment no alternative_names, remove the "reported" label
+        else:
+            self.labels.remove(label.pk)
```
longhotsummer, line 300: you must call super.apply_labels()

peachjam/tasks.py
```diff
@@ -114,6 +114,27 @@ def extract_citations(document_id):
     log.info("Citations extracted")
 
 
+@background(queue="peachjam", remove_existing_tasks=True)
+def apply_labels(document_id):
```
longhotsummer, line 118: this falls away because it's unused

peachjam_search/documents.py
```diff
@@ -229,7 +229,7 @@ def prepare_court(self, instance):
 
     def prepare_registry(self, instance):
         if hasattr(instance, "registry") and instance.registry:
-            return instance.registry.name
+            return instance.registry.namecourt
```
longhotsummer, line 232: ?

peachjam_search/documents.py
```diff
@@ -265,6 +265,9 @@ def prepare_taxonomies(self, instance):
         ]
         return list({t.slug for t in topics})
 
+    def prepare_labels(self, instance):
+        return [label.level for label in instance.labels.all()]
```
longhotsummer, line 269:
```suggestion
        return [label.code for label in instance.labels.all()]
```

peachjam/models/core_document_model.py
```diff
@@ -37,6 +38,29 @@
 from peachjam.storage import DynamicStorageFileField
 
 
+class Label(models.Model):
+    name = models.CharField(
+        _("name"), max_length=1024, unique=True, null=False, blank=False
+    )
+    code = models.SlugField(_("code"), max_length=1024, unique=True)
+    level = models.CharField(
+        _("level"), max_length=1024, null=False, blank=False, default="info"
```
longhotsummer, line 47: Add a `help_text="Level for styling, one of: primary, secondary, success, danger, warning or info."`

peachjam/js/components/FindDocuments/SearchResult.vue
```diff
@@ -52,6 +52,15 @@
         v-html="highlights(item)"
       />
     </div>
+      <span v-if="documentLabels.length">
+        <span v-for="(l, idx) in item.labels" :key="idx">
+          <span v-for="(label, index) in documentLabels" :key="index">
+            <span v-if="item.labels.includes(label.level)">
```
longhotsummer, line 58:
```suggestion
            <span v-if="item.labels.includes(label.code)">
```

longhotsummer (dismissed):

longhotsummer (approved):

PR #1417, opened 2023-07-31
search penalty

longhotsummer:
This allows us to demote certain documents in search results. The penalty is simply subtracted from the score.

This will allow us to push amendment legislation lower in results than principal legislation.

Disabled until the new field is populated.

### DIFF
--- a/peachjam/models/core_document_model.py
+++ b/peachjam/models/core_document_model.py
@@ -679,6 +679,11 @@ class CoreDocument(PolymorphicModel):
 
         return work_frbr_uris
 
+    def search_penalty(self):
+        """Optionally provide a penalty for this document in search results. This cannot be zero or None."""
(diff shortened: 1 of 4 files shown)


peachjam/models/core_document_model.py
```diff
@@ -679,6 +679,11 @@ def get_cited_work_frbr_uris(self):
 
         return work_frbr_uris
 
+    def search_penalty(self):
+        """Optionally provide a penalty for this document in search results. This cannot be zero or None."""
+        # provide a very small number instead of zero
+        return 0.0000001
+
```
actlikewill, line 687: Do we not need a database field?

actlikewill: It works okay without it, so I guess not lol.

actlikewill (approved):

PR #1420, opened 2023-08-01
Handle case numbers without number and year

nickmwangemi:
image


peachjam/models/judgment.py
```diff
@@ -349,7 +349,34 @@ def __str__(self):
     def get_case_number_string(self):
         if self.string_override:
             return self.string_override
-        return f"{self.matter_type or ''} {self.number} of {self.year}".strip()
+
+        parts = []
+
+        # Appeal 5 of 2018 (matter_type number year)
+        if self.matter_type and self.number and self.year:
```
longhotsummer, line 356: You're adding partially built components, and having to handle all the permutations. I suggest simply adding them one-by-one, which will handle things better:

```python
if self.matter_type:
  parts.append(self.matter_type)

if self.number:
  parts.append(self.number)
...
if self.year:
  parts.append(f'of {self.year}' if parts else self.year)

return ' '.join(parts)

PR #1422, opened 2023-08-04
Formalise simple public API

longhotsummer:
This formalises a basic portion of a public API. The API is read-only and is documented with openapi 3 / swagger.

The API has two endpoints:

* `/api/v1/gazettes` for fetching gazettes
* `/api/v1/judgments` for fetching judgments

In both cases, the API can provide PDF (if available) and plain text (if available) source documents.

Users need to both authenticate and have a special permission to access each content type via the API.

Note that the API doesn't use FRBR URIs to identify documents. I chose not to do that (yet) to avoid complexity.

Swagger-UI:
image

Redoc:
image


actlikewill (approved):
Great. I haven't used django-spectacular before, its good.

nickmwangemi (approved):

PR #1423, opened 2023-08-04
working prototype of popover

longhotsummer:
https://www.loom.com/share/036bc1ad76ca40198dfc86ac4ec631b8

We'd probably want to turn the button into a template that is called with the appropriate text from other templates.


longhotsummer:
Replaced by #1430

PR #1424, opened 2023-08-05
Preview dates for ratification bulk import

### LINKED ISSUE #1381: Preview of the ratifications importer does not display correctly the data to be imported
image

vs

image

actlikewill:
- We need to add the attribute value to show the values in the preview :roll_eyes: 

image

closes #1381

### DIFF
--- a/africanlii/resources.py
+++ b/africanlii/resources.py
@@ -3,41 +3,74 @@ from import_export import fields, resources, widgets
 
 from africanlii.models import Ratification, RatificationCountry
 from peachjam.models import Work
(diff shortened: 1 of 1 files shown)


actlikewill:
@longhotsummer The initial issue was that the "attribute" value was missing, so it wasn't showing values in the preview. However, I've gone with your suggestion of using the `RatificationCountry` model for the resource since that is the actual table whose rows are mostly affected, so it's a bit cleaner.

longhotsummer (approved):

PR #1425, opened 2023-08-06
Indigo adapter can handle multiple place codes; bg tasks

longhotsummer:
- indigo adapter can fetch for multiple place codes, simplifying cases where we need to fetch content for multiple places from the same API
- ingestors run as separate background tasks so that they fail and refresh independently

This will require existing ingestor adapters to be updated:

* ensure `api_url` is `https://api.laws.africa/v2`
* remove `url`
* ensure `places` is the correct list of place codes

### DIFF
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -1,6 +1,7 @@
 import logging
 import re
 from datetime import datetime
+from functools import cached_property, reduce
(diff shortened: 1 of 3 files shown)


actlikewill (approved):
Works great. Of course we will have to update all ingestors to use place code now instead of jurisdiction.

PR #1427, opened 2023-08-07
Convert source files to pdf

### LINKED ISSUE #1224: Performance: pro-actively convert documents to PDF
When a crawler is indexing our sites, we get a lot of requests to convert DOCX files to PDF files.

* this puts heavy load on the server
* it repeats work unnecessarily
* in some cases soffice hangs and pegs the CPU for 30 minutes before the task is aborted

It would  be better if we pro-actively created a PDF when a document is first uploaded, so that we do it once and can manage the impact of long-running tasks that fail more easily.

actlikewill:
- Proactively convert source files to pdf

closes #1224


actlikewill:
@longhotsummer This is mostly working, however, when the we use the DynamicStorageFileField in `file_as_pdf` I get maximum recursion error. Do you have an insight as to why?

actlikewill:
@longhotsummer It seems that [here](https://github.com/laws-africa/peachjam/blob/bd975a96e89a8eced50269d2f2af0b87f7d740fd/peachjam/storage.py#L35-L36) we are asserting that the name is always present, but since `file_as_pdf` is `None` when we first save the  model, the infinite recursion happens. It needs to be able to allow null values.

actlikewill:
@longhotsummer btw, do we need to have the file_as_pdf use the DynamicStorageFileField?

longhotsummer:
> @longhotsummer btw, do we need to have the file_as_pdf use the DynamicStorageFileField?

No, just use the normal file field.

peachjam/admin.py
```diff
@@ -167,7 +167,11 @@ def attachment_link(self, obj):
 class SourceFileInline(BaseAttachmentFileInline):
     model = SourceFile
     form = SourceFileForm
-    readonly_fields = (*BaseAttachmentFileInline.readonly_fields, "source_url")
+    readonly_fields = (
```
longhotsummer, line 170: Do you have a screenshot of what this looks like? I'm wondering if it's just too much detail and we can simply not have it at all.

peachjam/models/core_document_model.py
```diff
@@ -756,17 +756,37 @@ class SourceFile(AttachmentAbstractModel):
...
         suffix = os.path.splitext(self.filename)[1].replace(".", "")
-        return soffice_convert(self.file, suffix, "pdf")[0]
+        pdf = soffice_convert(self.file, suffix, "pdf")[0]
+        self.file_as_pdf = File(pdf, name=f"{self.file.name[:-5]}.pdf")
+        self.save()
+
+    def ensure_file_as_pdf(self):
+        from peachjam.tasks import convert_source_file_to_pdf
+
+        if self.mimetype != "application/pdf":
```
longhotsummer, line 788: Should you also check if `self.file_as_pdf` is `None` and not do anything if so?

peachjam/signals.py
```diff
@@ -75,3 +75,11 @@ def doc_deleted_update_extracted_citations(sender, instance, **kwargs):
     """Update language list on related work after a subclass of CoreDocument is deleted."""
     if isinstance(instance, CoreDocument):
         update_extracted_citations_for_a_work(instance.work_id)
+
+
+@receiver(signals.post_save)
```
longhotsummer, line 80: use `sender=SourceFile`

peachjam/forms.py
```diff
@@ -181,6 +181,11 @@ def _save_m2m(self):
             if self.instance.document.extract_content_from_source_file():
                 self.instance.document.save()
 
+                # if the file is changed, we need delete the existing pdf and re-generate
+                self.instance.file_as_pdf = None
```
longhotsummer, line 185: perhaps `self.instance.file_as_pdf.delete()` which also deletes the file from storage (and saves the model by default).

peachjam/models/core_document_model.py
```diff
@@ -756,17 +756,37 @@ class SourceFile(AttachmentAbstractModel):
...
         if self.filename.endswith(".pdf"):
             return self.file
+        elif self.file_as_pdf:
+            return self.file_as_pdf
+        else:
+            return None
 
-        # convert with soffice
+    def convert_to_pdf(self):
         suffix = os.path.splitext(self.filename)[1].replace(".", "")
```
longhotsummer, line 780: i think this should also check if it has to do this work.

Based on the spec:

> The background task should first check if the source file already has a PDF version, and do nothing if it does (to make the task idempotent).

It's also a good approach for async tasks: the point of this method isn't to blindly create the pdf, it's to ensure that a PDF exists. If, for some reason, it already exists, it can stop.

peachjam/admin.py
actlikewill: It may be too much detail. We can exclude the field.

peachjam/signals.py
actlikewill: Changed :+1:

peachjam/forms.py
actlikewill: Changed. Haven't used `file.delete` before.

peachjam/models/core_document_model.py
actlikewill: Makes sense. Added.

peachjam/signals.py
longhotsummer: See https://docs.djangoproject.com/en/3.2/topics/signals/#connecting-to-signals-sent-by-specific-senders

actlikewill: Much better. Thanks

longhotsummer (dismissed):

longhotsummer:
Let's coordinate the deployment of this on monday, to backfill PDF generation.

peachjam/admin.py
```diff
@@ -516,6 +517,15 @@ def apply_labels(self, request, queryset):
 
     apply_labels.short_description = "Apply labels"
 
+    def ensure_source_file_pdf(self, request, queryset):
+        count = queryset.count()
+        for doc in queryset:
```
longhotsummer, line 522:
```suggestion
        for doc in queryset.iterator():
```

longhotsummer: Can you apply the same change to the similar for loops above?

peachjam/admin.py
```diff
@@ -516,6 +517,15 @@ def apply_labels(self, request, queryset):
...
+            if hasattr(doc, "source_file"):
+                doc.source_file.ensure_file_as_pdf()
+        self.message_user(request, f"Ensuring PDF for {count} documents.")
+
+    ensure_source_file_pdf.short_description = "Ensure PDF for source file"
```
longhotsummer, line 527:
```suggestion
    ensure_source_file_pdf.short_description = "Ensure PDF for source file (background)"
```

longhotsummer:
@actlikewill I had to modify this slightly to handle no pdf.

actlikewill:
@longhotsummer looks good

PR #1428, opened 2023-08-09
Tweaks

longhotsummer:
* fix judges in judgment table (I broke this a while ago :( )
* cache request language, this gets called a lot (reduces SQL queries)
* move labels below document title

image

image

### DIFF
--- a/peachjam/templates/peachjam/_judgment_table.html
+++ b/peachjam/templates/peachjam/_judgment_table.html
@@ -4,7 +4,6 @@
     <thead>
       <tr>
         <th scope="col" style="width: 70%">{% trans 'Title' %}</th>
-        <th scope="col">{% trans 'Judges' %}</th>
         <th scope="col">{% trans 'Date' %}</th>
       </tr>
     </thead>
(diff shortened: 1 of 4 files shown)


peachjam/templates/peachjam/_judgment_table.html
```diff
@@ -13,8 +13,15 @@
...
             <a href="{{ document.get_absolute_url }}">{{ document.title }}</a>
+            {% include 'peachjam/_labels.html' with labels=document.labels.all %}
+          </td>
+          <td>
+            {% for bench in document.bench.all %}
+              {# djlint:off #}
+              {{ bench.judge.name }}{% if not forloop.last %},{% endif %}
+              {# djlint:on #}
+            {% endfor %}
           </td>
```
actlikewill, line 24: @longhotsummer I thought we removed these because some judgments have many judges which makes the table look a bit odd. Maybe we could limit the list to 3 or 4?

longhotsummer: Hmm, fair point. It's useful information for users to have, particularly those who know the judges.

Let's think it through in more detail, because a long list on mobile is particularly bad. Maybe we can do something like the first 3/4 and have "... and 3 more" or something.

nickmwangemi: I agree. We'll need a better way of displaying long lists.

longhotsummer: I've removed this change based on feedback from the team that the judges aren't actually useful in the listing.

longhotsummer:
I've removed the judges column in the judgment table because (a) it's complex to get right on both wide and narrow displays, and (b) it's not critical information.

actlikewill (approved):
Looks good.

PR #1430, opened 2023-08-09
Help links 

### LINKED ISSUE #1414: As a user I want to see help on how to use a LII website
How do we integrate the LII User Guide into the website?

actlikewill:
This adds a help popover button at various places on the site.
The popover should typically contain a link to the help pages hosted on Gitbook.
The screenshots below show a few places where the help button is useful, with the expectation that other places where the button can go will be identified in the future.

####  Search

image

#### Advanced Search

image

#### Legislation Page

image

####  Related Documents

image

#### Citations

image

closes #1414


peachjam/templates/peachjam/_help_popover.html
```diff
@@ -0,0 +1,7 @@
+<div class="mb-3">
+  <a href="#{{ link }}"
```
actlikewill, line 2: @longhotsummer here is the `#` referenced in the above comment

liiweb/views/legislation.py
```diff
@@ -46,6 +47,13 @@ def get_context_data(self, **kwargs):
 
         context["legislation_table"] = LegislationSerializer(qs, many=True).data
 
+        context["help_button_text"] = _("Help")
```
longhotsummer, line 50: This doesn't change, so just hardcode it into the template I think.

peachjam/templates/peachjam/_help_popover.html
longhotsummer: The page that embeds this template can provide the path to the page eg `/judgments/courts` and then we can hardcode the domain.

peachjam/js/peachjam.ts
```diff
@@ -137,6 +138,20 @@ class PeachJam {
...
   }
 
+  setupPopovers () {
+    document.querySelectorAll('[data-bs-toggle="help-popover"]').forEach((el) => {
+      // @ts-ignore
+      new window.bootstrap.Popover(el, {
+        html: true,
+        content: `
+        ${el.getAttribute('data-bs-content')}
+         <div><a href="${el.getAttribute('href')?.split('#')[1] || '#'}" target='_blank' rel='noopener noreferrer'>Learn more</a></div>
```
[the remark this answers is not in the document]
longhotsummer: The goal is to open the docs page in a new browser tab, and the popup can close. So I think the default behaviour is sufficient, unless I'm missing something.

peachjam/js/peachjam.ts
```diff
@@ -137,6 +138,32 @@ class PeachJam {
...
+      const helpPopover = new window.bootstrap.Popover(el, {
+        html: true,
+        content: `
+        ${el.getAttribute('data-bs-content')}
+         <div><a href="${el.getAttribute('href')?.split('#')[1] || '#'}" target='_blank' rel='noopener noreferrer'>Learn more</a></div>
+        `,
+        container: 'body'
+      });
+      // keep popover open when clicking inside it but close when clicking outside it
+      el.addEventListener('inserted.bs.popover', (e) => {
```
longhotsummer, line 153: I don't think we need this -- if the user clicks the link in the popover, the docs open in a new tab and the popover can close? Or am I confused?

longhotsummer:
This has some good basics. The team will work on documentation over the next 10 days or so, and we can start pulling in the details here then.

peachjam/js/peachjam.ts
actlikewill: @longhotsummer When the popover trigger is set to `click`, the popover can only be closed by clicking the button again. This behaviour is annoying because the popover moves about if you scroll the page without clicking it off. Its worse when you have multiple popovers open. They move about if they are not clicked closed. When  the trigger set to `focus`, the pop over disappears when you click outside but  does not capture the click event when you click on `Learn more`  i.e. it doesn't navigate away. This allows the popover to stay in place when you click on it so you can click on the link, and it disappears when you click outside it.

peachjam/js/peachjam.ts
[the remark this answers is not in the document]
actlikewill: @longhotsummer With the way it was, clicking on the help button triggered the browser to navigate away immediately because its technically an <a> anchor tag with a href attribute

longhotsummer:
@actlikewill  https://www.loom.com/share/4edb71a91d2444f4ba76228b75f969ef

I've updated this to simplify the button, remove the effort required to add a link to the page (and the translations required), and linked to the documentation that we have.

longhotsummer (approved):

PR #1432, opened 2023-08-10
By-laws on LawLibrary

nickmwangemi:


open_by_laws/views/by_laws.py
```diff
@@ -1,5 +1,16 @@
...
 from liiweb.views.legislation import LocalityLegislationListView
+from peachjam.models import Locality
 
 
 class MunicipalByLawsView(LocalityLegislationListView):
     template_name = "open_by_laws/municipal_by_laws_list.html"
+    navbar_link = "legislation/municipal"
+
+    def get(self, *args, **kwargs):
+        self.municipality = get_object_or_404(Locality, code=kwargs["code"])
```
longhotsummer, line 12: Just a reminder that this is exactly what LocalityLegislationListView does, using the word `locality` rather than `municipality`. So this isn't necessary.

liiweb/views/legislation.py
```diff
@@ -73,16 +74,11 @@ class LocalityLegislationView(TemplateView):
...
-        context = super().get_context_data(**kwargs)
-        localities = Locality.objects.all()
-        context["province_groups"] = list(chunks(localities, 2))
-        return context
-
 
 class LocalityLegislationListView(LegislationListView):
     template_name = "liiweb/locality_legislation_list.html"
     navbar_link = "legislation/locality"
+    provincial_codes = PROVINCIAL_CODES
```
longhotsummer, line 81: This can't be done in `liiweb`, because provinces are a lawlibrary thing only.

liiweb/views/legislation.py
```diff
@@ -73,16 +74,11 @@ class LocalityLegislationView(TemplateView):
     template_name = "liiweb/locality_legislation.html"
     navbar_link = "legislation/locality"
 
-    def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
-        localities = Locality.objects.all()
-        context["province_groups"] = list(chunks(localities, 2))
```
longhotsummer, line 79: This needs to stay here, but rename `province_groups` to `locality_groups`.

liiweb/templates/liiweb/_legislation_nav.html
```diff
@@ -0,0 +1,36 @@
+{% if locality_type == "province" %}
+  <nav class="nav nav-tabs mb-3 border-bottom">
```
longhotsummer, line 2: The two `if` blocks are identical except for whether subleg is included. Rather remove the if block, and then pass in a flag from the View object to indicate whether the subleg should be shown.

lawlibrary/templates/peachjam/legislation_detail.html
```diff
@@ -3,14 +3,25 @@
...
     {% if document.locality %}
-      <ol class="breadcrumb">
-        <li class="breadcrumb-item">
-          <a href="{% url 'locality_legislation' %}">{% trans 'Provincial Legislation' %}</a>
-        </li>
-        <li class="breadcrumb-item">
-          <a href="{% url 'locality_legislation_list' document.locality.code %}">{{ document.locality.name }}</a>
-        </li>
-      </ol>
+      {% if document.locality.code in PROVINCIAL_CODES %}
```
longhotsummer, line 6: Only the first part of the breadcrumb changes, the rest is the same. So put the if block only around the part that changes, otherwise you're duplicating code.

PR #1433, opened 2023-08-10
Sandrava 1336

Sandravaphilips:


peachjam/js/components/clipboard.ts
```diff
@@ -9,7 +9,7 @@ export class CopyToClipboard {
   }
 
   copy () {
-    navigator.clipboard.writeText(this.root.dataset.value || '')
+    navigator?.clipboard?.writeText(this.root.dataset.value || '')
```
longhotsummer, line 12: the `.then` will still fail. Instead, do `if (navigator.clipboard && navigator.clipboard.writeText)` or similar.

longhotsummer (changes requested):
the `.?` doesn't work cleanly, so it needs to be more explicit.

longhotsummer (approved):

PR #1444, opened 2023-08-15
Change headnote holding label to case summary

nickmwangemi:
image


peachjam/migrations/0097_auto_20230815_0818.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("peachjam", "0096_auto_20230814_1557"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
longhotsummer, line 13: This needs to be a rename operation, otherwise we'll lose the existing data.

longhotsummer: Django should give you that option when you create the migration -- it'll ask if you've renamed the field. Or just change this migration to be a rename.

longhotsummer:
Does the bulk importer need to change?

longhotsummer (approved):

PR #1451, opened 2023-08-17
better error handling during import/export

longhotsummer:
* make pdfjs-to-text not fail when the PDF is dodgy
* if an import fails, then show errors; this isn't perfect, because we can't easily override the template, but it's better than nothing and is an edge case

https://www.loom.com/share/6c4b4b1d601f48ca83f8ceebec794531


actlikewill (approved):
Great, a good improvement

Issue #1413, opened 2023-07-31
As a reader I want to be able to share a link to a particular portion of a document


longhotsummer:
If I'm reading a long document in particular, perhaps a PDF or perhaps a long piece of legislation, I want to be able to share a link just to a particular portion. Sharing should be through whatsapp, twitter, etc.

* what do I click on to share it? in the Table of Contents? in the document itself?
* what does the URL look like?

This is required for the law reader project.

longhotsummer:
This in largely done in #1453.

We also have the suggestion to add a "share this text" popup when a user selects text, so I'm keeping this open.

PR #1453, opened 2023-08-17
Sharing URLs take current page portion into account

longhotsummer:
When sharing a PDF, we'll share "Title - Page 2 - URL#page-2"

When sharing a portion of an HTML document, we'll share "Title - Section 32 - URL#sec_32"

This is required for the law reader project to ensure that users can easily share the current portion/page of the document they are reading.

About 77% of shares through our share buttons are for whatsapp.

### DIFF
--- a/peachjam/js/components/DocumentContent/index.ts
+++ b/peachjam/js/components/DocumentContent/index.ts
@@ -43,6 +43,7 @@ class DocumentContent {
     this.setupPdf();
     this.setupSearch();
     this.setupEnrichments();
+    this.setSharedPortion();
     if (this.root.getAttribute('data-display-type') !== 'pdf') {
       // for PDFs, this will be done once the pages are rendered
       this.setupPopups();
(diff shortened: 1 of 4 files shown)


longhotsummer:
Closes #1413

actlikewill (approved):
Works well. Maybe we could add the share buttons to the search popup that appears when you highlight the text of a document? Or add a share button that is fixed even when scrolling that picks up the portion that a user is looking at?

nickmwangemi (approved):

longhotsummer:
> Works well. Maybe we could add the share buttons to the search popup that appears when you highlight the text of a document? Or add a share button that is fixed even when scrolling that picks up the portion that a user is looking at?

Sharing based on selected text is a great idea, especially for desktop. I'm surprised that so many users use the share buttons on the page on mobile, when they can easily just share using the built-in browser sharing on mobile.

I haven't yet worked out how to add a share button on mobile that scrolls with the reader, that doesn't get in the way.

Issue #1461, opened 2023-08-21
As an admin, I want to be able to allow a legislation listing page to have an entity profile


longhotsummer:
This will allow admins to add partner logos and attribution on legislation listing pages.

This is required for LesothoLII.

## Proposal

* For `/legislation/<code>` (locality legislation listing page), load and render the entity profile for the locality
* For `/legislation`, we need to know which country to load an entity profile for. The simplest is to check if `peachjam_settings.document_jurisdictions` has a single country. If it does, load and render the locality for that jurisdiction.

This means for sites that cover multiple countries (only africanlii.org), we won't show an entity profile for the legislation page. That's fine for now, since AfricanLII.org doesn't have that page.

nickmwangemi:
@longhotsummer Since a country may hold multiple localities, how about we introduce a new model to hold a country's profile?

longhotsummer:
I think we'll need to, because there is no other way to associate an EntityProfile with a Country object, because the Country model is from countries_plus, right?

nickmwangemi:
> I think we'll need to, because there is no other way to associate an EntityProfile with a Country object, because the Country model is from countries_plus, right?

Yes, it's from the 3rd party package

PR #1477, opened 2023-08-25
locality legislation view uses entity profiles from locality

longhotsummer:
This enables us to have a richer experience on legislation views, similar to what we do on court and taxonomy detail views.

* lays the groundwork for #1461 to support profiles at the top-level legislation level (this only covers locality legislation)
* also fixes how we find entity profiles for taxonomy topics
* ref #1461

image

### DIFF
--- a/peachjam/models/core_document_model.py
+++ b/peachjam/models/core_document_model.py
@@ -9,6 +9,7 @@ from cobalt.akn import StructuredDocument, datestring
 from cobalt.uri import FrbrUri
 from countries_plus.models import Country
 from django.conf import settings
+from django.contrib.contenttypes.fields import GenericRelation
 from django.contrib.postgres.fields import ArrayField
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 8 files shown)


peachjam/models/core_document_model.py
```diff
@@ -87,6 +88,9 @@ class Locality(models.Model):
         Country, on_delete=models.PROTECT, verbose_name=_("jurisdiction")
     )
     code = models.CharField(_("code"), max_length=20, null=False)
+    entity_profile = GenericRelation(
+        "peachjam.EntityProfile", verbose_name=_("profile")
+    )
```
actlikewill, line 93: interesting that this works without a migration

longhotsummer: yeah, my thoughts too. I think it sees that there is a GenericForeignKey on the other model.

actlikewill (approved):

PR #1480, opened 2023-08-25
Judgment admin form tweaks

nickmwangemi:
https://www.loom.com/share/215cd90abd314207a691ef5cd0a50a7c?sid=baabc6eb-535f-4ab5-b733-0f57b38e62c9
 
Pending is addition of a new permission on the `CoreDocument` model to toggle the display of the `Work identification` and `Advanced` tabs.


peachjam/models/core_document_model.py
```diff
@@ -848,14 +849,19 @@ class AlternativeName(models.Model):
         related_name="alternative_names",
         verbose_name=_("document"),
     )
-    title = models.CharField(_("title"), max_length=1024, null=False, blank=False)
+    law_report_citation_or_alternative_known_name = models.CharField(
```
longhotsummer, line 852: Changing the actual field name is too big a change. The goal here is simply to change the label visible to the user. The field name of `title` must remain, but the label (the `_("law report...")`) is the required change.

peachjam/models/judgment.py
```diff
@@ -343,12 +340,12 @@ def pre_save(self):
 
 
 class CaseNumber(models.Model):
-    string_override = models.CharField(
-        _("string override"),
+    full_case_number_as_printed_on_judgment = models.CharField(
```
longhotsummer, line 343: As above, the `string_override` must remain, just the label part of the definition `_("full case number...")` must change.

peachjam/models/judgment.py
```diff
@@ -271,7 +268,7 @@ def generate_citation(self):
         )
 
     def generate_work_frbr_uri(self):
-        # enforce certain defaults for judgment FRBR URIs
+        # enforce certain defaults for judgment JudgmFRBR URIs
```
longhotsummer, line 271:
```suggestion
        # enforce certain defaults for judgment FRBR URIs
```

longhotsummer (dismissed):
Looks great.

The `default_document_jurisdiction` will need to be filled in manually on the LIIs. An alternative is to include a python migration as part of adding the new field which populates it with the country in the `document_jurisdictions` setting, if there is only one in that list (otherwise don't choose a default).

longhotsummer (approved):

PR #1483, opened 2023-08-28
icons

longhotsummer:
Introduce some icons to the document detail page. This is to add some visual variance and also help guide the user to features they may not know about.

I've add `.bi-pj-...` so that we can use semantic icon names in peachjam, rather than bootstrap-icon's names.


nickmwangemi:
Nice!

nickmwangemi (approved):

PR #1487, opened 2023-08-28
TypeError: Cannot read property 'then' of undefined

Sandravaphilips:


peachjam/js/components/clipboard.ts
```diff
@@ -9,7 +9,7 @@ export class CopyToClipboard {
   }
 
   copy () {
-    if (navigator && navigator.clipboard) {
+    if (navigator && navigator.clipboard && navigator.clipboard.writeText) {
```
longhotsummer, line 12: I think what's actually happening is that writeText() is returning null :/ -- so you'll need to check the response first.

This change won't hurt either.

This is like JS programming back in the early 2000s. Total wild west, you never knew what different browsers would do.

longhotsummer (approved):

longhotsummer:
Closes #1460

PR #1491, opened 2023-08-30
fixes doc mimetype check

actlikewill:
- check if source file pk is set
- check if pdf is available

### DIFF
--- a/peachjam/models/core_document_model.py
+++ b/peachjam/models/core_document_model.py
@@ -911,16 +911,18 @@ class DocumentContent(models.Model):
             text = " ".join(root.itertext())
 
         elif hasattr(document, "source_file"):
-            # get the text from the source file, via PDF if necessary
-            with tempfile.NamedTemporaryFile() as tmp:
-                # convert document to pdf and then extract the text
-                pdf = document.source_file.as_pdf()
-                shutil.copyfileobj(pdf, tmp)
-                tmp.flush()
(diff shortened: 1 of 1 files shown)


peachjam/models/core_document_model.py
```diff
@@ -911,16 +911,18 @@ def update_or_create_for_document(cls, document):
...
-                # convert document to pdf and then extract the text
-                pdf = document.source_file.as_pdf()
-                shutil.copyfileobj(pdf, tmp)
-                tmp.flush()
-                text = pdfjs_to_text(tmp.name)
-                # some PDFs have nulls, which breaks SQL insertion
-                # replace rather than deleting to keep string length the same
-                text = text.replace("\0", " ")
+            if document.source_file.pk:
+                # get the text from the source file, via PDF if necessary
```
actlikewill, line 915: during an update the _savem2m sometimes get called before saving the source file, i think when saving other models?

peachjam/models/core_document_model.py
```diff
@@ -911,16 +911,18 @@ def update_or_create_for_document(cls, document):
...
+                with tempfile.NamedTemporaryFile() as tmp:
+                    # convert document to pdf and then extract the text
+                    pdf = document.source_file.as_pdf()
+                    if pdf:
```
actlikewill, line 919: some pdfs have strange mimetypes like application/octet-stream

longhotsummer (approved):

PR #1492, opened 2023-08-30
Restrict Work identification and Advanced to advanced users 

nickmwangemi:
https://www.loom.com/share/ccba292df03b44f2b646e55f0bd642a2?sid=ff16657f-acd9-4fe4-b0df-f2e4ddf33bc4


peachjam/admin.py
```diff
@@ -437,7 +437,22 @@ def get_queryset(self, request):
     def get_fieldsets(self, request, obj=None):
         fieldsets = super().get_fieldsets(request, obj)
         if obj is None:
-            fieldsets = self.new_document_form_mixin.adjust_fieldsets(fieldsets)
+            if request.user.has_perm("peachjam.can_edit_advanced_fields"):
```
longhotsummer, line 440: I think we always need to call `adjust_fieldsets`. The manipulation of the fieldsets based on the user permission should happen whether or not `obj` is None.

So these two things seem unrelated to me, meaning you can move your changes to below the if block, and always do it.

peachjam/admin.py
```diff
@@ -437,7 +437,22 @@ def get_queryset(self, request):
...
+                restricted_fieldsets = [
+                    (x[0], x[1])
+                    for x in fieldsets
+                    if x[0]
+                    not in [
+                        gettext_lazy("Advanced"),
+                        gettext_lazy("Work identification"),
+                    ]
+                ]
+                fieldsets = restricted_fieldsets
```
longhotsummer, line 454: no need for the spare variable:

```suggestion
                fieldsets = [
                    x for x in fieldsets
                    if x[0] not in [
                        gettext_lazy("Advanced"),
                        gettext_lazy("Work identification"),
                    ]
                ]
```

longhotsummer (approved):
This looks good. Please just double check it works when using the site in a non-English language. I'm just not 100% how `gettext_lazy(...)` will compare with each other.

nickmwangemi:
> This looks good. Please just double check it works when using the site in a non-English language. I'm just not 100% how `gettext_lazy(...)` will compare with each other.

I've tested it

https://www.loom.com/share/f71c426068cb4bd38da5d33dfd5bcf6b?sid=3b964c3a-0546-49d9-bc38-a6bbff0c6c8c

PR #1505, opened 2023-09-04
Remove exising relationships in adapter

### LINKED ISSUE #1494: The indigo adapter must clean up relationships while importing
Suppose we have two works:

* Principal
* Commencement -- which commences Principal

When the Indigo adapter ingests Principal, it creates a relationship between Principal and Commencement. Note that when Commencement is ingested (which could happen before or after), no relationships change. Relationships are created by one side only.

Let's suppose Commencement didn't actually commence Principal. So it is corrected in Indigo and Principal is re-fetched. The old relationship is not removed, and so is still in place, even though it is incorrect.

## Proposal

The Indigo Adapter explicitly works with certain relationships, and it knows which direction they go in. When ingesting a document, it should delete any existing relationships (that it is responsible for) before creating new ones.

In other words, because it knows that Principal should have specific "commenced-by" relationships, it can delete all existing "commenced-by" relationships on Principal before ingesting the new ones. It *cannot* delete *all* relationships, because other works may be responsible for managing those.

Once this is done, all indigo adapters must be refreshed for all sites.

actlikewill:
- removes existing relationships before ingesting

fixes #1494

### DIFF
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -321,10 +321,9 @@ class IndigoAdapter(Adapter):
 
     def remove_existing_relationships(self, subject_work):
         # delete any existing relationships first
-        predicates = [p for p in self.predicates.keys()]
         relationships = Relationship.objects.filter(
             subject_work=subject_work,
-            predicate__slug__in=predicates,
+            predicate__slug__in=list(self.predicates.keys()),
         )
         logger.info(f"Deleting {relationships.count()} relationships")
         relationships.delete()
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -319,8 +319,19 @@ class IndigoAdapter(Adapter):
             created_document.parent_work = None
         created_document.save()
 
+    def remove_existing_relationships(self, subject_work):
+        # delete any existing relationships first
+        predicates = [p for p in self.predicates.keys()]
+        relationships = Relationship.objects.filter(
+            subject_work=subject_work,
+            predicate__slug__in=predicates,
+        )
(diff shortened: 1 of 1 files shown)


actlikewill:
@longhotsummer we should not touch the related work object itself, should we?

longhotsummer:
> @longhotsummer we should not touch the related work object itself, should we?

No

peachjam/adapters/adapters.py
```diff
@@ -319,6 +320,15 @@ def set_parent(self, imported_document, created_document):
             created_document.parent_work = None
         created_document.save()
 
+    def remove_commencement_relationships(self, created_document):
```
longhotsummer, line 323: I suggest doing this as a part of `fetch_relationships`.

Also `repealed-by` and `amended-by` must also be cleaned up.

actlikewill: Ok, changed :+1:

peachjam/adapters/adapters.py
```diff
@@ -319,8 +319,19 @@ def set_parent(self, imported_document, created_document):
             created_document.parent_work = None
         created_document.save()
 
+    def remove_existing_relationships(self, subject_work):
+        # delete any existing relationships first
+        predicates = [p for p in self.predicates.keys()]
+        relationships = Relationship.objects.filter(
+            subject_work=subject_work,
+            predicate__slug__in=predicates,
```
longhotsummer, line 327:
```suggestion
        relationships = Relationship.objects.filter(
            subject_work=subject_work,
            predicate__slug__in=list(self.predicates.keys()),
```

longhotsummer (approved):

PR #1511, opened 2023-09-06
Timeline

### LINKED ISSUE #996: As a legislation reader I want full and partial commencements included in the timeline.
Currently we don't include commencement notes or partial commencements.

goose-life:
will close #996 

https://www.loom.com/share/42bceb64c6424d82b9c7e36aefe3f969


goose-life:
@actlikewill despite updating the sample documents, when I load e.g. http://127.0.0.1:8000/akn/za/act/1979/70/eng@2020-10-22 and inspect `self.object` in https://github.com/laws-africa/peachjam/blob/main/peachjam/views/legislation.py#L180 it show empty lists for both `timeline` and `commencements_json` on that object. am I missing something obvious? I'm pretty sure it's the right object, the pk is 3040.

goose-life:
also I have no idea what is being reformatted in those files?

goose-life:
> @actlikewill despite updating the sample documents, when I load e.g. http://127.0.0.1:8000/akn/za/act/1979/70/eng@2020-10-22 and inspect `self.object` in https://github.com/laws-africa/peachjam/blob/main/peachjam/views/legislation.py#L180 it show empty lists for both `timeline` and `commencements_json` on that object. am I missing something obvious? I'm pretty sure it's the right object, the pk is 3040.

it was something obvious — just needed to load the fixtures again lol

goose-life:
it's still saying
```
reformatted peachjam/migrations/0105_legislation_timeline.py
reformatted peachjam/migrations/0106_legislation_commencements_json.py
reformatted peachjam/models/generic_document.py
reformatted peachjam/views/legislation.py
```
but it doesn't say how those files were reformatted

longhotsummer:
A loom demo would be grand :)

actlikewill (approved):
Looks good, Will we need to switch the ingestor urls the sites?

goose-life:
> Looks good, Will we need to switch the ingestor urls the sites?

As discussed just now, for the record: v2 of the indigo API includes the /timeline already, so we can start using it now and switch over to v3 as a totally separate exercise. (I've opened a separate issue for that.)

goose-life:
However! Everything needs to be ingested again, otherwise the `timeline_json` on each existing Legislation object will just be a blank list. 

So tomorrow:
- trigger an update from edit.laws.africa
- verify that all works from that place now have a non-blank timeline on the live site
- repeat for all places (do we want to stagger this to avoid huge background task queues?)
- once all Legislation objects have non-empty timelines, deploy this

@actlikewill does this sound good?

goose-life:
https://www.loom.com/share/902ea08d6e3f429daf18f7da9757c0b9?sid=bb1b2d50-61dc-493b-ac9d-f6599c545478

longhotsummer (approved):
looks good, and the deployment strategy seems sane to me.

actlikewill (approved):
Works well.

PR #1512, opened 2023-09-07
Taxonomy topics from indigo adapter

### LINKED ISSUE #1475: Import taxonomy topic tags from Indigo
Use Indigo's new `taxonomy_topics` detail in the document API to import certain taxonomy topic tags for a document.

Since we don't want all topics imported, we should be able to specify, as an adapter setting, a list of root slugs. Only topics that start with those slugs should be imported. This should work under the assumption that the appropriate topics are already part of the local taxonomy tree.

The document may have other taxonomy topics applied locally, and those should not be disturbed. Additionally, we can't just add, we must also remove. So, first remove any existing topics on the document that are from that root, and delete them. Then add only the topics that are from the root.

actlikewill:
- Get taxonomy topics from indigo adapter
- clear existing taxonomy topics

closes #1475

### DIFF
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -293,26 +293,27 @@ class IndigoAdapter(Adapter):
             # the source file is the PDF version
             self.download_source_file(f"{url}.pdf", created_doc, title)
 
-        # clear any existing taxonomies
-        created_doc.taxonomies.filter(
-            topic__slug__startswith=self.taxonomy_topic_root
-        ).delete()
+        if self.taxonomy_topic_root:
+            # clear any existing taxonomies
+            created_doc.taxonomies.filter(
+                topic__slug__startswith=self.taxonomy_topic_root
+            ).delete()
 
-        if document["taxonomy_topics"] and self.taxonomy_topic_root:
-            # get topics beginning with "subject-areas"
-            topics = [
-                t
-                for t in document["taxonomy_topics"]
-                if t.startswith(self.taxonomy_topic_root)
-            ]
-            if topics:
-                taxonomies = Taxonomy.objects.filter(slug__in=topics)
-                for taxonomy in taxonomies:
-                    DocumentTopic.objects.create(
-                        document=created_doc,
-                        topic=taxonomy,
-                    )
-                logger.info(f"Added {len(taxonomies)} taxonomies to {created_doc}")
+            if document["taxonomy_topics"]:
+                # get topics beginning with "subject-areas"
+                topics = [
+                    t
+                    for t in document["taxonomy_topics"]
+                    if t.startswith(self.taxonomy_topic_root)
+                ]
+                if topics:
+                    taxonomies = Taxonomy.objects.filter(slug__in=topics)
+                    for taxonomy in taxonomies:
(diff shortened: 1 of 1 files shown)


peachjam/adapters/adapters.py
```diff
@@ -290,6 +292,21 @@ def update_document(self, url):
...
             self.download_source_file(f"{url}.pdf", created_doc, title)
 
+        if document["taxonomy_topics"]:
+            # get topics beginning with "subject-areas"
+            topics = [
+                t for t in document["taxonomy_topics"] if t.startswith("subject-areas")
+            ]
+            if topics:
+                taxonomies = Taxonomy.objects.filter(slug__in=topics)
+                created_doc.taxonomies.all().delete()
```
longhotsummer, line 302: Check the proposal -- you need to delete only those taxonomies that are part of `subject-areas` -- otherwise manually added taxonomies will be lost.

Basically, this adapter only handles taxonomies under a certain topic, and should only play within those walls.

actlikewill: Ah, right. Updated.

peachjam/adapters/adapters.py
```diff
@@ -290,6 +292,23 @@ def update_document(self, url):
...
+        if document["taxonomy_topics"]:
+            # get topics beginning with "subject-areas"
+            topics = [
+                t for t in document["taxonomy_topics"] if t.startswith("subject-areas")
+            ]
+            if topics:
+                taxonomies = Taxonomy.objects.filter(slug__in=topics)
+                created_doc.taxonomies.filter(
+                    topic__slug__startswith="subject-areas"
+                ).delete()
```
longhotsummer, line 304: What about the case where an existing document previously had topics under this tree, but no longer does?

actlikewill: updated :+1:

longhotsummer (commented):
Looks good, remaining is to make `subject-areas` an optional adapter setting.

actlikewill:
> Looks good, remaining is to make `subject-areas` an optional adapter setting.

@longhotsummer added

peachjam/adapters/adapters.py
```diff
@@ -225,7 +230,9 @@ def update_document(self, url):
             "date": datetime.strptime(document["expression_date"], "%Y-%m-%d").date(),
         }
         if document["locality"]:
-            frbr_uri_data["locality"] = Locality.objects.get(code=document["locality"])
+            frbr_uri_data["locality"] = Locality.objects.gapi_urlet(
```
longhotsummer, line 233: ??

peachjam/adapters/adapters.py
```diff
@@ -87,6 +89,9 @@ def __init__(self, settings):
             }
         )
         self.api_url = self.settings["api_url"]
+        self.taxonomy_topic_root = self.settings.get(
```
longhotsummer, line 92: Don't specify a default, otherwise we can't disable the functionality.

peachjam/adapters/adapters.py
actlikewill: yikes :grimacing: i need to get used to using git commit -p

longhotsummer (approved):

PR #1514, opened 2023-09-07
Brand Open Law Africa sites

nickmwangemi:
## AfricanLII
image
image

## LawLibrary
image
image

## TanzLII
image
image

## ULII
image
image

## GhaLII
image
image

## SeyLII
image
image


actlikewill:
@nickmwangemi I think you can have single image in the base peachjam static folder instead of separate ones in each lii

ulii/static/images/open-law-africa-footer.png
longhotsummer, line 1: just add the images once, in the `liiweb` project.

longhotsummer: actually, `peachjam` makes more sense.

longhotsummer (changes requested):
In general this looks good, but we need a few changes:

1. right-align the OLA logo in the headers
2. when right-aligned, the images shouldn't have `class="me-4"` because that puts margin on the right, which is not necessary
3. we'll need to do something on africanlii because on mobile the images are too wide.

The footer stuff looks good.

longhotsummer (changes requested):
I've adjusted widths on mobile and removed the need for specific heights.

The OLA logos are still duplicated between liiweb and peachjam. They can move into peachjam. I think we only need two: the normal and the inverted.

Please also check with @actlikewill how to reduce them down to a decent size -- 256KB is pretty large.

longhotsummer (approved):

PR #1523, opened 2023-09-08
Taxonomy n+1

actlikewill:
taxonomy n+1 queries

___Before___
image

___After___
image

### DIFF
--- a/peachjam/templates/peachjam/_taxonomy_list.html
+++ b/peachjam/templates/peachjam/_taxonomy_list.html
@@ -3,14 +3,14 @@
   {% if taxonomies|length > 0 %}
     <div class="row">
       {% for taxonomy in taxonomies %}
-        {% if taxonomy.is_root and taxonomy.get_children_count > 0 %}
+        {% if taxonomy.children|length > 0 %}
           <div class="col-12 col-lg-6 mb-4">
             <div class="card">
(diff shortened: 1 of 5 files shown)


peachjam/templates/peachjam/_taxonomy_list.html
```diff
@@ -3,14 +3,14 @@
...
           <div class="col-12 col-lg-6 mb-4">
             <div class="card">
-              <div class="card-header">{{ taxonomy }}</div>
+              <div class="card-header">{{ taxonomy.data.name }}</div>
               <ul class="list-group list-group-flush">
-                {% for child in taxonomy.get_children %}
+                {% for child in taxonomy.children %}
                   <li class="list-group-item">
-                    <a href="{% build_taxonomy_url child prefix=taxonomy_link_prefix|default:"taxonomy" %}">{{ child }}</a>
+                    <a href="{% url taxonomy_url taxonomy.data.slug child.data.slug %}">{{ child.data.name }}</a>
```
actlikewill, line 13: I have removed the build_taxonomy_url tag here because it has a call to taxonomy.get_root() which was compounding the n+1 queries. Since the dump bulk has the root as the first level in the loop,  this works okay.

longhotsummer (approved):

PR #1525, opened 2023-09-08
Check possible duplicates for judgment

### LINKED ISSUE #798: As a peachjam site editor, I would like to receive a warning if a judgment record I am adding is a potential duplicate of another existing judgment record
The fields that could be compared are the court name, case name, the case number and judgment date 
examples of duplicates from https://tanzlii.org/judgments/High-Court-Dar-es-Salaam-Main-Registry/2022
 
image

actlikewill:
Showing potential duplicates

https://www.loom.com/share/1103a2ea91604ed7a0522b0b25fb093b

closes #798


actlikewill:
@longhotsummer @nickmwangemi  Jazzmin is using select2 library for the dropdown fields. This was causing issues with using vanilla js. You can't listen for events and capture values using vanilla js because of the library. Select2 uses jquery so using it works okay.

peachjam_api/views.py
```diff
@@ -53,3 +55,72 @@ def post(self, request, ingestor_id):
...
+            if judgment_id:
+                qs = qs.exclude(id=judgment_id)
+            duplicate = qs.first()
+
+            if duplicate:
+                # reverse to admin change view
+                url = reverse("admin:peachjam_judgment_change", args=[duplicate.pk])
+                return Response(
+                    f'<div id="duplicate-alert" class="alert alert-danger" role="alert">Possible duplicate of '
+                    f'<a target="_blank" href="{url}">{duplicate}</a'
```
longhotsummer, line 123:
```suggestion
                    f'<a target="_blank" href="{url}">{duplicate}</a>'
```

longhotsummer: this string needs to be translateable

peachjam/templates/admin/change_form.html
```diff
@@ -7,3 +7,92 @@
...
+  // check if we're in the judgment change form
+  if (currentPath.includes("/judgment/")) {
+      let judgmentId = ""
+      // get the judgment id from the url
+      if (currentPath.includes("/change/")) {
+          judgmentId = currentPath.match(/\/judgment\/(\d+)\/change\/$/)[1];
+      }
+
+    // add container for duplicate alert
+    let actions_block = $("#jazzy-actions").find("div:first-child");
```
longhotsummer, line 23:
```suggestion
    let actions_block = document.querySelector("#jazzy-actions div:first-child");
```

peachjam/static/js/judgment_duplicates.js
```diff
@@ -0,0 +1,97 @@
...
+    clearTimeout(timeout);
+    timeout = setTimeout(later, wait);
+    if (callNow) func.apply(context, args);
+  };
+};
+
+function getCaseNumbers () {
+  const visible = {};
+  const hidden = {};
+  const inputs = document.querySelectorAll("input[name^='case_numbers-'], select[name^='case_numbers-']");
```
actlikewill, line 19: @longhotsummer I actually learned that you can pass multiple selectors to this function and it will bundle everything into one array :smile_cat:

peachjam_api/views.py
```diff
@@ -53,3 +57,73 @@ def post(self, request, ingestor_id):
...
+            "year",
+            "matter_type",
+        ]
+        exclude = ["document", "id"]
+
+
+class CheckDuplicatesView(APIView):
+    permission_classes = []
+
+    def post(self, request):
```
actlikewill, line 83: @longhotsummer  I have kept this as an APIView. FormView is focused on processing a single form and there is no real benefit of using it since we have to subclass the post method. We could use the base django view but we already have APIView and Response classes imported in the module

longhotsummer (approved):
great :)

nickmwangemi (approved):

PR #1527, opened 2023-09-12
Fix n+1 on judgment detail page

### LINKED ISSUE #1516: N+1 Query
Sentry Issue: [LII-HF](https://lawsafrica.sentry.io/issues/4374683698/?referrer=github_integration)
|  |  |
| **Offending Spans** | db - SELECT "peachjam_judge"."id", "peachjam_ju...

nickmwangemi:
Resolves #1516 

Before

Screenshot from 2023-09-12 07-43-43

After
Screenshot from 2023-09-12 07-50-18


peachjam/views/generic_views.py
```diff
@@ -168,6 +168,9 @@ def get_context_data(self, **kwargs):
         )
         context["labels"] = doc.labels.all()
 
+        if doc.doc_type == "judgment":
+            context["judges"] = self.add_judges(doc)
```
longhotsummer, line 172: Naming: all the other `add_xxx` methods take the context and add something to the context, hence the "add" in the name. This one doesn't add anything, it just returns something. So from a readability standpoint, either `add_judges` should be `get_judges`, or it should take `context` and update the context.

Additionally, rather put this into `views.judgment.JudgmentDetailView` since it only applies to judgments, then you don't need to check the `doc_type`.

longhotsummer (approved):

PR #1536, opened 2023-09-14
Search tweaks

### LINKED ISSUE #1462: As a user of a LII website, I would like to filter my search results by Reported cases only
Offer the user a filter option for only Reported cases. Suggested this be placed above the search results and not in the left sidebar, so users can switch on and off the reported cases results
image

actlikewill:
- [x] Facet Option Badges
- [x] Scroll to top of search page on search
- [x] Toggle advanced search back to search after click
- [x] Move sort by above doc count in mobile view
- [x] Found more than 10,000 documents
- [x] Change h5 result title

https://www.loom.com/share/41b23a15915446beafd6a8c42920b94f?sid=bb185d5d-bae9-4268-94a4-cbe6ce2066c3

closes #1462

### DIFF
--- a/peachjam/js/components/FindDocuments/FacetBadges.vue
+++ b/peachjam/js/components/FindDocuments/FacetBadges.vue
@@ -1,0 +1,48 @@
+<template>
+  <div class="my-2">
+    <button
+      v-for="option in activeOptions"
+      :key="option.value"
+      class="btn btn-outline-secondary me-2 mb-2"
+      :name="option.value"
+      @click="updateModel"
+    >
+      <i class="bi bi-x pe-none" />
+      {{ option.label }}
+    </button>
+  </div>
+</template>
+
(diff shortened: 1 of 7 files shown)


actlikewill:
one more thing: https://www.loom.com/share/b1b72909f3504d8bb4506a9fad091ad6

actlikewill:
@longhotsummer I found the above bug while working on this issue,

longhotsummer:
> @longhotsummer I found the above bug while working on this issue,

Fun!

I think the root cause is that we allow a facet to be active even when no search results match that facet. Maybe we need to compare the chosen facet values against the aggregates that we get back from ES, and only allow active facet values that are also in the list of values returned by ES. This should be relatively easy to do  inside the facet component.

longhotsummer (approved):
Great

PR #1538, opened 2023-09-15
Render PDF using single canvas object

Sandravaphilips:


peachjam/js/components/DocumentContent/pdf-renderer.ts
```diff
@@ -186,44 +187,37 @@ class PdfRenderer {
...
+
     if (this.pdfContentWrapper) {
       this.pdfContentWrapper.appendChild(pageContainer);
     }
 
     // render the page
     await page.render(renderContext).promise;
     // add the text layer
-    pageContainer.appendChild(await this.addTextLayer(page, containerWidth, canvas));
+    pageContainer.append(this.addImageLayer(page, containerWidth, canvas));
```
longhotsummer, line 198: we need to keep `addTextLayer` -- that allows the user to copy and paste text off of the page, and allows us to add text annotations.

So the image should be placed in the div, and then the image layer put on top as-per-usual. So I think you to call both `addTextLayer()` and your new `addImageLayer`.

package.json
```diff
@@ -52,6 +52,6 @@
   },
   "dependencies": {
     "bootstrap": "^5.3.1",
-    "pdfjs-dist": "^2.14.305"
+    "pdfjs-dist": "^2.16.105"
```
longhotsummer, line 55: Let's keep the older version, since it doesn't fix anything right now, and it means there are no whitespace changes in the extracted text.

longhotsummer (changes requested):
Needs to keep the text layer, pls.

longhotsummer (approved):
cool

PR #1547, opened 2023-09-20
Add functionality for reporting a document with an issue

nickmwangemi:
https://www.loom.com/share/6ede1862410a422ea5186b9bd1b20cdc?sid=143cd6aa-cab8-4f4f-89df-d80d85fd15f5


peachjam/static/js/report_document_issue.js
```diff
@@ -0,0 +1,35 @@
...
+  modalLabel.textContent = '';
+  reportForm.innerHTML = '';
+  reportForm.innerHTML = response;
+
+  setTimeout(() => {
+    const modal = document.getElementById('reportModal');
+    modal.remove();
+    window.location.href = url;
+  }, 1000);
+}
```
actlikewill, line 12: @nickmwangemi In my opinion, its better to allow the user to close the modal on their own. It will ensure that they read the success message and acknowledge that the message is sent, perhaps you could render a close button on the modal instead of closing it automatically?

actlikewill: @nickmwangemi I just saw that the close button is present in the modal. So I think all you need to do is remove this bit of code. you could then put the submit button on the same row as the close button. The default styles by bootstrap look quite good

https://getbootstrap.com/docs/5.2/components/modal/
image

longhotsummer:
This is a great start, nice end-to-end experience. I'd like to adjust it slightly to polish the experience. In particular, there are a few changes to ensure that this form isn't abused (I never believed something like this would be abused before I implemented the same feature on pmg.org.za).

1. I'd prefer the view to be a normal FormView, rather than an API view. This isn't an API that sends or expects JSON. The frontend is submitting a form, and FormView helps you process that. In particular, it performs CSRF validation, which is essential. It will also help you do form validation and separate the flows for an invalid and a valid form. Plain old HTML  and HTTP :)
2. The javascript should be a part of our javascript system, in particular  Vue component. The value of a Vue component is that the form HTML will only be rendered when the button is clicked, which will prevent abuse.
3. The page doesn't need to refresh when the feedback is sent, the user can simply dismiss the modal. Again, a Vue component makes changing the buttons easy.
4. Use modal best practices: put the buttons in the modal footer, they will right-align automatically. Make the "success" path green (".btn-success").

I'll push a skeleton of changes for the JS stuff.

peachjam/js/components/report_document_issue.ts
```diff
@@ -0,0 +1,39 @@
...
+    const form = new FormData(this.form || undefined);
+
+    fetch('/api/document-problem/', {
+      method: 'post',
+      body: form
+    }).then(response => {
+      if (response.ok) {
+        return response.text();
+      } else {
+        throw new Error('Something went wrong');
```
longhotsummer, line 25: Since this form can handle an error, how about `return "Something went wrong"` so that it's put into the body of the modal?

peachjam/templates/peachjam/emails/document_issue_email_error.html
```diff
@@ -0,0 +1,3 @@
+<div class="alert alert-danger" role="alert">
+  An error occurred while trying to send your message. Please try again later.
```
longhotsummer, line 2: These two templates must be translateable.

peachjam/templates/peachjam/emails/document_problem_email.html
```diff
@@ -0,0 +1,14 @@
...
+  <head>
+    <meta charset="UTF-8"/>
+    <title>Reported Document Problem</title>
+  </head>
+  <body>
+    <p>Hello,</p>
+    <p>We have received a report regarding the following document: {{ document_link }}.</p>
+    <p>The reported problem is described as follows:</p>
+    <p>{{ problem_description }}.</p>
+    <p>Thank you for bringing this to our attention.</p>
```
longhotsummer, line 12: This email gets sent to us as admins, not to the person reporting it.

peachjam/templates/peachjam/emails/document_problem_email.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8"/>
+    <title>Reported Document Problem</title>
+  </head>
+  <body>
+    <p>Hello,</p>
+    <p>We have received a report regarding the following document: {{ document_link }}.</p>
```
longhotsummer, line 9: Make the link clickable so we can open it easily -- remove the trailing .

peachjam_api/urls.py
```diff
@@ -43,4 +43,9 @@
         views.CheckDuplicatesView.as_view(),
         name="check_duplicates",
     ),
+    path(
+        "document-problem/",
```
longhotsummer, line 47: Move this into the normal peachjam URLs -- it's not actually an API.

peachjam_api/views.py
```diff
@@ -127,3 +131,44 @@ def post(self, request):
...
+        )
+
+        recipient_list = [settings.DEFAULT_FROM_EMAIL]
+        if email_address:
+            recipient_list.append(email_address)
+
+        return (
+            render(request, "peachjam/emails/document_issue_email_success.html")
+            if send_mail(
+                subject=f"Document problem reported on {settings.PEACHJAM['APP_NAME']}",
```
longhotsummer, line 166: Must be translateable

peachjam_api/views.py
```diff
@@ -127,3 +131,44 @@ def post(self, request):
...
+            "peachjam/emails/document_problem_email.html", context=context
+        )
+
+        recipient_list = [settings.DEFAULT_FROM_EMAIL]
+        if email_address:
+            recipient_list.append(email_address)
+
+        return (
+            render(request, "peachjam/emails/document_issue_email_success.html")
+            if send_mail(
```
longhotsummer, line 165: It's uncommon to use the inline `if` syntax when the condition is a function call -- it makes it difficult to read and debug. Rather call `send_mail()` and store the result in a variable.

peachjam_api/views.py
```diff
@@ -127,3 +131,44 @@ def post(self, request):
                 )
                 return Response(html)
         return Response()
+
+
+class DocumentProblemView(APIView):
```
longhotsummer, line 136: Make this a regular `FormView` view, and have a `DocumentProblemForm` class that has the various fields that get validated.

longhotsummer:
I've pushed some changes -- I decided to forego the Vue suggestion. Instead, there's now a Typescript component linked to the modal with the `data-component` attribute. That's how we hook javascript functionality up to plain HTML.

peachjam/forms.py
```diff
@@ -197,3 +201,44 @@ class AttachedFilesForm(AttachmentFormMixin, forms.ModelForm):
...
+        plain_txt_msg = render_to_string(
+            "peachjam/emails/document_problem_email.txt",
+            context=context,
+        )
+
+        subject = _("Document problem reported on %(app_name)s") % {
+            "app_name": settings.PEACHJAM["APP_NAME"]
+        }
+        from_email = settings.DEFAULT_FROM_EMAIL
+        recipient_list = [settings.ADMINS]
```
longhotsummer, line 235: `settings.ADMINS` should already be a list -- is making it a list here necessary?

peachjam/views/document_problem.py
```diff
@@ -0,0 +1,26 @@
...
+from peachjam.forms import DocumentProblemForm
+
+
+class DocumentProblemView(FormView):
+    template_name = "peachjam/_report_issue_modal.html"
+    form_class = DocumentProblemForm
+
+    def form_valid(self, form):
+        form = self.get_form()
+        if form.is_valid():
```
longhotsummer, line 13: This is known to be true, so you don't need to call it. Otherwise `self.form_valid()` would not be called.

peachjam/views/document_problem.py
```diff
@@ -0,0 +1,26 @@
...
+                return render(
+                    self.request, "peachjam/emails/document_issue_email_success.html"
+                )
+            else:
+                return render(
+                    self.request, "peachjam/emails/document_issue_email_error.html"
+                )
+        return super().form_valid(form)
+
+    def get_success_url(self):
```
longhotsummer, line 25: This falls away because the `form_valid()` always returns a render response.

peachjam/forms.py
```diff
@@ -197,3 +201,44 @@ class AttachedFilesForm(AttachmentFormMixin, forms.ModelForm):
...
         model = AttachedFiles
         fields = "__all__"
+
+
+class DocumentProblemForm(forms.Form):
+    document_link = forms.CharField(max_length=255, required=True)
+    problem_description = forms.CharField(widget=forms.Textarea, required=True)
+    email_address = forms.EmailField(required=False)
+
+    def send_email(self):
```
longhotsummer, line 211: Nice -- putting it here moves the logic that works with the data, close to the data.

peachjam/views/document_problem.py
```diff
@@ -0,0 +1,26 @@
...
+
+from peachjam.forms import DocumentProblemForm
+
+
+class DocumentProblemView(FormView):
+    template_name = "peachjam/_report_issue_modal.html"
+    form_class = DocumentProblemForm
+
+    def form_valid(self, form):
+        form = self.get_form()
```
longhotsummer, line 12: the `form` is passed in as a parameter

peachjam/views/document_problem.py
longhotsummer: See https://ccbv.co.uk/projects/Django/4.2/django.views.generic.edit/FormView/#post -- the FormView calls `form.is_valid()` and then calls `self.form_valid()` or `self.form_invalid()` -- which means your `form_valid` method can be really simple: the form is valid, just send the email and return the response.

peachjam/views/document_problem.py
```diff
@@ -0,0 +1,20 @@
...
+from django.views.generic.edit import FormView
+
+from peachjam.forms import DocumentProblemForm
+
+
+class DocumentProblemView(FormView):
+    template_name = "peachjam/_report_issue_modal.html"
+    form_class = DocumentProblemForm
+
+    def form_valid(self, form):
```
longhotsummer, line 11: nice, so simple :)

longhotsummer:
Nice. I decided we had to have a Vue component because otherwise we had no way to reset the form.

I have also renamed files and things so that the naming is consistent: DocumentProblem{Form, Modal, etc.}, rather than the mish-mash of "document issue", "report issue", "document problem", etc.

longhotsummer:
I also changed to use `mail_admins` which takes care of the admin recipients, checking they exist, etc.

longhotsummer (dismissed):

longhotsummer (approved):

longhotsummer (approved):

PR #1548, opened 2023-09-22
Bulk import attorneys

nickmwangemi:
https://www.loom.com/share/2ed7708a87664d63a815f3e1df5796ac?sid=1296b3db-64c9-4255-93e7-2501636e4f8b

### DIFF
--- a/peachjam/resources.py
+++ b/peachjam/resources.py
@@ -377,7 +377,6 @@ class DocumentNatureWidget(ForeignKeyWidget):
 
 class ManyToManyFieldWidget(ManyToManyWidget):
     def clean(self, value, row=None, *args, **kwargs):
-
         # Remove extra white space around and in between the judges' or attorneys' names
         if value:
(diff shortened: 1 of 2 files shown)


peachjam/resources.py
```diff
@@ -674,3 +673,12 @@ def before_save_instance(self, instance, using_transactions, dry_run):
...
             instance.is_staff = True
             instance.save()
+
+
+class AttorneyResource(resources.ModelResource):
+    class Meta:
+        model = Attorney
+        skip_unchanged = True
+        report_skipped = True
+        exclude = ("id", "description")
```
actlikewill, line 683: I don't think we need to exclude these fields,

longhotsummer: What's the impact of `report_skipped` and `skip_unchanged`? We don't use those settings elsewhere (at least not explicitly), and we can safely both create and update?

actlikewill (approved):

PR #1550, opened 2023-09-23
fix latest expression in taxonomy view; faster faceting.

longhotsummer:
This does three things:

1. allows us to fetch only the latest expression in taxonmy views, since they may include legislation, which has multiple points in time. Otherwise, the same act will show up multiple times. It's opt-in, because it's a bit more expensive, and many views (eg. court and place listing views) don't care about it.
2. fixes https://github.com/laws-africa/peachjam/issues/1522 by applying language filtering only to the queryset that is actually used for document listing. This makes faceting calculations faster because we don't care about languages when calculating facets.
3. pushes facet calculations into the db, rather than fetching all rows and then calculating facets in python

Before:

image

After:

image

### DIFF
--- a/peachjam/models/core_document_model.py
+++ b/peachjam/models/core_document_model.py
@@ -249,8 +249,8 @@ class CoreDocumentManager(PolymorphicManager):
 
 class CoreDocumentQuerySet(PolymorphicQuerySet):
     def latest_expression(self):
-        """Select only the most recent expression for documents with the same frbr_uri."""
-        return self.distinct("work_frbr_uri").order_by("work_frbr_uri", "-date")
+        """Select only the most recent expression for documents from the same work."""
+        return self.distinct("work_id").order_by("work_id", "-date")
 
     def preferred_language(self, language):
         """Return documents whose language match the preferred one,
--- a/peachjam/views/authors.py
+++ b/peachjam/views/authors.py
@@ -27,9 +27,10 @@ class AuthorDetailView(FilteredDocumentListView):
         context = super().get_context_data(**kwargs)
         doc_types = list(
             set(
-                self.form.filter_queryset(
-                    self.get_base_queryset(), exclude="doc_type"
-                ).values_list("doc_type", flat=True)
+                self.form.filter_queryset(self.get_base_queryset(), exclude="doc_type")
+                .order_by()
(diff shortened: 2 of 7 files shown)


longhotsummer:
Additionally, this also pushes facet calculation into the db:

Before:

image


After:

image

peachjam/models/core_document_model.py
```diff
@@ -249,8 +249,8 @@ def get_queryset(self):
 
 class CoreDocumentQuerySet(PolymorphicQuerySet):
     def latest_expression(self):
-        """Select only the most recent expression for documents with the same frbr_uri."""
-        return self.distinct("work_frbr_uri").order_by("work_frbr_uri", "-date")
+        """Select only the most recent expression for documents from the same work."""
+        return self.distinct("work_id").order_by("work_id", "-date")
```
longhotsummer, line 253: better to use work_id, which is indexed

peachjam/views/authors.py
```diff
@@ -27,9 +27,10 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         doc_types = list(
             set(
-                self.form.filter_queryset(
-                    self.get_base_queryset(), exclude="doc_type"
-                ).values_list("doc_type", flat=True)
+                self.form.filter_queryset(self.get_base_queryset(), exclude="doc_type")
+                .order_by()
```
longhotsummer, line 31: this clears the ordering applied by the form, otherwise we'll fetch fields required for ordering, which we don't care about

nickmwangemi (dismissed):

peachjam/views/generic_views.py
```diff
@@ -100,16 +103,20 @@ def add_facets(self, context):
...
+                    )
+                    .order_by()
+                    .values_list("authors__name", flat=True)
+                    .distinct()
                     if a
                 }
             )
 
         years = list(
             set(
```
actlikewill, line 115: could we omit `set` since we have distinct below?

actlikewill (dismissed):

peachjam/views/generic_views.py
longhotsummer: Ah, yes, we can drop set everywhere then.

PR #1553, opened 2023-09-26
fix for africanlii taxonomy listing page

longhotsummer:
image


longhotsummer:
Fixes https://lawsafrica.sentry.io/issues/4505124151/

PR #1558, opened 2023-09-28
Delay loading of flag icons, move flag icons to africanlii

nickmwangemi:
The changes include:
1) adding the `loading` attribute on the flag `` elements to instruct the browser to defer loading of the flag images that are off-screen until the user scrolls near them.
2) moving the flag icons to the `africanlii` app where they are used


longhotsummer:
This still means on africanlii the css is loaded always, unnecessarily, but at least it's cached. We don't have an easy way to only load those styles in the search page.

longhotsummer (approved):

PR #1563, opened 2023-10-03
Show entity profile for AU institutions

nickmwangemi:
image


africanlii/templates/africanlii/au_institution_detail.html
```diff
@@ -19,11 +19,9 @@
   </div>
 {% endblock %}
 {% block entity-profile %}
-  {% with entity_profile=author.entity_profile.first entity_profile_title=author.name %}
```
actlikewill, line 22: @nickmwangemi I think all you want to do here is `author.au_institution.entity_profile.first` and it works, you don't even need changes in the view code

actlikewill (approved):

PR #1568, opened 2023-10-05
Show full text in book

### LINKED ISSUE #1545: When viewing a book and I click on a TOC entry, I have no way to go back to the whole document

actlikewill:
- adds a table of content element to show books in full

https://www.loom.com/share/40e18fa86fd742b487e378909a3680d8?sid=c199ecab-2b25-47ce-85a8-64328a5d7809

closes #1545

### DIFF
--- a/peachjam/js/components/DocumentContent/index.ts
+++ b/peachjam/js/components/DocumentContent/index.ts
@@ -178,15 +178,15 @@ class DocumentContent {
       this.tocController?.addEventListener('itemTitleClicked', (e) => {
         if (this.originalDocument && this.documentElement) {
           const id = (e as CustomEvent).detail.target.getAttribute('href');
-          if (id) {
+          if (!id || id === '#') {
+            this.documentElement.replaceChildren(...Array.from(this.originalDocument.children).map(node => node.cloneNode(true)));
+          } else {
+            // @ts-ignore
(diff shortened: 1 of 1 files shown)


peachjam/js/components/DocumentContent/index.ts
```diff
@@ -178,15 +178,15 @@ class DocumentContent {
       this.tocController?.addEventListener('itemTitleClicked', (e) => {
         if (this.originalDocument && this.documentElement) {
           const id = (e as CustomEvent).detail.target.getAttribute('href');
-          if (id) {
+          if (!id || id === '#') {
+            this.documentElement.replaceChildren(...Array.from(this.originalDocument.children).map(node => node.cloneNode(true)));
```
longhotsummer, line 182: Minor: I don' think you need the `...` -- that makes a copy of the existing array AFAIK. You're already cloning the nodes.

longhotsummer (approved):

peachjam/js/components/DocumentContent/index.ts
actlikewill: it breaks when we remove the ellipses,

longhotsummer: ugh, javascript

PR #1570, opened 2023-10-11
Configure CONN_MAX_AGE

### LINKED ISSUE #1556: Configure CONN_MAX_AGE on db connections
Now that we use RDS proxy, the proxy closes inactive db connections after 10 minutes. This means that django unexpectedly finds closed connections, particularly on long-running background tasks that take more than 10 minutes (such as PDF conversions).

The docs indicate that if the database closes idle connections, then CONN_MAX_AGE should be  set to a number below that: https://docs.djangoproject.com/en/4.2/ref/databases/#connection-management

Let's try out setting CONN_MAX_AGE to, say, 9 minutes and see if that eliminates the database closed errors.

actlikewill:
- closes #1556

### DIFF
--- a/peachjam/settings.py
+++ b/peachjam/settings.py
@@ -183,6 +183,8 @@ DATABASES = {
     "default": default_db_config,
     "gazettes_africa": gazette_db_config,
 }
+CONN_MAX_AGE = int(os.environ.get("CONN_MAX_AGE", 0))
+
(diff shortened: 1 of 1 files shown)


actlikewill:
@longhotsummer sounds good. I've set it as an env variable with default to zero

longhotsummer (approved):

PR #1573, opened 2023-10-13
Fix doc count string

actlikewill:
image

The string wasn't matching correctly with the period.
The translations already have it so its fine.


longhotsummer (approved):
This behaved a bit weirdly when I was doing the translations. As long as it works :)

PR #1575, opened 2023-10-16
Fix n+1 in judgment admin

### LINKED ISSUE #1393: N+1 Query
Sentry Issue: [LII-BT](https://lawsafrica.sentry.io/issues/4300357191/?referrer=github_integration)
|  |  |
| **Offending Spans** | db - SELECT "peachjam_court"."id", "peachjam_co...

actlikewill:
- This was caused by the registry drop down.
- The __str__ method has self.name and self.court which causes the n+1 issue.
- Maybe a simple solution  could be to remove the court from the str method. The names of the registries that Tanzli are using are quite descriptive themselves, indicating the court itself, which may be enough
image

closes #1393

### DIFF
--- a/peachjam/admin.py
+++ b/peachjam/admin.py
@@ -762,11 +762,6 @@ class JudgmentAdmin(ImportExportMixin, DocumentAdmin):
 
         return fieldsets
 
-    def formfield_for_foreignkey(self, db_field, request, **kwargs):
-        if db_field.name == "registry":
-            kwargs["queryset"] = CourtRegistry.objects.select_related("court")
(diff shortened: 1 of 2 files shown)


longhotsummer (dismissed):

longhotsummer:
Another approach is to change the default manager for the CourtRegistry object so that it always loads the court. That would probably do the right thing in this case. We can't rely on registry names being sufficiently distinct, I don't think.

actlikewill:
@longhotsummer I think adding the model manager is more appropriate in this case.

longhotsummer (approved):

PR #1584, opened 2023-10-21
Exclude stubs without publication documents

goose-life:
- don't ingest them going forward
- delete existing ones


actlikewill:
@goose-life we have a similar `if document['stub']` condition here https://github.com/laws-africa/peachjam/blob/fb9f2f23f531b2b736b0509dea553f12040fb04d/peachjam/adapters/adapters.py#L295

maybe you could bundle the change under that?

actlikewill:
@goose-life if it makes logical sense to do that :smile: but it could bebetter to keep them separate

longhotsummer (approved):

goose-life:
> @goose-life if it makes logical sense to do that 😄 but it could bebetter to keep them separate

I did see that, thanks @actlikewill, but I thought it might be better to do the present check all the way at the top before going any further

nickmwangemi (approved):

PR #1594, opened 2023-10-30
Update legislation notice

nickmwangemi:
image


nickmwangemi:
> looks good! but I think the translation strings need to be updated too — or does that happen separately on peachjam?

Translation happens separately via CrowdIn

goose-life:
I was curious how you do it differently so I checked the readme, and it looks like you do need to update them?
https://github.com/laws-africa/peachjam#adding-translation-strings says to run `scripts/extract-translations.sh` and commit the changes. is the readme out of date?

nickmwangemi:
> I was curious how you do it differently so I checked the readme, and it looks like you do need to update them? https://github.com/laws-africa/peachjam#adding-translation-strings says to run `scripts/extract-translations.sh` and commit the changes. is the readme out of date?

The readme is up to date. I'll make the changes.

goose-life (approved):

PR #1597, opened 2023-10-30
Add regex to article url

### LINKED ISSUE #1591: ValidationError: '“%(value)s” value has an invalid date format. It must be in YYYY-MM-DD format.'
Sentry Issue: [LII-W2](https://lawsafrica.sentry.io/issues/4582790977/?referrer=github_integration)

```
ValidationError: '“%(value)s” value has an invalid date format. It must be in YYYY-MM-DD format.'
(14 additional frame(s) were not displayed)
...
  File "peachjam/views/article.py", line 63, in get_queryset
    return qs.filter(
```

actlikewill:
fixes #1591 

The error occurs because the date captured by the urlconf was not being validated correctly.

Adding a more explicit regex that looks for the format `YYYY-MM-DD` fixes it.

### DIFF
--- a/peachjam/urls.py
+++ b/peachjam/urls.py
@@ -16,7 +16,7 @@ Including another URLconf
 from django.conf import settings
 from django.conf.urls.static import static
 from django.contrib import admin
-from django.urls import include, path, re_path
+from django.urls import include, path, re_path, register_converter
 from django.views.decorators.cache import cache_page
 
 from peachjam.feeds import (
@@ -27,6 +27,7 @@ from peachjam.feeds import (
     LegalInstrumentAtomSiteNewsFeed,
     LegislationAtomSiteNewsFeed,
 )
+from peachjam.helpers import ISODateConverter
 from peachjam.views import (
     AboutPageView,
     ArticleDetailView,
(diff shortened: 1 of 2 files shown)


peachjam/urls.py
```diff
@@ -206,8 +206,8 @@
         ArticleTopicListView.as_view(),
         name="article_topic_list",
     ),
-    path(
-        "articles/<str:date>/<str:author>/<slug:slug>",
+    re_path(
+        "^articles/(?P<date>[0-9]{4}-[0-9]{2}-[0-9]{2})/(?P<author>[-a-zA-Z0-9_]+)/(?P<slug>[-a-zA-Z0-9_]+)$",
```
longhotsummer, line 210: It think this allows `/articles/9999-88-77/...` which still isn't valid. I don't think you can escape parsing the date in the view, catching an exception on an invalid value, and returning a 404.

Put another way: you always need to validate date URL params in some way :)

actlikewill: @longhotsummer I went with a custom path converter as shown here https://docs.djangoproject.com/en/4.2/topics/http/urls/#registering-custom-path-converters
I think its a better place to do the url validation rather than in the view itself

longhotsummer (approved):
fancy

nickmwangemi (approved):

PR #1600, opened 2023-10-31
Lawreader

longhotsummer:
* updated styles for books for lawreader
* TOC shows active portion only at first level (otherwise behaves like normal below that)
* don't randomize generated TOC ids, to allow them to work when shared
* allow blockquote in html
* adjust CSS styles for book document type, for readability and visual interest

https://www.loom.com/share/7388e9997ec34af1b99748a8cd6c8b22

### DIFF
--- a/peachjam/js/components/DocumentContent/index.ts
+++ b/peachjam/js/components/DocumentContent/index.ts
@@ -27,24 +27,28 @@ class OffCanvas {
 class DocumentContent {
   protected root: HTMLElement;
   private pdfRenderer: PdfRenderer | undefined;
-  private searchApp: any;
   private navOffCanvas: OffCanvas | undefined;
-  private enchrichmentsManager: EnrichmentsManager | null = null;
+  private enrichmentsManager: EnrichmentsManager | null = null;
   private originalDocument: HTMLElement | undefined;
   private tocController: HTMLElement| null = null;
   private documentElement: Element | null;
+  private tocShowActiveItemOnly: boolean = false;
+  private displayType: string;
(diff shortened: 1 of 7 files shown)


actlikewill:
@longhotsummer I noticed that the Book model does not take on the "book" frbr_uri_subtype and so does not get the proper TOC treatment
I think that this pre_save method is not working properly: https://github.com/laws-africa/peachjam/blob/86cb721f6f87ea19afa16455c8e63e3128d642d2/peachjam/models/journals_books.py#L9C2-L14C1

longhotsummer:
> @longhotsummer I noticed that the Book model does not take on the "book" frbr_uri_subtype and so does not get the proper TOC treatment I think that this pre_save method is not working properly: https://github.com/laws-africa/peachjam/blob/86cb721f6f87ea19afa16455c8e63e3128d642d2/peachjam/models/journals_books.py#L9C2-L14C1

Thanks. It looks like we need to force the `nature` object, i'll do that.

longhotsummer:
> > @longhotsummer I noticed that the Book model does not take on the "book" frbr_uri_subtype and so does not get the proper TOC treatment I think that this pre_save method is not working properly: https://github.com/laws-africa/peachjam/blob/86cb721f6f87ea19afa16455c8e63e3128d642d2/peachjam/models/journals_books.py#L9C2-L14C1
> 
> Thanks. It looks like we need to force the `nature` object, i'll do that.

Fixed.

actlikewill (approved):
Works well :+1:

PR #1605, opened 2023-11-01
Switch indigo api from v2 to v3

nickmwangemi:
- Updates the hardcoded Indigo API URL 
- Updates the field to fetch commencements from when creating relationships for imported documents
- Includes a data migration to update the `api_url` field for ingestors 

https://www.loom.com/share/d3bf125b17684c2fb1e0b452ed13e840?sid=3379c8b7-f231-4b1e-90a8-0b133a5c8283


peachjam/adapters/adapters.py
```diff
@@ -395,8 +395,9 @@ def fetch_relationships(self, imported_document, created_document):
                         object_work=amending_work,
                     )
 
-        if imported_document["commencements"]:
-            for commencement in imported_document["commencements"]:
+        commencements = created_document.commencements_json
```
goose-life, line 398: I'm not as familiar with this code so it might be obvious, but why are we now looking at the created document's commencements rather than the imported document's?

actlikewill: the imported_document is the document metadata as received from indigo which originally contained commencements. Now the commencement data is a field on the document itself. We could  pull this code out into its own function like `get commencements`, do the api call there and update the created document and the relationships in one step, but I think its okay this way too.

actlikewill (dismissed):

actlikewill:
@nickmwangemi have you thought about how you will deploy this? is there is a potential for breakage?

nickmwangemi:
> @nickmwangemi have you thought about how you will deploy this? is there is a potential for breakage?

I'll do a data migration to update the ingestors, which will be much easier than manually doing this across the sites. I'll include that in this PR.

peachjam/adapters/adapters.py
nickmwangemi: @actlikewill Thanks for the explanation

longhotsummer (approved):

actlikewill (approved):

PR #1607, opened 2023-11-02
fixes rwandalii deploy

actlikewill:
fixes deploy


actlikewill:
@longhotsummer could this be the reason why the build is failing? I thought once the code has been pushed to the server everything should work https://github.com/laws-africa/peachjam/actions/runs/6731675145/job/18297765138

longhotsummer (approved):
Yes, probably, because it doesn't have a db config to connect to, so it's trying localhost and failing.

PR #1613, opened 2023-11-06
Add overturned and upheld predicates

### LINKED ISSUE #1245: As a reader I'd like to know when a judgment has been overturned by a higher court

nickmwangemi:
Resolves #1245


peachjam/migrations/0109_add_overturned_and_upheld_predicates.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 3.2.19 on 2023-11-14 09:13
+
+from django.db import migrations
+
+
+def add_overturned_and_upheld_predicates(apps, schema_editor):
```
longhotsummer, line 6: This migration will fail if this predicate already exists, so it's much better to use `update_or_create`. The goal of the migration is to ensure it exists, and it must tolerate the case where it already does.

longhotsummer (changes requested):
this still doesn't work correctly

longhotsummer (approved):

PR #1618, opened 2023-11-08
Allow judges to be deleted

actlikewill:
- Remove protected attribute from bench model to allow judges to be deleted in the admin.


longhotsummer:
Isn't the correct approach to unlink the judgment from the judge? The point of this protection is so that a judge linked to judgments is not accidentally deleted.

actlikewill:
@longhotsummer Right. closing this.

PR #1621, opened 2023-11-15
Filter ingested documents by actor

nickmwangemi:
To test locally:
1) Make sure you have "African Regional Bodies" setup as a country.
image

2) Make sure you have "African Union" setup as a locality.
image

3) Setup an ingestor with the following `token`, `api_url`, `places` and `actor` ingestor settings.
image

4) Run `./manage.py process_tasks`.

### DIFF
--- a/peachjam/adapters/adapters.py
+++ b/peachjam/adapters/adapters.py
@@ -109,8 +109,18 @@ class IndigoAdapter(Adapter):
             url = f"{self.api_url}/akn/{place_code}/.json"
             while url:
                 res = self.client_get(url).json()
-                results.extend(res["results"])
-                url = res["next"]
+
+                # Filter by actor, if setting is present
+                actor = self.settings.get("actor", None)
+                if actor:
+                    results_filtered_by_actor = [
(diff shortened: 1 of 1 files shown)


peachjam/adapters/adapters.py
```diff
@@ -109,8 +109,18 @@ def get_doc_list(self):
             url = f"{self.api_url}/akn/{place_code}/.json"
             while url:
                 res = self.client_get(url).json()
-                results.extend(res["results"])
-                url = res["next"]
+
+                # Filter by actor, if setting is present
+                actor = self.settings.get("actor", None)
```
goose-life, line 114: doesn't .get() return None by default if nothing is found?

goose-life (approved):
lgtm

peachjam/adapters/adapters.py
nickmwangemi: > doesn't .get() return None by default if nothing is found?

Yes, it does.

longhotsummer (approved):

PR #1625, opened 2023-11-21
Make recent document list template reusable

nickmwangemi:
Includes changes to make the template reusable 

This [PR](https://github.com/laws-africa/paplii/pull/4) depends on it.


longhotsummer:
This is simply turning the entire template into a block. A cleaner way to override that is just to add a new `templates/peachjam/_recent_document_list.html` file to the app where it must be overridden.

PR #1629, opened 2023-11-27
Improve reporting of issues on documents

nickmwangemi:


peachjam/static/js/pdf.worker-prod.js
longhotsummer, line 1: neither of the js files should have changed.

peachjam/js/components/DocumentProblemModal.vue
```diff
@@ -25,12 +26,45 @@
...
+              <option
+                v-for="option in options"
+                :key="option.text"
+                :value="option.value"
+              >
+                {{ option.text }}
+              </option>
+            </select>
+          </div>
+          <div v-if="selectedOption === 'Other'">
```
longhotsummer, line 49: This is overly complicated. The user already has a box they can type in a message -- this field isn't necessary and can be removed.

peachjam/js/components/DocumentProblemModal.vue
```diff
@@ -83,7 +120,18 @@ export default {
       problem: '',
       submitted: false,
       success: true,
-      url: window.location.toString()
+      url: window.location.toString(),
+      selectedOption: '',
+      other: '',
+      isVisible: true,
+      options: [
```
longhotsummer, line 127: I think it's easier simply to put the options directly in the HTML inside the select element.

Additionally, the labels (not the values) must be translateable.

peachjam/forms.py
```diff
@@ -205,17 +211,23 @@ class Meta:
 
 class DocumentProblemForm(forms.Form):
     document_link = forms.CharField(max_length=255, required=True)
-    problem_description = forms.CharField(widget=forms.Textarea, required=True)
-    email_address = forms.EmailField(required=False)
+    problem_description = forms.CharField(widget=forms.Textarea, required=False)
+    other_problem_description = forms.CharField(widget=forms.Textarea, required=False)
```
longhotsummer, line 215: this field can fall away everywhere, it's not helpful

peachjam/models/settings.py
```diff
@@ -104,5 +104,20 @@ def __str__(self):
         return "Settings"
 
 
+class AdminEmail(models.Model):
```
longhotsummer, line 107: I think this is too complicated. We only need the email address. A space-separated list of email addresses on a single CharField on PeachJamSettings is simpler and is sufficient.

peachjam/forms.py
```diff
@@ -232,9 +244,21 @@ def send_email(self):
...
+        admin_emails = [admin[1] for admin in settings.ADMINS]
+
+        # get admin emails from site settings
+        site_admin_emails = list(
+            pj_settings().admin_emails.values_list("email", flat=True)
+        )
+
+        recipients = admin_emails + site_admin_emails
+
+        email = EmailMultiAlternatives(
```
longhotsummer, line 257: `send_mail` will do all this for you. It's very similar to the `mail_admins` that was there previously, you just need to add `recipient_list=...`.

longhotsummer (changes requested):
A number of things can be simplified, please.

peachjam/forms.py
```diff
@@ -228,10 +237,19 @@ def send_email(self):
...
-            "app_name": settings.PEACHJAM["APP_NAME"]
+            "app_name": app_name
         }
 
+        # fetch admin emails
+        site_admin_emails = pj_settings().admin_emails.split()
+        if site_admin_emails:
+            # list of (app_name, email) tuples
+            app_admin_emails = [(app_name, email) for email in site_admin_emails]
+            settings.ADMINS.extend(app_admin_emails)
```
longhotsummer, line 251: Don't manipulate the `ADMINS` constant -- you're going to manipulate it multiple times and it's going to go wrong when the setting in the admin area is changed.

Instead, as suggested, use `send_mail` instead and give it both the emails from `settings.ADMINS` and your new `admin_emails` as receipients.

peachjam/models/settings.py
```diff
@@ -96,6 +96,13 @@ class PeachJamSettings(SingletonModel):
         null=True,
         blank=True,
     )
+    admin_emails = models.CharField(
```
longhotsummer, line 99: Much simpler :)

peachjam/js/components/DocumentProblemModal.vue
```diff
@@ -24,13 +25,31 @@
...
+              {{ $t('Problem category') }}
+              <span class="text-danger">*</span>
+            </label>
+            <select
+              v-model="problem_category"
+              class="form-control"
+              name="problem_category"
+              required
+            >
+              <option>{{ $t('Incorrect information') }}</option>
```
longhotsummer, line 39: Do these do the correct thing even though they don't have `value` attributes?

peachjam/forms.py
```diff
@@ -232,9 +241,25 @@ def send_email(self):
...
+                )
+            else:
+                send_mail(
+                    subject=subject,
+                    message=plain_txt_msg,
+                    from_email=None,
+                    recipient_list=default_admin_emails,
+                    html_message=html,
+                    fail_silently=False,
+                )
```
longhotsummer, line 265:
```suggestion
        default_admin_emails = [email for name, email in settings.ADMINS]
        site_admin_emails = pj_settings().admin_emails.split()
        send_mail(
            subject=subject,
            message=plain_txt_msg,
            from_email=None,
            recipient_list=default_admin_emails + site_admin_emails,
            html_message=html,
            fail_silently=False,
        )
```

longhotsummer (approved):

PR #1630, opened 2023-11-29
Show search explanations and scores for users with suitable perms

longhotsummer:
This should make debugging search results in production a bit easier.

https://www.loom.com/share/79bf916f354e4629aa6c46e3b5d41b14


actlikewill (approved):
super useful

PR #1631, opened 2023-11-29
Rss tweaks

nickmwangemi:


peachjam/feeds.py
```diff
@@ -85,11 +84,25 @@ class CoreDocumentAtomSiteNewsFeed(CoreDocumentFeed, BaseAtomFeed):
...
+    def add_item_elements(self, handler, item):
+        super().add_item_elements(handler, item)
+        handler.addQuickElement("image", item.get("image", ""))
+        handler.addQuickElement("body", item.get("body", ""))
+
+
 class ArticleFeed(BaseFeed):
     model = Article
-    title = "Articles"
+    title = "AfricanLII Articles"
```
longhotsummer, line 96: This is in peachjam, not africanlii. So we shouldn't hardcode. What if you pull in the app name from settings.py?

peachjam/templates/peachjam/layouts/base.html
```diff
@@ -40,7 +40,7 @@
       {% ifequal request.path '/articles/' %}
         <link rel="alternate"
               type="application/atom+xml"
-              title="Articles RSS Feed"
+              title="AfricanLII Articles RSS Feed"
```
longhotsummer, line 43: This is in `peachjam`, not AfricanLII, so you can't hardcode this. Use the APP_NAME from settings.py instead.

longhotsummer (changes requested):
base.html must use APP_NAME

please include a screenshot of the resulting RSS feed.

longhotsummer (approved):

PR #1635, opened 2023-12-04
calculate pagerank pivot to improve pagerank scores

longhotsummer:
this allows us to exclude the zero-ranked pages, which dilute the geometric mean calculated by ES too much

ref https://github.com/laws-africa/peachjam/issues/1161#issuecomment-1838515467

Before:

image

After:

image


actlikewill (approved):
very interesting

PR #1644, opened 2023-12-07
Handle NoneType error in case admin emails are not set

nickmwangemi:
Resolves https://github.com/laws-africa/peachjam/issues/1643


longhotsummer (dismissed):
Another option is:

```python
site_admin_emails = (pj_settings().admin_emails or '').split()
```

PR #1645, opened 2023-12-11
Add locality and country fields to court model

nickmwangemi:
- Adds locality and country fields to the Court model
- Enforces the court's country and locality on the judgment, if set.


peachjam/models/judgment.py
```diff
@@ -344,6 +359,13 @@ def pre_save(self):
         self.assign_title()
         super().pre_save()
 
+    def save(self, *args, **kwargs):
+        # ensure judgment locality matches the court's locality
+        if self.court.locality:
```
longhotsummer, line 364: Do this in `pre_save` because it changes the MNC. 

Also force the country if it is set.

peachjam/migrations/0113_book_content_markdown.py
```diff
@@ -7,7 +7,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("peachjam", "0112_peachjamsettings_pagerank_pivot_value"),
+        ("peachjam", "0113_auto_20231211_0917"),
```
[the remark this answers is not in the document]
nickmwangemi: Reverse, as in, my migration depends on yours?

peachjam/models/judgment.py
```diff
@@ -101,6 +102,20 @@ class Court(models.Model):
         "peachjam.EntityProfile", verbose_name=_("profile")
     )
     order = models.IntegerField(_("order"), null=True, blank=True)
+    country = models.ForeignKey(
+        Country,
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE,
```
longhotsummer, line 109: Why cascade this, but protect the locality? I think better to protect against deleting both.

longhotsummer:
I don't think this is working. Please include a video showing it does.

nickmwangemi:
https://www.loom.com/share/bc67808f879d4a9aa69abaf425e664f3?sid=58d64b7c-e8a9-4521-9cf0-18354ef01c10

peachjam/models/judgment.py
```diff
@@ -339,6 +354,14 @@ def pre_save(self):
         if self.registry:
             self.court = self.registry.court
 
+        # ensure judgment locality matches the court's locality
+        if self.locality and self.court.locality:
```
longhotsummer, line 358: surely we want the locality to be set if the court has one, even if the self.locality is unset? Same for jurisdiction. If the court has either, set it on the judgment.

longhotsummer (approved):

PR #1670, opened 2024-01-08
Handle NoneType error

nickmwangemi:


peachjam/models/judgment.py
```diff
@@ -354,7 +354,7 @@ def pre_save(self):
         if self.registry:
             self.court = self.registry.court
 
-        if self.court.country:
+        if self.court is not None and self.court.country:
             self.jurisdiction = self.court.country
 
         if self.court.locality:
```
longhotsummer, line 360: same check must be added here

longhotsummer (approved):

PR #1678, opened 2024-01-12
Don't show relationships for deleted documents

nickmwangemi:
Before:
image

After:
Screenshot from 2024-01-12 15-50-44

### DIFF
--- a/peachjam/views/generic_views.py
+++ b/peachjam/views/generic_views.py
@@ -218,16 +218,22 @@ class BaseDocumentDetailView(DetailView):
         return result
 
     def add_relationships(self, context):
-        context["relationships_as_subject"] = rels_as_subject = list(
-            Relationship.for_subject_document(context["document"])
-            .prefetch_related("subject_work", "object_work")
-            .select_related("predicate")
-        )
-        context["relationships_as_object"] = rels_as_object = list(
-            Relationship.for_object_document(context["document"])
-            .prefetch_related("object_work", "subject_work")
-            .select_related("predicate")
-        )
+        relationships_as_subject = [
+            relationship
+            for relationship in Relationship.for_subject_document(context["document"])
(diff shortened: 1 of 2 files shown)


peachjam/views/generic_views.py
```diff
@@ -218,16 +218,22 @@ def fetch_docs(self, works):
...
+        ]
+
+        relationships_as_object = [
+            relationship
+            for relationship in Relationship.for_object_document(context["document"])
+            if relationship.subject_work
+            and CoreDocument.objects.filter(work=relationship.subject_work)
+        ]
+
+        context["relationships_as_subject"] = rels_as_subject = relationships_as_subject
```
longhotsummer, line 235: no need for both `rels_as_subject` and `relationships_as_subject`

peachjam/views/generic_views.py
```diff
@@ -218,16 +218,22 @@ def fetch_docs(self, works):
...
-        context["relationships_as_object"] = rels_as_object = list(
-            Relationship.for_object_document(context["document"])
-            .prefetch_related("object_work", "subject_work")
-            .select_related("predicate")
-        )
+        relationships_as_subject = [
+            relationship
+            for relationship in Relationship.for_subject_document(context["document"])
+            if relationship.object_work
+            and CoreDocument.objects.filter(work=relationship.object_work)
```
longhotsummer, line 225: This will call the database once for every relationship, which is very inefficient. You generally want to do cross-object filtering like this in the database.

Try something like this instead:

```python
Relationship.for_subject_document(context["document"])
  .filter(subject_work__documents__isnull=False)
  .distinct('pk')
```

when we join `subject_work` to `documents` we could have multiple results, so we need to add the `distinct` to ensure we get unique relationships only.

longhotsummer (approved):

PR #1681, opened 2024-01-16
Make order outcome field multi-select

nickmwangemi:
- Adds functionality for selecting multiple order outcomes for a judgment.

Before:
Screenshot from 2024-01-16 16-10-18

After:
Screenshot from 2024-01-16 16-39-21

Screenshot from 2024-01-18 13-45-58

Screenshot from 2024-01-18 13-45-42


longhotsummer:
Does this discard any existing data? We probably need to add the new field, do a data migration to populate it, then delete the old field.

The judgment detail page must be updated to show a list of outcomes, rather than a singleton.

peachjam/migrations/0115_copy_and_delete_old_order_outcome_field_data.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 3.2.19 on 2024-01-17 09:20
+
+from django.db import migrations, models
+
+
+def copy_data(apps, schema_editor):
+    Judgment = apps.get_model("peachjam", "Judgment")
+
+    for judgment in Judgment.objects.all().iterator(chunk_size=100):
```
longhotsummer, line 9:
```suggestion
    for judgment in Judgment.objects.filter(order_outcome__isnull=False).iterator(chunk_size=100):
```

otherwise we have to iterate and update 70,000 judgments that aren't changing.

longhotsummer: Also, looks like the migration needs to be tweaked:

```
  Applying peachjam.0115_copy_and_delete_old_order_outcome_field_data...Traceback (most recent call last):
  File "/Users/greg/coding/laws.africa/peachjam/.env/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.NotNullViolation: null value in column "orderoutcome_id" of relation "peachjam_judgment_order_outcomes" violates not-null constraint
```

longhotsummer:
This is looking good. Please check if peachjam_search/documents.py needs to be updated to support serialising a list.

peachjam_search/serializers.py
```diff
@@ -72,8 +72,12 @@ def get_court(self, obj):
...
         return obj["nature" + self.language_suffix]
 
-    def get_order_outcome(self, obj):
-        return obj["order_outcome" + self.language_suffix]
+    def get_order_outcomes(self, obj):
+        order_outcomes = obj["order_outcomes"]
+        if order_outcomes:
+            return [
+                order_outcome + self.language_suffix for order_outcome in order_outcomes
+            ]
```
longhotsummer, line 80:
```suggestion
        return obj["order_outcomes" + self.language_suffix]
```

nickmwangemi: @longhotsummer I'm getting this error when I apply your suggestion.
image

peachjam/templates/peachjam/judgment_detail.html
```diff
@@ -42,14 +42,18 @@
...
-  {% if document.order_outcome %}
-    <dt>
-      {% trans 'Order' %}
-    </dt>
-    <dd class="text-muted">
-      {{ document.order_outcome.name }}
-    </dd>
-  {% endif %}
+  {% with document.order_outcomes.all as order_outcomes %}
+    {% if document.order_outcomes %}
```
longhotsummer, line 46:
```suggestion
    {% if order_outcomes %}
```

peachjam/templates/peachjam/judgment_detail.html
```diff
@@ -42,14 +42,18 @@
...
+      <dt>
+        {% trans 'Order' %}
+      </dt>
+      <dd class="text-muted">
+        {% for order_outcome in order_outcomes %}
+          {% if not forloop.last %}{{ order_outcome.name }},{% endif %}
```
longhotsummer, line 52: always render the name. the forloop.last only applies to the comma

longhotsummer:
please test fully with a judgment with multiple outcomes, including the court detail page for that judgment. Filtering the court listing page by an outcome doesn't work because forms.py must be updated.

nickmwangemi:
> please test fully with a judgment with multiple outcomes, including the court detail page for that judgment. Filtering the court listing page by an outcome doesn't work because forms.py must be updated.

Judgment with multiple order outcomes:

image

Filtering on the court listing page:
Screenshot from 2024-01-19 09-44-20

PR #1691, opened 2024-01-18
Create DocumentMedia model

nickmwangemi:


peachjam/adapters/adapters.py
```diff
@@ -345,6 +346,36 @@ def update_document(self, url):
...
         self.fetch_relationships(document, created_doc)
+        self.download_and_save_document_images(document, created_doc)
+
+    def list_images_from_content_api(self, document):
+        links = document["links"]
+        if links:
+            for link in links:
+                if link["href"].endswith("media.json"):
+                    response = self.client_get(link["href"])
+                    if response.status_code == 200 and response.json()["results"]:
```
longhotsummer, line 357: `self.client_get` will raise an exception on a non-200 response, so you can remove that check.

peachjam/adapters/adapters.py
```diff
@@ -345,6 +346,36 @@ def update_document(self, url):
...
+                        return response.json()["results"]
```
longhotsummer, line 358: the function name says it returns images, so it should filter on the mime type here. Filter on `.startswith("image/")` not `in`.

peachjam/admin.py
```diff
@@ -1076,6 +1077,11 @@ class AttorneyAdmin(ImportExportMixin, admin.ModelAdmin):
     list_display = ("name", "description")
 
 
+@admin.register(DocumentMedia)
+class DocumentMediaAdmin(admin.ModelAdmin):
```
longhotsummer, line 1081: This should be an inline in document admin, so it shows up as a tab (like topics, etc.)

peachjam/models/core_document_model.py
```diff
@@ -912,3 +912,30 @@ def update_or_create_for_document(cls, document):
...
+        verbose_name=_("document"),
+    )
+    filename = models.CharField(_("filename"), max_length=1024, null=False, blank=False)
+    file = models.FileField(
+        _("file"),
+        upload_to=file_location,
+        max_length=1024,
+    )
+    mime_type = models.CharField(
+        _("mime_type"), max_length=1024, null=False, blank=False
```
longhotsummer, line 932:
```suggestion
        _("mime type"), max_length=1024, null=False, blank=False
```

peachjam/urls.py
```diff
@@ -257,6 +258,12 @@
     ),
     # django-markdown-editor
     path("martor/", include("martor.urls")),
+    # Get a specific media file for a document
+    re_path(
+        r"documents/(?P<document_id>[0-9]+)/media/(?P<filename>.*)$",
```
longhotsummer, line 263: this must be the FRBR URI version of this url, not the document id version. Otherwise the media won't load correctly. See how it is described in the original issue.

peachjam/adapters/adapters.py
```diff
@@ -345,6 +346,37 @@ def update_document(self, url):
...
+    def download_and_save_document_images(self, document, created_document):
+        image_list = self.list_images_from_content_api(document)
+        if image_list:
+            for result in image_list:
+                if result["mime_type"].startswith("image/"):
+                    with NamedTemporaryFile() as file:
+                        r = self.client_get(result["url"])
+                        file.write(r.content)
+
+                        Image.objects.get_or_create(
```
longhotsummer, line 369: Knowing when to update the file is difficult. I suggest simply deleting all images associated with the file right at the start of the function (line 360). Then assume that the filenames are unique, and simply create the file rather than `get_or_create`.

peachjam/views/generic_views.py
```diff
@@ -165,6 +165,7 @@ def get_context_data(self, **kwargs):
             context["display_type"] = (
                 "akn" if context["document"].content_html_is_akn else "html"
             )
+        if not context["document"].content_html_is_akn:
```
longhotsummer, line 168: You've broken the indent of the existing if statement.

longhotsummer:
Looking good -- pls include a video/screenshot.

nickmwangemi:
@longhotsummer I've updated the `prefix_images` method to remove the extra `/media` from the image url

Screenshot from 2024-01-25 19-06-27
Screenshot from 2024-01-25 19-07-12

peachjam/views/generic_views.py
```diff
@@ -274,14 +277,14 @@ def get_notices(self):
...
-        """Rewrite image URLs so that we can server them correctly."""
+        """Rewrite image URLs so that we can serve them correctly."""
         root = html.fromstring(document.content_html)
 
         for img in root.xpath(".//img[@src]"):
             src = img.attrib["src"]
             if not src.startswith("/") and not src.startswith("data:"):
                 img.attrib["src"] = (
-                    document.expression_frbr_uri + "/media/" + img.attrib["src"]
+                    document.expression_frbr_uri + "/" + img.attrib["src"]
```
longhotsummer, line 287: this must not change. The documents that have `content_html_is_akn == True` have `media/` already added, and so must not be processed through this function (which you have already added a guard for on line 170 above).

The documents that have `content_html_is_akn == False` MUST have this added because they don't prefix their images at all. This is for when images are embedded in Word documents.

longhotsummer (approved):

PR #1702, opened 2024-01-25
As an editor fixing mistakes, it would be useful to be able to temporarily unpublish a document rather than deleting it

### LINKED ISSUE #1688: As an editor fixing mistakes, it would be useful to be able to temporarily unpublish a document rather than deleting it
It is occasionally useful to be able to remove a document from public visibility without actually deleting it. This means it can be corrected (for example anonymised), which can take a few days, without it having to be deleted completely and re-added later.

> I had one last week when I landed up deleting then recreating because I couldn't Unpublish. Don't immediately remember circumstances. We used to use Unpublish a fair amount in my early days at AfricanLII. I have found it useful on occasion when a case needs to be inaccessible to the public for a while eg belated decision to anonymise. It saves having to recapture the record when the problem has been resolved.

Sandravaphilips:
Closes #1688
https://www.loom.com/share/f81c6a462db9449ca867fda83ff2bd49?sid=fa93b543-0405-4817-829c-e5bade43d3ed


longhotsummer:
This is a good start.

Most (all, I think) of our Listing views and our Detail views extend base classes within peachjam.views.generic_views. You should be able to add `filter(published=True)` there, rather than in every other type of view.

Check `BaseDocumentDetailView.get_object` and `DocumentListView.get_base_queryset`.

For `DocumentListView`, I think it's safer to add the filter in `get_base_queryset` (rather than changing `queryset`) because that method tries `.queryset` and `.model.objects`.

peachjam/admin.py
```diff
@@ -384,6 +386,7 @@ class DocumentAdmin(BaseAdmin):
                     "title",
                     "date",
                     "language",
+                    "published",
```
longhotsummer, line 389: Pls move this under the "Advanced" tab, after `allow_robots`.

peachjam/models/core_document_model.py
```diff
@@ -427,6 +427,11 @@ class CoreDocument(PolymorphicModel):
         help_text=_("Allow this document to be indexed by search engine robots."),
     )
 
+    published = models.BooleanField(
+        _("published"),
+        default=True,
```
longhotsummer, line 432:
```suggestion
        default=True,
        help_text=_("Is this document published and visible on the website?"),
```

longhotsummer: pls add db_index=True so it's indexed

Sandravaphilips:
@longhotsummer thanks for the feedback.
However, not all ListViews extend the base class, so I retained the ones that don't. 
For the DetailView, I stuck with raising a 404 error in the DetailViewResolver instead. Implementing it in BaseDetailView raises a _cannot find matching query error_ even when I say to raise the 404 error if the object is empty.

peachjam/views/judgment.py
```diff
@@ -14,7 +14,9 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
         context["court_classes"] = CourtClass.objects.prefetch_related("courts")
-        context["recent_judgments"] = Judgment.objects.order_by("-date")[:30]
+        context["recent_judgments"] = Judgment.objects.exclude(
```
longhotsummer, line 17: Hmm, we need to do the same in all the `HomePageView` classes -- also search for `recent_*` and ensure those are corrected (mostly in homepage views I think).

longhotsummer (commented):
Great, much cleaner.

Just the homepage views, pls.

longhotsummer:
Please also update `peachjam_search.documents.SearchableDocument.should_index_object` to skip unpublished documents.

I don't yet have a good recommendation for removing an unpublished doc from search, but we can leave that for now.

longhotsummer (approved):

PR #1703, opened 2024-01-26
Ingest aliases

nickmwangemi:


longhotsummer (dismissed):
Looks good. However, until Indigo has this field added, it will break because `document["aliases"]` won't exist.

So either you need to wait until it's deployed. Or, make it more robust with:

```
aliases = document.get("aliases", [])
```

that's probably easier and safer

longhotsummer (approved):

shailen-naidoo:
This looks great, I am going to give it a try.

shailen-naidoo:
I tried this on the site but it is not working, I am assuming that it is merged in but still needs to be QA tested?

longhotsummer:
Hi @shailen-naidoo the supporting part of this in another system is not yet enabled.

PR #1727, opened 2024-02-13
use p99 as pagerank pivot

longhotsummer:
the geometric mean is too small; only at higher values do we get reasonable score influence

see https://colab.research.google.com/drive/1KlYC7A9JeqS_uaLhL8yv_IrAmHHXNCdK

### DIFF
--- a/peachjam/graph/ranker.py
+++ b/peachjam/graph/ranker.py
@@ -1,5 +1,5 @@
 import logging
-from statistics import geometric_mean
+import math
 
 import igraph as ig
 from elasticsearch import helpers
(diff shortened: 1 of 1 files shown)


actlikewill:
@longhotsummer I dont have permissions to view the collab, Is it yours?

longhotsummer:
with pivot (k) = p99:

image

with geometric mean:

image

longhotsummer:
> @longhotsummer I dont have permissions to view the collab, Is it yours?

@wilson moved into shared folder, try now?

actlikewill (approved):
interesting. maybe at some point we should discuss these changes in a bit more detail so I can understand it better.

longhotsummer:
> interesting. maybe at some point we should discuss these changes in a bit more detail so I can understand it better.

definitely. i can't say that I'm a pro, but I can outline how it works.

PR #1745, opened 2024-02-28
Fix bulk importer updates

actlikewill:
- This adds a method for calling the objects without defer in the fields for use in the importer


actlikewill:
@longhotsummer I'm thinking maybe it would have been better to leave the original queryset untouched and then have a manager with something like `defered_objects` so we can opt in to the defered qs when necessary

longhotsummer (approved):

PR #1747, opened 2024-02-29
Lower court judges

### LINKED ISSUE #1666: As an editor, for analytics purposes, i.e. not public, I would like to capture in metadata the name of the judge in the lower court judgment
So, this would create two categories of judges associated with a judgment 
- judges who have written and signed the current judgment document
- judge(s) who wrote the judgment that is being appealed or overturned

actlikewill:
- This adds lower court judges

image

closes #1666


actlikewill:
@longhotsummer I'm trying to add an LowerBench inline similar to what we did for the Bench. I'm reusing the same table as shown below but I keep getting an error when I run make migrations. How did you solve this? I think you moved the migrations around somehow
image
image

longhotsummer:
@actlikewill remove the "db_table" setting; that was only used to build on the existing table, which isn't necessary for this new table.

PR #1755, opened 2024-03-04
Fixes isodate converter

actlikewill:
- This allows the converter reverse function to work with datetime objects.

### DIFF
--- a/peachjam/helpers.py
+++ b/peachjam/helpers.py
@@ -70,6 +70,8 @@ class ISODateConverter:
 
     def to_url(self, value):
         # invalid values will raise ValueError which will raise NoReverseMatch
+        if isinstance(value, datetime):
+            return value.strftime("%Y-%m-%d")
         return datetime.strptime(value, "%Y-%m-%d").date().strftime("%Y-%m-%d")
 
 


peachjam/helpers.py
```diff
@@ -70,6 +70,8 @@ def to_python(self, value):
 
     def to_url(self, value):
         # invalid values will raise ValueError which will raise NoReverseMatch
+        if isinstance(value, datetime):
```
longhotsummer, line 73: Minor: we're now repeating the same code in two places. A better approach would be to get the value into the expected format (datetime) if it's not there already:

```python
if isinstance(value, str):
    value = datetime.strptime(value, "%Y-%m-%d")
elif not isinstance(value, (datetime, date)):
    raise ValueError(value)
return value.strftime("%Y-%m-%d")
```

longhotsummer (approved):

peachjam/helpers.py
actlikewill: @longhotsummer agreed, converting the values to datetime and raising errors for all others is a better approach, updated

longhotsummer (approved):

Issue #1737, opened 2024-02-20
As an editor uploading judgments, I want to see the correct document when I make edits to a file.


niiroobiro:
Editors occasionally need to replace the initially uploaded version of judgment with an alternative version. The reason is that the first copy that is sent is a video graphic version and the second is usually the correct one. Most of the time, the editors find that they have already uploaded the first version and need to change that.

**User needs Description:**

The updated file does not appear on the site after replacing the original file, even though it shows as changed in the source files.
When an editor clicks view on the site, you still view the original file instead of the replaced copy.

**Conclusion:**

I believe what’s happening from the two examples below is that an editor will replace a file and when the change is not reflected on the site, they will delete the file altogether because it’s better to have no document than a wrong one. That’s why most users find no pdf /pdf downloads on the site.

**Examples of such judgments:**

https://tanzlii.org/akn/tz/judgment/tzhc/1981/1/eng@1981-12-03
https://tanzlii.org/akn/tz/judgment/tzhc/2023/23435/eng@2023-12-15

niiroobiro:
@longhotsummer We have determined that the source file is not updated immediately because of caching which is 24 hours according to @actlikewill. Can we reduce the time the files are updated?
Another issue that might be a bug is, replacing docx files with pdfs.
When we clear the content HTML the PDF file is not visible. The page is empty. We realized that an editor has to go back and clear the contents in the source tab. Do you think this can be done automatically?

longhotsummer:
> Another issue that might be a bug is, replacing docx files with pdfs.
> When we clear the content HTML the PDF file is not visible. The page is empty. We realized that an editor has to go back and clear the contents in the source tab. Do you think this can be done automatically?

@actlikewill is fixing this now

longhotsummer:
> We have determined that the source file is not updated immediately because of caching which is 24 hours according to @actlikewill. Can we reduce the time the files are updated?

I suggest we reduce this to the 30 minutes that general pages on the site are cached *AND* disable it for staff users, which also matches the logic for the site-wide cache.

This will mean that editors/uploaders immediately see the changes, and general users will pick up the change after at most 30 minutes.

longhotsummer:
> > We have determined that the source file is not updated immediately because of caching which is 24 hours according to @actlikewill. Can we reduce the time the files are updated?
> 
> I suggest we reduce this to the 30 minutes that general pages on the site are cached _AND_ disable it for staff users, which also matches the logic for the site-wide cache.
> 
[...]

PR #1760, opened 2024-03-05
Update source file details on file change

actlikewill:
- This allows us to update the source file meta data when the file is changed in the admin. 

#1737

### DIFF
--- a/peachjam/admin.py
+++ b/peachjam/admin.py
@@ -322,11 +322,6 @@ class DocumentForm(forms.ModelForm):
             return self.instance.content_html
         return self.cleaned_data["content_html"]
 
-    def _save_m2m(self):
-        super()._save_m2m()
-        # update document text
-        self.instance.update_text_content()
-
 
 class AttachedFilesInline(BaseAttachmentFileInline):
     model = AttachedFiles
@@ -504,7 +499,9 @@ class DocumentAdmin(BaseAdmin):
             cp.queue_re_extract_citations(form.instance.date)
 
         super().save_related(request, form, formsets, change)
+
         form.instance.save()
+        form.instance.update_text_content()
 
     def get_urls(self):
         return [
--- a/peachjam/forms.py
+++ b/peachjam/forms.py
@@ -188,6 +188,15 @@ class SourceFileForm(AttachmentFormMixin, forms.ModelForm):
                 self.instance.file_as_pdf.delete()
                 self.instance.ensure_file_as_pdf()
 
+    def clean(self):
+        cleaned_data = super().clean()
+        if "file" in self.changed_data:
+            # clear all the fields that are derived from the file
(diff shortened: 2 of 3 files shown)


peachjam/admin.py
```diff
@@ -322,11 +322,6 @@ def clean_content_html(self):
             return self.instance.content_html
         return self.cleaned_data["content_html"]
 
-    def _save_m2m(self):
-        super()._save_m2m()
-        # update document text
-        self.instance.update_text_content()
-
```
actlikewill, line 329: @longhotsummer this is getting called before the source file is saved and updated so we can move it to after the save_related method

peachjam/models/core_document_model.py
```diff
@@ -711,13 +711,10 @@ class Meta:
         abstract = True
 
     def save(self, *args, **kwargs):
-        if not self.filename:
-            self.filename = self.file.name
-        if not self.size:
-            self.size = self.file.size
-        if not self.mimetype:
-            self.file.seek(0)
-            self.mimetype = magic.from_buffer(self.file.read(), mime=True)
```
actlikewill, line 720: this prevents us from updating the source file details

longhotsummer: This messes with setting these fields explicitly. Instead, clear them in the admin form so that they get reset here.

longhotsummer (changes requested):
Small change needed

peachjam/forms.py
```diff
@@ -188,6 +188,15 @@ def _save_m2m(self):
                 self.instance.file_as_pdf.delete()
                 self.instance.ensure_file_as_pdf()
 
+    def clean(self):
+        cleaned_data = super().clean()
+        if "file" in self.changed_data:
+            # clear all the fields that are derived from the file
+            for field in ["mimetype", "size", "file_as_pdf", "filename"]:
+                self.instance.__dict__[field] = None
```
longhotsummer, line 196: `setattr` is the preferred method: `setattr(self.instance, field, None)`

longhotsummer: that is exactly what `self.instance.field = None` does

peachjam/forms.py
```diff
@@ -188,6 +188,15 @@ def _save_m2m(self):
...
+
+        return cleaned_data
```
actlikewill, line 198: @longhotsummer clearing the fields on the instance from the clean method allows us to compute them again on save.

peachjam/forms.py
longhotsummer: or just do it explicitly: `self.instance.mimetype = None` etc. :)

actlikewill: thanks, updated

longhotsummer (approved):

PR #1771, opened 2024-03-13
Remove extra call to update text content

### LINKED ISSUE #1737: As an editor uploading judgments, I want to see the correct document when I make edits to a file.

actlikewill:
- This removes the extra call to updating text content.
- Since the SourceFile form also calls update text content, we dont need to call it from the admin class.
- We need to update the filename when updating the source file,
https://www.loom.com/share/3b21adb696284b599ba21d7934d46e2d

closes #1737


peachjam/forms.py
```diff
@@ -169,6 +169,10 @@ def save(self, commit=True):
         if "file" in self.changed_data:
             self.instance.size = None
             self.instance.mimetype = None
+            ext = splitext(self.instance.file.name)[1]
```
longhotsummer, line 172: I think it's useful to keep the uploaded file's filename, because that allows editors to see what was uploaded from their records. This filename is never used externally.

longhotsummer:
Please also add this so that we reset the pdf when reading, otherwise it fails in weird and silent ways.

```
diff --git a/peachjam/models/core_document_model.py b/peachjam/models/core_document_model.py
index 11994602..323488d5 100644
--- a/peachjam/models/core_document_model.py
+++ b/peachjam/models/core_document_model.py
@@ -916,6 +916,7 @@ class DocumentContent(models.Model):
                     pdf = document.source_file.as_pdf()
                     if pdf:
                         shutil.copyfileobj(pdf, tmp)
+                        pdf.seek(0)
                         tmp.flush()
                         text = pdfjs_to_text(tmp.name)
                         # some PDFs have nulls, which breaks SQL insertion
```

peachjam/forms.py
actlikewill: ok, we also change it here, in the process upload method, should we be keeping the original file.name?

https://github.com/laws-africa/peachjam/blob/d362a9ece67bb8acebd872c778505eccfcb9c9af/peachjam/forms.py#L47-L51

peachjam/forms.py
```diff
@@ -46,9 +46,8 @@ def process_upload_file(self, upload_file):
         file_ext = splitext(upload_file.name)[1]
         SourceFile(
             document=self.instance,
-            file=File(
-                upload_file, name=f"{slugify(self.instance.title[-250:])}{file_ext}"
-            ),
+            file=File(upload_file, name=upload_file.name),
+            filename=f"{slugify(self.instance.title[-250:])}{file_ext}",
```
longhotsummer, line 50: I think we should drop this line, if we want to preserve the original filename?

peachjam/forms.py
longhotsummer: I think we should drop this line, if we want to preserve the original filename?

peachjam/forms.py
```diff
@@ -43,12 +41,10 @@ def _save_m2m(self):
...
         # store the uploaded file
         upload_file.seek(0)
-        file_ext = splitext(upload_file.name)[1]
         SourceFile(
             document=self.instance,
-            file=File(
-                upload_file, name=f"{slugify(self.instance.title[-250:])}{file_ext}"
-            ),
+            file=File(upload_file, name=upload_file.name),
+            filename=upload_file.name,
```
actlikewill, line 47: @longhotsummer this now allows us to keep the original filename

longhotsummer (approved):
great

PR #1772, opened 2024-03-13
Change matomo site search tracking to use trackSiteSearch

### LINKED ISSUE #1757: Change matomo site search tracking to use trackSiteSearch
https://developer.matomo.org/guides/tracking-javascript-guide#internal-search-tracking

This will make it possible to identify searches with zero results, and to track the major content type facet (which is the "category" in the API).

Sandravaphilips:
Since the URL params contain both keywords and facets, I figured the easiest way to track would be to use the entire params as the keyword like this:
` 
      this.paq.push(['trackSiteSearch', keyword, false, searchCount]);
`
If you'd rather I destructure it instead, please let me know.
Closes #1757

### DIFF
--- a/peachjam/js/components/FindDocuments/index.vue
+++ b/peachjam/js/components/FindDocuments/index.vue
@@ -659,17 +659,16 @@ export default {
 
     trackSearch (params) {
       const keywords = [];
+      const facets = [];
+      const fields = this.facets.map(facet => facet.name).concat(['date__range', 'date__gte', 'date__lte']);
+
       params.forEach((value, key) => {
         if (key.startsWith('search')) {
           const s = key === 'search' ? '' : (key.substring(8) + '=');
-          keywords.push(s + value.strip());
+          keywords.push(s + value.trim());
+        } else if (fields.includes(key)) {
+          facets.push(`${key}=${params.getAll(key).join(',')}`);
         }
-      });
-
-      const facets = [];
-      const fields = this.facets.map(f => f.name).concat('date__range', 'date__gte', 'date__lte');
-      fields.forEach(facet => {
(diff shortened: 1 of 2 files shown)


peachjam/js/components/FindDocuments/index.vue
```diff
@@ -639,6 +640,9 @@ export default {
             if (response.ok) {
               this.error = null;
               this.searchInfo = await response.json();
+              if (searchParams) {
+                analytics.trackSiteSearch(searchParams, this.searchInfo.count);
```
longhotsummer, line 644: Matomo lets us push three things:

* the search term
* a search category (optional)
* number of hits

Providing the full searchParams string is too overwhelming. Instead, let's do this:

* search term: use `q` for simple searches, for advanced searches provide all the `search_*` fields; eg `search__title=x&search__content=-y`
* category: use the document type and court facets, if in use, separated by semicolons; eg. `Judgment; High Court` or `High Court` or `Judgment`

Using `params` before it is turned into a string should make this a bit easier.

For GA, just include whatever it supports (probably just the search term)

longhotsummer:
Ok, this is looking much better. Please sanity check that date facets get pulled into the reporting, otherwise I think it's good to go.

longhotsummer (dismissed):
great!

peachjam/js/components/FindDocuments/index.vue
```diff
@@ -662,11 +662,13 @@ export default {
       const facets = [];
       const fields = this.facets.map(facet => facet.name).concat(['date__range', 'date__gte', 'date__lte']);
 
+      let currentKey = '';
```
longhotsummer, line 665: A better way of doing this is iterating over the keys only, not the key/value pairs. Then you don't need to de-dup like this.

longhotsummer (approved):

PR #1775, opened 2024-03-15
Admin tests

actlikewill:
- this adds tests for the judgment admin


actlikewill:
@longhotsummer these tests are already revealing some issues. There is some inconsistency with the way we set mimetypes for source documents. When adding new documents, we read from file.content_type, when we are modifying, we use the python magic library to read the mimetype. These two dont always give the same results. It could also be because of the way Webtest is mocking the files but its making the tests that I'm writing fail.

longhotsummer:
> @longhotsummer these tests are already revealing some issues. There is some inconsistency with the way we set mimetypes for source documents. When adding new documents, to set the mimetype, we read from file.content_type but when we are modifying, we use the python magic library. These two dont always give the same results. It could also be because of the way Webtest is mocking the files but its making the tests that I'm writing fail.

This is good to learn. We `file.content_type` trusts the browser, whereas the python magic library actually looks at the file contents. So `b"pdf judgment content"` won't be considered a PDF by the magic library.

An option is to add some small PDF and DOCX files as fixture objects and post real files.

actlikewill:
> An option is to add some small PDF and DOCX files as fixture objects and post real files.

@longhotsummer I've tried this but it doesn't seem to work properly. The webtes.Upload doesn't seem to work with actual files. Still checking though

peachjam/tests/test_admin.py
```diff
@@ -0,0 +1,123 @@
...
+        form2["source_file-0-file"] = Upload(
+            "file.docx",
+            b"test case file judgment details",
+            "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+        )
+        response2 = form2.submit()
+        self.assertRedirects(response2, judgment_list_url)
+
+        judgment.refresh_from_db()
+        self.assertIn("test case", judgment.content_html)
```
actlikewill, line 120: @longhotsummer this is the assertion that revealed the issue. I'm checking that the content_html has been changed, but since the mimetype is incorrect, it is not extracting the content. The mimetype is reported as `text/plain`

peachjam/tests/test_admin.py
```diff
@@ -0,0 +1,155 @@
...
+        # date uses a multi-field widget
+        form["date_0"] = "21"
+        form["date_1"] = "2"
+        form["date_2"] = "2000"
+
+        with open(
+            os.path.abspath("peachjam/fixtures/source_files/zagpjhc_judgment.docx"),
+            "rb",
+        ) as docx_file:
+            docx_file_content = docx_file.read()
```
actlikewill, line 38: @longhotsummer found a hack to pass file contents to the webtest

actlikewill:
@longhotsummer this pr will merge to this one #1771 with the bug fix, the we can merge it to master

longhotsummer (approved):

Issue #1774, opened 2024-03-14
As an administrator, I want to add a new "Contact Us" Google Form for MalawiLII and ZambiaLII


niiroobiro:
* [x] check with recipients that they can receive google form emails
* [x] create google forms
* [x] add a peachjam setting for a form embed url, and use that in the contact us page template

niiroobiro:
Users receiving this;
For MalawiLII: victoriachindamba@gmail.com and tinnanamadingo10@gmail.com

For ZambiaLII: mwamikabwabwa@gmail.com and kasomamaria.mulenga@gmail.com

niiroobiro:
Malawi form url: https://forms.gle/kzBkqDex9n1yf7sN9
Zambia form url: https://forms.gle/NugvQnT4L26riDZ98

PR #1780, opened 2024-03-19
As an administrator, I want to add a new "Contact Us" Google Form for MalawiLII and ZambiaLII

Sandravaphilips:
added a peachjam setting for a form embed url, and used that in the contact us page template
#1774


zambialii/templates/peachjam/about.html
```diff
@@ -121,6 +121,16 @@ <h2 class="mb-4" id="contact">{% trans 'Contact us' %}</h2>
         </address>
       {% endblocktrans %}
       </p>
+      {% if PEACHJAM_SETTINGS.contact_form_url %}
```
longhotsummer, line 124: Can you move this into peachjam's about.html, so that it just works on all sites? You may need to introduce a new block into the base template to put it into. So lawlibrary, zambialii and malawilii would no longer need special treatment.

longhotsummer:
Great - can you include a screenshot or two to demo the changes?

Sandravaphilips:
@longhotsummer 
laws.africa
Screenshot 2024-03-22 at 14 04 21
africanlii
Screenshot 2024-03-22 at 14 06 56
Screenshot 2024-03-22 at 14 07 09

longhotsummer (approved):

PR #1787, opened 2024-04-05
Add an extra phrase-match SHOULD search for document content

### LINKED ISSUE #1758: Search: ensure that docs with pages don't score higher than docs with plain text content
We add a SHOULD search to the page (child documents) search for PDF docs, but not for simple `content`-driven docs. This means that PDF docs get boosted.

We should be adding the extra SHOULD to the `content` field search as well.

longhotsummer:
1. move away from the weird class-based query building to just building the query in one place, this makes things much simpler to change and understand
2. add a SHOULD PhraseMatch on `content` to match the same thing on `pages.body` (the latter only applies to PDFs). This means we don't penalise non-PDFs any more. 

Closes #1758 

https://www.loom.com/share/013874e5471e4383a29607fdc1af6ede

### DIFF
--- a/peachjam_search/views.py
+++ b/peachjam_search/views.py
@@ -9,16 +9,14 @@ from django.views.decorators.cache import cache_page
 from django.views.decorators.vary import vary_on_cookie
 from django.views.generic import TemplateView
 from django_elasticsearch_dsl_drf.filter_backends import (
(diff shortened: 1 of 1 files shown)


actlikewill (approved):
great. i'm always keen on how we tweak elasticsearch to work better

PR #1788, opened 2024-04-05
don't lose embedded /akn citations from markdown docs

longhotsummer:
when re-extracting citations from markdown docs, we don't want to lose any existing /akn citations that may already be in place.

in time, we probably want to guard against this in all document types, but having /akn refs in non-markdown (and non-AKN) docs is very rare.

This allows our law reader books to have /akn refs baked into them.


actlikewill (approved):
interesting

PR #1792, opened 2024-04-10
Auto assign title field

actlikewill:
- This adds the auto assign title flag, to control whether the system generates a title for judgments


actlikewill:
@longhotsummer this is how I propose we should control whether to assign titles in the judgments, to allow KL to retain titles coming from Tausi

peachjam/models/judgment.py
```diff
@@ -261,6 +261,13 @@ class Judgment(CoreDocument):
     )
     hearing_date = models.DateField(null=True, blank=True)
 
+    auto_assign_title = models.BooleanField(
+        help_text=_("Whether or not the system should assign the title"),
+        null=True,
+        blank=True,
```
longhotsummer, line 267: drop both of these, so it must be either True or False, with a default of True. It'll backfill with True.

peachjam/models/judgment.py
```diff
@@ -261,6 +261,13 @@ class Judgment(CoreDocument):
     )
     hearing_date = models.DateField(null=True, blank=True)
 
+    auto_assign_title = models.BooleanField(
```
longhotsummer, line 264: it needs a translatable field title, "Auto-assign title"

longhotsummer:
looks good, simplest way for now.

longhotsummer (approved):

Issue #1759, opened 2024-03-05
Search: index AKN document provisions separately; boost title matches; show in results


longhotsummer:
We can do a much better job of showing provision-based results for AKN documents.

* match on titles and provision content
* boost title matches
* show TOC hierarchy when showing a provision match
* make the TOC items clickable

longhotsummer:
# Problem

Consider a search like "children court attendance". On LawLibrary.org.za, we have results like this (note the scores):

![image](https://github.com/laws-africa/peachjam/assets/4178542/3cd9abcc-069e-4c5e-aa55-3c46154528f6)

After the first three matches, the scores drop drastically (below 40). The Children's Act is not in the first page of results, even when the search is restricted to legislation. This is despite the fact that there is a section related to this:  

![image](https://github.com/laws-africa/peachjam/assets/4178542/be4b61cc-3a8f-4f00-9492-fe663ef9c957)

The part of the problem is that headings are treated as regular text when indexing AKN documents. We can do better.

# Proposal

The idea is to take advantage of the AKN Table of Contents (TOC) and use that in smart ways, in particular:

1. apply search terms to the TOC when searching
2. display the TOC to provide context for legislation search result snippets
3. make the TOC context clickable to let the user jump directly to the relevant portion of the document (like we do with page numbers in PDF results)

For example, for the search "children court attendance" you could imagine a search result like this:

```
Children's Act
  2023-11-08 Legislation
  Act 38 of 2005
    Chapter 4 - Children's Courts
      Part 2 - Court proceedings
        56. Attendance at proceedings
          ... children's court ... attended ...
```

1. the TOC itself is searched section 56 is matched because it and its parents have a strong match to the keywords
2. the content of section 56 is also matched because it has a strong match
3. the content of section 56 is highlighted and the snippet is shown with the context of the TOC above it
4. the TOC elements are clickable, not just the Children's Act title

longhotsummer:
# Indexing

When indexing AKN legislation content, we don't index the text as a big chunk. Instead, we drive indexing through the TOC, a bit like we do when indexing PDFs pages. So we'll index provisions instead.

1. traverse the TOC without going below `basic_unit: True`.
2. for each element in the TOC, index that as a `provision` sub-document in elasticsearch.
3. a provision is indexed with:
  * its TOC title (eg. "Part 2 - Court proceedings")
  * its provision eId (eg. `chp_4__part_2`) -- not searched
[...]

PR #1801, opened 2024-04-19
Search legislation TOC

longhotsummer:
Closes https://github.com/laws-africa/peachjam/issues/1759

* make TOC searchable
* show TOC in search hits for legislation
* use cards for search hits to better visually separate hits; this reduces the amount of indentation needed

https://www.loom.com/share/04d48383ebbe44299f2548f4967f282d

image

## Note

~This requires re-creating index mappings and re-indexing all content :(~

This automatically adds the new field to the mapping as a migration. We can then re-index legislation content only, at our leisure, without breaking search at all.


peachjam_search/migrations/0001_es_mapping_add_provisions.py
```diff
@@ -0,0 +1,29 @@
...
+
+from django.db import migrations
+
+
+def forwards(apps, schema_editor):
+    from django.conf import settings
+    from django_elasticsearch_dsl.registries import registry
+
+    if not settings.DEBUG:
+        for ix in registry.get_indices():
```
longhotsummer, line 11: this allows us to add the new field to the mapping (which is allowed without having to destroy and re-create the mapping).

we only do it when DEBUG = False because we don't have an ES to talk to when testing.

actlikewill: neat

actlikewill (approved):
really cool. It really improves legislation search

PR #1806, opened 2024-04-29
pins allauth version 0.56.0

actlikewill:


longhotsummer (changes requested):
needs to be 0.62.0 or greater i think

longhotsummer (approved):

PR #1811, opened 2024-05-08
Track searches

longhotsummer:
* track searches made, and keep for 90 days
* track clicks on search results
* open search result provisions and pages in new tabs

This makes it much easier for us to understand how users are using search.


actlikewill:
@longhotsummer I'm curious, is this not information that matomo provides? what dimension of data are we interested in that matomo is not tracking?

longhotsummer:
> @longhotsummer I'm curious, is this not information that matomo provides? what dimension of data are we interested in that matomo is not tracking?

Matomo doesn't let us slice and dice the searches, as far as I can tell. So we can't easily see how a user refined a search (eg. perform a search, add filter facets) and then which item they clicked on. We also can't easily see all the `doctype=judgment` searches.

actlikewill (approved):

PR #1813, opened 2024-05-14
Advanced search fixes

### LINKED ISSUE #1795: Advanced search:  Investigate and Implement all keywords option for cross-field search
A user mentioned that the option to search "for all these keywords" would be useful. For multi-keyword searches, we require a majority of keywords to match, but not all.

This differs a bit from "exact phrase" search.

longhotsummer:
* build the search query field-by-field, so that it's correct
* don't mix simple and advanced search for "all" field queries; so advanced search now has `search__all`; this means we can apply stricter match requirements
* adjust field chooser so that the user can't choose no fields, or all + other fields
* advanced search on the "content" field must search content, pages and provisions 

Closes #1795 

https://www.loom.com/share/e0e891516fdd488cab852e9ca663c387

There is still a problem with negative queries on pages or legislation with provisions. Because each provision/page is indexed and search individually, it's possible to have a search `dog -cat` that returns a document with `cat` in it, because it'll match pages or provisions that have `dog` but not `cat`.

So for negative searches we actually need to index and search the full content, not on a per-page or per-provision basis.

### DIFF
--- a/peachjam_search/views.py
+++ b/peachjam_search/views.py
@@ -129,7 +129,7 @@ class MainSearchBackend(BaseSearchFilterBackend):
         for field in view.search_fields.keys():
             if field == "content":
                 # advanced search on the "content" field (which must include pages and provisions too), is handled
-                # by the
+                # by build_advanced_content_queries
                 continue
             query = request.query_params.get(self.search_param + "__" + field)
             if query:


longhotsummer:
I've opened https://github.com/laws-africa/peachjam/issues/1815 for the negative search issue

actlikewill (approved):
awesome. works well
I think this also closes laws-africa/kenyalaw-pj#62

PR #1814, opened 2024-05-14
Citations

longhotsummer:
https://www.loom.com/share/5ca9c3ec44ca45779abfaa9050977876

* show citation counts in the citations tab
* pre-calculate citation counts for works
* group related documents by predicate

image

image


peachjam/models/core_document_model.py
```diff
@@ -169,6 +169,10 @@ class Work(models.Model):
     )
     # the rank (weight) of this work in the graph network, computer by peachjam.graph.ranker
     ranking = models.FloatField(_("ranking"), null=True, blank=False, default=0.0)
+    # number of outgoing citations
+    n_cited_works = models.IntegerField(_("number of cited works"), default=0)
+    # number of incoming citations
+    n_citing_works = models.IntegerField(_("number of incoming citations"), default=0)
```
actlikewill, line 175: I was worried about this slowing down the pages with many citing documents but this solves that :)

actlikewill (approved):
Great. I was considering splitting the citations tabs (outgoing and incoming) like we discussed as part of the KL improvements. We can discuss whether this is necessary.

longhotsummer:
> Great. I was considering splitting the citations tabs (outgoing and incoming) like we discussed as part of the KL improvements. We can discuss whether this is necessary.

Yes, I think in future we probably will.

PR #1816, opened 2024-05-15
Split order and outcome fields

actlikewill:
- This splits the order and outcome fields.
- We now have the order as a text field on the judgment model
- Renamed OrderOutcome to Order and added migrations for moving the data
- After merging, we can remove the OrderOutcome model.

closes https://github.com/laws-africa/kenyalaw-pj/issues/57


actlikewill:
@longhotsummer Since we have a new field, we need to rebuild the index. I think there was a way to do this without rebuilding the entire indexes?

peachjam/migrations/0132_backfill_outcomes.py
```diff
@@ -0,0 +1,41 @@
...
+            o.name_en = outcome.name_en
+        if hasattr(outcome, "name_fr"):
+            o.name_fr = outcome.name_fr
+        if hasattr(outcome, "name_pt"):
+            o.name_pt = outcome.name_pt
+        if hasattr(outcome, "name_sw"):
+            o.name_sw = outcome.name_sw
+        o.save()
+
+    for judgment in Judgment.objects.all().iterator(chunk_size=100):
```
longhotsummer, line 25:
```suggestion
    for judgment in Judgment.objects.filter(order_outcomes__exist=True).order_by('pk').distinct('pk').iterator(chunk_size=100):
```

this excludes anything that doesn't have an order_outcome

peachjam/migrations/0132_backfill_outcomes.py
```diff
@@ -0,0 +1,41 @@
...
+        j_outcomes = judgment.order_outcomes.all().values_list("name", flat=True)
+        outcomes = Outcome.objects.filter(name__in=j_outcomes)
+        if outcomes:
+            for outcome in outcomes:
```
longhotsummer, line 29: What about `judgment.outcomes.set(outcomes)`?

longhotsummer:
> @longhotsummer Since we have a new field, we need to rebuild the index. I think there was a way to do this without rebuilding the entire indexes?

we can use a technique like this to add the new fields (will have to do it for each translated field): https://github.com/laws-africa/peachjam/blob/main/peachjam_search/migrations/0001_es_mapping_add_provisions.py#L10

We'll definitely have to test locally.

peachjam/migrations/0132_backfill_outcomes.py
actlikewill: changed

peachjam_search/serializers.py
```diff
@@ -105,10 +105,12 @@ def get_nature(self, obj):
         return obj["nature" + self.language_suffix]
 
     def get_outcome(self, obj):
-        val = obj["outcome" + self.language_suffix]
-        if val is not None:
-            val = list(val)
-        return val
+        if hasattr(obj, "outcome" + self.language_suffix):
```
longhotsummer, line 108: Why would this have to be different to the ones above?

peachjam/migrations/0132_outcome_es_index.py
```diff
@@ -0,0 +1,37 @@
...
+import os
+
+from django.db import migrations
+
+
+def forwards(apps, schema_editor):
+    from django.conf import settings
+    from django_elasticsearch_dsl.registries import registry
+
+    # if not settings.DEBUG and os.environ.get("ELASTICSEARCH_HOST"):
```
longhotsummer, line 12: You'll need this to avoid failing tests

peachjam_search/serializers.py
actlikewill: The migration for mapping is not actually updating the indexed cocument object, only the mapping object. This causes a key error in the results when searching. I'm starting to think that we may have to rebuild the index :( 
Even when resaving a document, the indexed document is not picking up the new field. only rebuilding the index fully works. I may be missing something. How did you get provisions to work?

peachjam/migrations/0132_outcome_es_index.py
```diff
@@ -0,0 +1,37 @@
...
+    for ix in registry.get_indices():
+        if not ix._mapping:
+            continue
+        print(f"Adding outcome mapping for {ix._name}")
+        ix.connection.indices.put_mapping(
+            index=ix._name,
+            body={
+                "properties": {
```
actlikewill, line 20: @longhotsummer I also noticed that your code did not include the "properties" keyword. I get errors when I omit it

peachjam_search/serializers.py
longhotsummer: makes sense

longhotsummer (dismissed):

actlikewill:
> > @longhotsummer Since we have a new field, we need to rebuild the index. I think there was a way to do this without rebuilding the entire indexes?
> 
> we can use a technique like this to add the new fields (will have to do it for each translated field): https://github.com/laws-africa/peachjam/blob/main/peachjam_search/migrations/0001_es_mapping_add_provisions.py#L10
> 
> We'll definitely have to test locally.

This is now working :sweat_drops:

longhotsummer (approved):

PR #1817, opened 2024-05-16
Download and edit buttons in detail page

longhotsummer:
* also fixes ordering of citation groups
* don't show jurisdiction unneccesarily

image


longhotsummer:
> @longhotsummer can we consider having the buttons outside the tab panels where they are always visible regardless of what tab we are on?, with this change, the edit button is in the document detail tab only

Great idea, here's what it looks like:

image

actlikewill (approved):
great

PR #1824, opened 2024-05-24
when counting citations, only count works with documents

longhotsummer:
otherwise the totals shown on the detail page is wrong, since we need documents to link to


actlikewill (approved):
looks good

PR #1830, opened 2024-05-28
Causelists in tanzlii header

actlikewill:
- Add causelist to tanzlii header

image


peachjam/views/books.py
```diff
@@ -7,6 +7,7 @@
 
 
 class BookListView(FilteredDocumentListView):
+    queryset = Book.objects.all()
```
actlikewill, line 10: just noticed that these listing pages were not working properly

niiroobiro (approved):

PR #1833, opened 2024-05-30
Case histories

actlikewill:
This adds support for case histories

image

If there is a historical judgment we show the title and link to it, otherwise we show a case number/citation with no link. We also have judges and court fields on the case history.

part of #https://github.com/laws-africa/kenyalaw-pj/issues/58


peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -421,6 +435,14 @@ <h5>
           <div class="container">{% include 'peachjam/_citations.html' %}</div>
         </div>
       {% endif %}
+      {% if document.case_histories.all|length %}
```
longhotsummer, line 438: It's better to call `.case_histories.exists()`, otherwise it loads all the data, counts the length, and then throws the data away. Using `.exists()` just checks for at least one row in the DB without loading any data.

peachjam/models/judgment.py
```diff
@@ -486,3 +486,42 @@ def get_case_number_string(self):
...
+        verbose_name=_("outcome"),
+    )
+    judges = models.ManyToManyField(Judge, verbose_name=_("judges"), blank=True)
+    court = models.ForeignKey(
+        Court, on_delete=models.PROTECT, null=True, blank=True, verbose_name=_("court")
+    )
+    date = models.DateField(_("date"), null=True, blank=True)
+
+    class Meta:
+        ordering = ["date"]
```
longhotsummer, line 522:
```suggestion
        ordering = ["-date"]
```

so it works as a timeline, most recent date first

peachjam/templates/peachjam/_case_histories.html
```diff
@@ -0,0 +1,35 @@
+{% load i18n %}
+<table class="table table-striped">
+  <thead>
+    <tr>
+      <th>{% trans 'Case' %}</th>
```
longhotsummer, line 5: Please put the date first, so that it works as a timeline going back in time.

Put the outcome last.

peachjam/templates/peachjam/_case_histories.html
```diff
@@ -0,0 +1,35 @@
...
+      <th>{% trans 'Outcome' %}</th>
+      <th>{% trans 'Court' %}</th>
+      <th>{% trans 'Judges' %}</th>
+      <th>{% trans 'Date' %}</th>
+    </tr>
+  </thead>
+  <tbody>
+    {% for history in document.case_histories.all %}
```
longhotsummer, line 13: It will actually be better to load this all in the view, so that the right fields are prefetched. Otherwise you're going to the DB for every history, and for every historical_judgment, and for every judge, which could be 5-10 queries.

peachjam/templates/peachjam/layouts/document_detail.html
actlikewill: ok. changed. We now load the case histories in from view with prefetched related fields

peachjam/templates/peachjam/_case_histories.html
actlikewill: done

peachjam/templates/peachjam/_case_histories.html
actlikewill: done

PR #1834, opened 2024-06-01
Responsible document table and improved legislation listing

longhotsummer:
* replace tables with responsive document tables that collapse on mobile
* legislation listing uses FRBR year, rather than date, and always has this column
* legislation listing shows group headings based on sort key
* legislation listing includes labels
* document tables make better use of whitespace, particularly on mobile, to show more data with less padding
* legislation table now looks the same as the other document tables

https://www.loom.com/share/aae4f89d15b74fe5b1bbc1756f545952

## desktop

image

image

image

## mobile

image

image


actlikewill (dismissed):
very great change. I like that we can standardize the listing changes.

PR #1835, opened 2024-06-01
fix direction of related judgments in admin view

longhotsummer:
* subject/object should be reversed
* also makes work autocomplete a bit broader

image

image

image

@NotoriousMBB I owe you a croissant


actlikewill (approved):
looks good, do we have do do a migration to correct any erroneous data

longhotsummer:
> looks good, do we have do do a migration to correct any erroneous data

We'll have to do it manually.

PR #1836, opened 2024-06-04
Gazette volume number and resource

actlikewill:
This adds the volume number and gazette bulk import resource


peachjam/models/gazette.py
```diff
@@ -20,6 +20,7 @@ class Gazette(CoreDocument):
     key = models.CharField(
         _("key"), max_length=512, null=True, blank=True, db_index=True
     )
+    volume_number = models.IntegerField(_("volume number"), null=True, blank=True)
```
longhotsummer, line 23: string field

actlikewill: changed

longhotsummer:
Looks good. 

Do you think we should add it to the default gazette listing pages, near where we put the sub-publication or number? maybe in the same td?

actlikewill:
> Looks good.
> 
> Do you think we should add it to the default gazette listing pages, near where we put the sub-publication or number? maybe in the same td?

ah right. here it is
image

longhotsummer (approved):

PR #1847, opened 2024-06-07
correctly handle background task errors in sentry

longhotsummer:
before this, the error was handled by background_task and never raised for sentry to handle.

instead, we now report it during background_task's signal handler


actlikewill (approved):
I'm curious, do you know what changed in our setup that caused this? because we used to see background errors before

PR #1879, opened 2024-06-29
don't re-call pushState after popState

### LINKED ISSUE #1799: on the search page, the browser back button does not work properly
- search for a document using search
- the search view loads
- apply a filter or two or three
- click back button a few times
- it only goes back once or twice
- it does not allow me to go back all the way

longhotsummer:
otherwise the back button doesn't work correctly.

fixes #1799

https://www.loom.com/share/e357738423ac4c638e33265e2a94d4bf


actlikewill (approved):
great

PR #1881, opened 2024-06-29
Minor layout changes to the homepage

longhotsummer:
* recent documents shown in columns to take up less space
* recent documents use new table mechanism
* taxonomies shown in three columns to take up less vertical space
* list courts with colour background just to add some variety
* increase padding between homepage sections
* better block names for homepage template
* remove preview warning banner for lawlibrary

image

image

localhost_8001_


actlikewill (approved):
This looks much better. Good use of available space.

Issue #1887, opened 2024-07-04
User can manage collections after saving a document


niiroobiro:
Allow users to manage their collections by adding, removing, editing, and deleting collections and documents.

Issue #1690, opened 2024-01-18
As a User, I want to be able to save documents I find interesting on the site.


niiroobiro:

niiroobiro:
Here is the link to the proposal:
`https://docs.google.com/document/d/1UHUXtD5lUf9TCMibt0pgv1BFPIeDQixOEPmJDgfZQbg/edit?usp=sharing`

niiroobiro:
@longhotsummer I have made edits to the feature proposal. Let me know if it still needs some more work.

longhotsummer:
This is ready to implement.

PR #1892, opened 2024-07-05
Save documents

Sandravaphilips:
closes https://github.com/laws-africa/peachjam/issues/1690
closes https://github.com/laws-africa/peachjam/issues/1918
closes https://github.com/laws-africa/peachjam/issues/1887


peachjam/templates/peachjam/layouts/base.html
```diff
@@ -62,7 +62,8 @@
     {% block head-css %}{% endblock %}
     {% block head-js %}{% endblock %}
   </head>
-  <body class="{% block body-class %}{% endblock %}">
+  <body hx-headers='{"X-CSRFToken": "{{ csrf_token }}"}'
+        class="{% block body-class %}{% endblock %}">
```
actlikewill, line 66: @Sandravaphilips is this necessary? if the csrf token is in the forms we post it should work, or it can be added to the button that makes the request. keeping the behaviour local

peachjam/views/save_document.py
```diff
@@ -0,0 +1,91 @@
...
+        user_profile = get_object_or_404(UserProfile, user=self.request.user)
+        folders = user_profile.folders.all()
+        context["folders"] = folders
+        return context
+
+    def post(self, request: HttpRequest):
+        post_data = request.POST
+        instance = SavedDocument.objects.filter(
+            document=post_data.get("document"),
+            user_profile=post_data.get("user_profile"),
```
longhotsummer, line 56: the user information must come from `request.user`

peachjam/js/components/DocumentProblemModal.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="modal-dialog">
-    <div class="modal-content">
+    <div id="documentDetailModal" class="modal-content">
```
longhotsummer, line 3: revert?

peachjam/models/save_document.py
```diff
@@ -0,0 +1,45 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from .core_document_model import CoreDocument
+
+
+class Folder(models.Model):
+    name = models.CharField(_("name"), max_length=1024)
+    user_profile = models.ForeignKey(
```
longhotsummer, line 9: change to user

peachjam/models/save_document.py
```diff
@@ -0,0 +1,45 @@
...
+        verbose_name_plural = _("folders")
+        unique_together = ("name", "user_profile")
+
+    def __str__(self):
+        return f"{self.name}"
+
+
+class SavedDocument(models.Model):
+    document = models.ForeignKey(CoreDocument, on_delete=models.CASCADE)
+    user_profile = models.ForeignKey(
```
longhotsummer, line 28: change to FK to user

peachjam/templates/peachjam/_document_detail_toolbar.html
```diff
@@ -9,6 +9,61 @@
...
+                <div class="modal-header">
+                  <h5 class="modal-title">Login</h5>
+                  <button type="button"
+                          class="btn-close"
+                          data-bs-dismiss="modal"
+                          aria-label="Close"></button>
+                </div>
+                <div class="modal-body">
+                  <p>
+                    <strong>You are not logged in.</strong>
```
longhotsummer, line 52: Translateable

peachjam/templates/peachjam/_header.html
```diff
@@ -54,13 +54,18 @@
                 </a>
                 <ul class="dropdown-menu">
                   <li>
-                    <a class="dropdown-item" href="{% url 'admin:index' %}">{% trans 'Admin' %}</a>
+                    {% if user.is_superuser %}
+                      <a class="dropdown-item" href="{% url 'admin:index' %}">{% trans 'Admin' %}</a>
+                    {% endif %}
                     <a class="dropdown-item" href="{% url 'account_logout' %}">{% trans 'Logout' %}</a>
+                    <a class="dropdown-item" href="">{% trans 'Dashboard' %}</a>
```
longhotsummer, line 61: url

peachjam/templates/peachjam/folders_list.html
```diff
@@ -0,0 +1,33 @@
+<div class="modal-dialog" id="testing">
+  <div class="modal-content">
+    <div class="modal-header">
+      <p class="h5 modal-title">Add new folder</p>
```
longhotsummer, line 4: translateable

peachjam/templates/peachjam/save_document.html
```diff
@@ -0,0 +1,89 @@
+<div class="modal-dialog" id="testing">
+  <div class="modal-content">
+    <div class="modal-header">
+      <p class="h5 modal-title">Choose Folder</p>
```
longhotsummer, line 4: translateable

peachjam/templates/peachjam/save_document.html
```diff
@@ -0,0 +1,89 @@
...
+      <form id="save_document-form"
+            hx-post="{% url 'save_document' %}"
+            hx-target="#saveDocument">
+        <label class="form-label"
+               for="{{ save_document_form.document.id_for_label }}">
+          {{ save_document_form.document.label }}:
+        </label>
+        <p class="h5">{{ document }}</p>
+        {{ save_document_form.document }}
+        {{ save_document_form.user_profile }}
```
longhotsummer, line 20: bad :)

peachjam/urls.py
```diff
@@ -90,6 +92,8 @@
         AboutPageView.as_view(),
         name="about",
     ),
+    path("folders/", CreateFolderView.as_view(), name="folders"),
+    path("save-document/", SaveDocumentView.as_view(), name="save_document"),
```
longhotsummer, line 96: saved-documents
saved-documents/<int:pk>
saved-documents/<int:pk>/delete

peachjam/views/generic_views.py
```diff
@@ -240,6 +243,13 @@ def get_object(self, *args, **kwargs):
             self.model, expression_frbr_uri=add_slash(self.kwargs.get("frbr_uri"))
         )
 
+    def show_save_doc_button(self):
+        if pj_settings().allow_save_documents and self.request.user.has_perm(
```
longhotsummer, line 247: `return ...`

peachjam/views/save_document.py
actlikewill: moved the user logic to the server

peachjam/models/save_document.py
actlikewill: done

peachjam/models/save_document.py
actlikewill: done

peachjam/templates/peachjam/_document_detail_toolbar.html
actlikewill: done

peachjam/templates/peachjam/_header.html
actlikewill: changed

peachjam/templates/peachjam/folders_list.html
actlikewill: done

peachjam/templates/peachjam/save_document.html
actlikewill: done

peachjam/templates/peachjam/save_document.html
actlikewill: updated

peachjam/urls.py
actlikewill: we split up the views to be more inline with this pattern, code feels much simpler

peachjam/views/generic_views.py
actlikewill: changed

peachjam/forms.py
```diff
@@ -301,3 +314,46 @@ def send_email(self):
...
+        if self.cleaned_data.get("new_folder"):
+            folder, _ = Folder.objects.get_or_create(
+                name=self.cleaned_data["new_folder"],
+                user=self.cleaned_data["user"],
+            )
+            self.cleaned_data["folder"] = folder
+        return self.cleaned_data
+
+    def save(self, commit=True):
+        return super().save()
```
longhotsummer, line 351: this can be dropped

peachjam/settings.py
```diff
@@ -151,16 +153,20 @@
...
 # email addresses are required for new accounts
 ACCOUNT_EMAIL_REQUIRED = True
 ACCOUNT_PRESERVE_USERNAME_CASING = False
 ACCOUNT_SESSION_REMEMBER = True
 ACCOUNT_EMAIL_SUBJECT_PREFIX = EMAIL_SUBJECT_PREFIX
-ACCOUNT_DEFAULT_HTTP_PROTOCOL = "https"
+ACCOUNT_DEFAULT_HTTP_PROTOCOL = "http" if DEBUG else "https"
 LOGIN_URL = "account_login"
 LOGIN_REDIRECT_URL = "home_page"
+ACCOUNT_EMAIL_VERIFICATION = "mandatory"
```
longhotsummer, line 164: This may be a problem, because it forces all accounts to be verified before a user can log in, including admins.

peachjam/signals.py
```diff
@@ -55,3 +65,22 @@ def extracted_citation_deleted(sender, instance, **kwargs):
     """Update citation counts on works."""
     ExtractedCitation.update_counts_for_work(instance.citing_work)
     ExtractedCitation.update_counts_for_work(instance.target_work)
+
+
+@receiver(signals.post_save, sender=User)
+def add_saved_document_permissions(sender, instance, created, **kwargs):
+    if pj_settings().allow_save_documents and created:
+        permissions = Permission.objects.filter(
```
longhotsummer, line 73: It's much better to do this with a group, because then the perms can be added or removed on a group basis. Otherwise if we want to disable this permission, we have to go through each user and remove the perm. So we'll need to create a default group (perhaps "All users", it can be hardcoded) and then ensure all new users belong to that group. It's probably easiest to get_or_create the Group object so that if it's deleted, it's always re-created.

peachjam/templates/peachjam/save_document.html
```diff
@@ -0,0 +1,100 @@
...
+  <div class="modal-content">
+    <div class="modal-header">
+      <p class="h5 modal-title">{% trans 'Save document' %}</p>
+      <button type="button"
+              class="btn-close"
+              data-bs-dismiss="modal"
+              aria-label="Close"></button>
+    </div>
+    <div class="modal-body">
+      {% if updated %}<div class="alert alert-success">Document saved</div>{% endif %}
```
longhotsummer, line 12: Translateable

peachjam/templates/peachjam/save_document.html
```diff
@@ -0,0 +1,100 @@
...
+    <div class="modal-footer d-flex justify-content-between">
+      {% if saved_document %}
+        <button class="btn btn-danger"
+                type="button"
+                data-bs-dismiss="modal"
+                hx-target="#saveDocument"
+                hx-include="#save_document-form"
+                hx-headers='{"X-CSRFToken": "{{ csrf_token }}"}'
+                hx-post="{% url 'unsave_document' saved_document.id %}"
+                hx-confirm="Are you sure you want to unsave this document?">
```
longhotsummer, line 63: Translateable

peachjam/settings.py
actlikewill: should we disable this then?

actlikewill: or force admins to verify themselves

actlikewill: so admins will still be able to login to the admin site, but can't login via the all auth login pages

longhotsummer: let's keep it like this and change if necessary

peachjam/urls.py
```diff
@@ -286,9 +287,18 @@
...
     # Saved Documents
-    path("save-document/", SaveDocumentView.as_view(), name="save_document"),
     path(
-        "save-document/<int:pk>/delete/",
+        "save-document/<int:doc_id>/new",
+        SaveDocumentView.as_view(),
+        name="save_document",
+    ),
+    path(
+        "save-document/<int:doc_id>/update/<int:pk>",
```
longhotsummer, line 296: @actlikewill I'm not sure it makes sense to include the document ID as part of the URL. It's not needed to load the the SavedDocument id.

actlikewill: no its not needed to load in the update view, but i added it in the create view, instead of having it in the form. so I left it there to keep the urls consistent

peachjam/migrations/0150_alter_coredocument_polymorphic_ctype_and_more.py
actlikewill, line 1: @longhotsummer django is generating the migration file for me, is it related to the django upgrade or should I remove it?

peachjam/js/peachjam.ts
```diff
@@ -65,16 +66,24 @@ class PeachJam {
...
     // htmx:load is fired both when the page loads (weird) and when new content is loaded. We only care about the latter
     // case. See https://github.com/bigskysoftware/htmx/issues/1500
-    const htmxHelper = { firstLoad: true};
-    document.body.addEventListener('htmx:load', (e) => {
+    const htmxHelper = { firstLoad: true };
+    let token: string = '';
+    document.body.addEventListener('htmx:load', async (e) => {
       if (htmxHelper.firstLoad) {
         htmxHelper.firstLoad = false;
+        token = await csrfToken();
```
longhotsummer, line 74: This will load a CSRF token (and hence set the CSRF cookie) on ever page load, which defeats the purpose of doing it via javascript. Instead, we must do this at the last minute when an htmx POST request is about to happen.

peachjam/views/save_document.py
```diff
@@ -0,0 +1,153 @@
...
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        instance = SavedDocument()
+        instance.user = self.request.user
+        kwargs["instance"] = instance
+        doc_id = self.request.GET.get("doc_id")
+        if doc_id:
+            try:
+                document = get_object_or_404(CoreDocument, pk=int(doc_id))
+                kwargs.update({"initial": {"document": document}})
```
longhotsummer, line 135: Minor: you can also set `instance.document` since a ModelForm will use the `instance` to populate initial form values.

peachjam/views/save_document.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+        context["folders"] = self.get_queryset()
+        return context
+
+
+class FolderListView(BaseFolderView, ListView):
+    permission_required = "peachjam.view_folder"
+
+
+class BaseFolderFormView(BaseFolderView, FormMixin):
```
longhotsummer, line 55: Don't use `FormMixin`, because CreateView and UpdateView already include it.

peachjam/views/save_document.py
```diff
@@ -0,0 +1,153 @@
...
+
+class AllowSavedDocumentMixin:
+    def dispatch(self, *args, **kwargs):
+        if not pj_settings().allow_save_documents:
+
+            raise Http404("Saving documents is not allowed.")
+        return super().dispatch(*args, **kwargs)
+
+
+class BaseFolderView(
```
longhotsummer, line 29: Best practice is to name this `...Mixin` so that it's clear it's not a concrete view.

peachjam/views/save_document.py
```diff
@@ -0,0 +1,153 @@
...
+        instance.user = self.request.user
+        kwargs["instance"] = instance
+        return kwargs
+
+
+class FolderUpdateView(BaseFolderFormView, UpdateView):
+    permission_required = "peachjam.change_folder"
+
+
+class FolderDeleteView(BaseFolderFormView, DeleteView):
```
longhotsummer, line 80: This view's form class shouldn't be the same as Create or Update -- deletion doesn't need to validate the details of the object being deleted. So it should extend `BaseFolderView` instead.

peachjam/forms.py
```diff
@@ -301,3 +314,48 @@ def send_email(self):
...
+        model = Folder
+        fields = ["name"]
+
+    def __init__(self, *args, **kwargs):
+        self.request = kwargs.pop("request", None)
+        super().__init__(*args, **kwargs)
+
+        if self.is_bound and self.request:
+            self.data = self.data.copy()
+            if not self.data.get("name") and self.request.htmx:
```
longhotsummer, line 330: It would be a better separation of concerns if the view determined if it was an htmx request and so injected the correct data into the form. The form shouldn't be so closely bound to the request.

longhotsummer: Or, put another way, the view is responsible for deciding how the user has provided data (either as a POST or part of the htmx request), and then passing that into the form.

peachjam/views/save_document.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def get_template_names(self):
+        if self.request.htmx:
+            return ["peachjam/_folders_list.html"]
+        return ["peachjam/folders_list.html"]
+
+    def get_queryset(self):
+        return self.request.user.folders.all()
+
+    def get_context_data(self, *args, **kwargs):
```
longhotsummer, line 42: Be careful about abusing a mixin. If this logic is only used by one subclass, then it shouldn't be here.

A rule of thumb is to implement all the classes, see what's duplicated, and only if it's used more than two times, then create a mixin. Otherwise it gets complicated to work out when the various mixin methods are applicable.

peachjam/views/save_document.py
actlikewill: done

peachjam/views/save_document.py
actlikewill: removed

peachjam/views/save_document.py
actlikewill: Right, I've moved the get_context_data to the list view where it is used.

peachjam/js/peachjam.ts
actlikewill: I've updated this to use the check for the confirm event. Its working okay but it still fires on all post requests. We could check for specific target elements i.e the saveDocument button to make it only work for the saveDocuments functionality

actlikewill:
@longhotsummer there is still a bit of an issue. we need to make it such that the form is disabled before the document is saved, let me work it out

peachjam/views/save_document.py
actlikewill: done

peachjam/views/save_document.py
actlikewill: changed

peachjam/templates/peachjam/saved_document_create.html
```diff
@@ -0,0 +1,29 @@
...
+    <div class="modal-content">
+      <div class="modal-header">
+        <p class="h5 modal-title">{% trans 'Save document' %}</p>
+        <button type="button"
+                class="btn-close"
+                data-bs-dismiss="modal"
+                aria-label="Close"></button>
+      </div>
+      <div class="modal-body">
+        Saving...
```
longhotsummer, line 23: translateable

peachjam/templates/peachjam/saved_document_update.html
```diff
@@ -66,18 +61,16 @@
...
-                  hx-confirm="Are you sure you want to unsave this document?">
-            {% trans 'Unsave' %}
-          </button>
-        {% endif %}
+        <button class="btn btn-danger"
+                type="button"
+                data-bs-dismiss="modal"
+                hx-target="#saveDocumentWrapper"
+                hx-include="#save-document-form"
+                hx-headers='{"X-CSRFToken": "{{ csrf_token }}"}'
```
longhotsummer, line 69: this can fall away

peachjam/templates/peachjam/saved_document_create.html
actlikewill: one day I will remember to be translating my strings :sweat_smile:

actlikewill: done

peachjam/templates/peachjam/saved_document_update.html
actlikewill: removed

longhotsummer (approved):
looks good.

longhotsummer (approved):

PR #1893, opened 2024-07-05
Auto assign details

actlikewill:
Adds auto assign details field


peachjam/models/judgment.py
```diff
@@ -431,7 +432,7 @@ def pre_save(self):
 
         self.doc_type = "judgment"
         self.assign_mnc()
```
longhotsummer, line 434: I suggest putting an `if` before calling `assign_mnc` -- that's one of the key things that must be gated on the flag

peachjam/models/judgment.py
```diff
@@ -358,10 +358,11 @@ def generate_citation(self):
...
-        self.frbr_uri_date = str(self.date.year) if self.date else ""
-        self.frbr_uri_number = str(self.serial_number) if self.serial_number else ""
+        if self.auto_assign_details:
+            self.frbr_uri_doctype = "judgment"
+            self.frbr_uri_actor = self.court.code.lower() if self.court else None
+            self.frbr_uri_date = str(self.date.year) if self.date else ""
+            self.frbr_uri_number = str(self.serial_number) if self.serial_number else ""
         return super().generate_work_frbr_uri()
 
     def clean(self):
```
longhotsummer, line 368: the assign_mnc here would also need to be gated

actlikewill: done

peachjam/models/judgment.py
actlikewill: done

longhotsummer (approved):

PR #1898, opened 2024-07-08
Ignore for 404 errors in adapter

actlikewill:
Ignores 404 errors in indigo adapter

closes https://github.com/laws-africa/peachjam/issues/1890


longhotsummer:
This is actually an interesting case because if the publication document URL is present, that URL should NOT 404. The same for downloading an image. I'm checking on the underlying data.

longhotsummer:
These two things should match but they don't:

image

longhotsummer:
I'd like to NOT implement this change, because it caught bad data. I've deleted the offending documents in indigo, and resolved the sentry issue.

PR #1901, opened 2024-07-09
Paginated legislation listing

actlikewill:
image


liiweb/views/legislation.py
```diff
@@ -8,25 +8,43 @@
...
-class LegislationListView(TemplateView):
+class LegislationListView(FilteredDocumentListView):
     template_name = "liiweb/legislation_list.html"
     variant = "current"
     navbar_link = "legislation"
     model = Legislation
+    queryset = Legislation.objects.prefetch_related(
+        "taxonomies", "taxonomies__topic", "work"
+    )
     extra_context = {"doc_table_citations": True, "legislation_list_sort": "title"}
```
longhotsummer, line 22: these will need to be honoured in the new templates

liiweb/views/legislation.py
```diff
@@ -8,25 +8,43 @@
...
 
     def get_queryset(self):
+        qs = super().get_queryset()
         qs = (
-            self.model.objects.exclude(published=False)
-            .distinct("work_frbr_uri")
-            .order_by("work_frbr_uri", "-date", "language__pk")
+            qs.exclude(published=False)
+            .distinct(
+                "title",
```
longhotsummer, line 29: The distinct is required to ensure we only get the latest expression. I'm not sure how we also get ordering by title into that for pagination.

peachjam/forms.py
```diff
@@ -140,6 +141,9 @@ def filter_queryset(self, queryset, exclude=None):
         if outcomes and exclude != "outcomes":
             queryset = queryset.filter(outcomes__name__in=outcomes).distinct()
 
+        if q and exclude != "q":
```
longhotsummer, line 144: This doesn't work as we need it to. Currently, this will mean that the filter is applied when the facet values are calculated, which we don't want. The filtering must come entirely after the faceting is done. It's a post-aggregation filter.

peachjam/templates/peachjam/layouts/document_list.html
```diff
@@ -2,26 +2,11 @@
 {% load i18n %}
 {% block title %}{{ document.title }}{% endblock %}
 {% block page-content %}
-  {% block breadcrumbs %}{% endblock %}
+  {% block breadcrumbs %}
```
longhotsummer, line 5: the entity profile must go in the block below this, not the breadcrumbs block

peachjam/templates/peachjam/layouts/main.html
```diff
@@ -10,6 +10,9 @@
         type="text/css"/>
 {% endblock %}
 {% block head-js %}
+  <script src="https://unpkg.com/htmx.org@2.0.0"
```
longhotsummer, line 13: please install via package.json and then import into app.js

peachjam/templates/peachjam/_document_table.html
```diff
@@ -1,9 +1,10 @@
 {% load i18n peachjam %}
-<div class="table-responsive">
+<div class="table-responsive" id="doc-table">
   <table class="doc-table">
     <thead>
       <tr>
-        <th scope="col" class="cell-title">{% trans 'Title' %}</th>
+        <th></th>
+        <th scope="col" class="w-75">{% trans 'Title' %}</th>
```
longhotsummer, line 7: the new `doc-table` class handles column spacing, let me check exactly what we need to do here.

peachjam/templates/peachjam/layouts/document_list.html
```diff
@@ -48,13 +34,97 @@ <h1 class="mt-4">{% trans 'Documents' %}</h1>
           <div>
             {% block desktop-taxonomy-list %}{% endblock %}
           </div>
-          <div class="pocketlaw-hidden" data-list-facets></div>
+          <div class="pocketlaw-hidden">
+            <form method="get" id="filter-form">
```
longhotsummer, line 38: I would have thought the form element would wrap the entire table (facets, search, results)?

liiweb/views/legislation.py
```diff
@@ -8,25 +8,43 @@
...
+                "language__pk",
+            )
             .preferred_language(get_language(self.request))
         )
+
         return qs
 
+    def get_template_names(self):
+        if self.request.htmx:
+            return ["peachjam/_document_table.html"]
```
longhotsummer, line 45: the view code needs to be able to differentiate between a refresh of the facets + table, or just a refresh of the table

liiweb/templates/liiweb/legislation_list.html
```diff
@@ -1,97 +1,56 @@
...
-  {% block entity-profile %}
-    {% if entity_profile %}
-      <div class="mt-3">{% include 'peachjam/_entity_profile.html' %}</div>
-    {% endif %}
-  {% endblock %}
-  <section class="pb-5">
-    <div class="container">
-      {% block page-heading %}
-        <h1 class="my-4">{% trans 'Legislation' %}</h1>
-        {% include "peachjam/_help_button.html" with link="legislation/" %}
```
longhotsummer, line 17: we don't want to lose this

peachjam/templates/peachjam/faceted_doc_table.html
longhotsummer, line 1: This filename should start with an underscore because it's an include. Also, I think it should be similar to the existing `_document_table` which it wraps. How about `_document_table_form.html` since this contains the whole form?

longhotsummer: a  name like that will also keep it adjacent to the other `_document_table*` files in the list.

peachjam/templates/peachjam/_document_table_form.html
```diff
@@ -23,6 +23,7 @@
                   <div class="d-flex align-items-center">
                     <a href="{{ request.path }}"
                        hx-get="{{ request.path }}"
+                       hx-params="not {{ facet_name }}"
```
longhotsummer, line 26: nice!

peachjam/templates/peachjam/_document_table_form.html
```diff
@@ -0,0 +1,113 @@
...
+          </li>
+          {% for facet_name, facet in facet_data.items %}
+            {% if facet.options %}
+              <li class="list-group-item">
+                <div class="d-flex justify-content-between mb-2">
+                  <strong>{{ facet.label }}</strong>
+                  <div class="d-flex align-items-center">
+                    <a href="{{ request.path }}"
+                       hx-get="{{ request.path }}"
+                       hx-params="not {{ facet_name }}"
```
actlikewill, line 26: @longhotsummer after much fiddling, I found outh that apparently htmx does its own thing with url params, this is the correct way to handle the clearing,

peachjam/views/generic_views.py
```diff
@@ -142,7 +142,7 @@ def add_facets(self, context):
         natures = list(
             doc_n
             for doc_n in self.form.filter_queryset(
-                self.get_base_queryset(), exclude="natures"
+                self.get_queryset(), exclude="natures"
```
longhotsummer, line 145: Using get_base_queryset() is the correct thing to do; we don't want post-aggregate filters applied to aggregations.

liiweb/templates/liiweb/legislation_list.html
```diff
@@ -1,97 +1,56 @@
...
-          <li class="nav-item">
-            <a class="nav-link {% if view.variant == 'all' %}active{% endif %}"
-               href="{% url 'legislation_list_all' %}">
-              {% trans 'All legislation' %}
-            </a>
-          </li>
-        </nav>
-      {% endblock %}
-      {% if view.variant == 'repealed' %}
-        <div class="alert alert-danger">{% trans 'You are viewing repealed legislation which is no longer in force.' %}</div>
```
longhotsummer, line 68: This warning has been lost.

peachjam/templates/peachjam/_document_table_form.html
```diff
@@ -0,0 +1,132 @@
...
+                <div>
+                  <div class="letter-radiobox-container">
+                    {% for option in facet.options %}
+                      <label class="letter-radiobox" for="{{ facet_name }}_{{ forloop.counter0 }}">
+                        <input id="{{ facet_name }}_{{ forloop.counter0 }}"
+                               name="{{ facet_name }}"
+                               type="{{ facet.type }}"
+                               hx-get="{{ request.path }}"
+                               hx-push-url="true"
+                               {% if option == facet.values %}checked{% endif %}
```
longhotsummer, line 60: how does `option == facet.values` work if `option` is a single value, but `facet.values` is a list?

liiweb/templates/liiweb/legislation_list.html
actlikewill: replaced

peachjam/templates/peachjam/_document_table_form.html
actlikewill: this is the special case for the alphabet selector where in the backend we only get a single value, not a list, so this works okay

peachjam/templates/peachjam/_document_table_form.html
```diff
@@ -0,0 +1,132 @@
...
+        {% for facet_name, facet in facet_data.items %}
+          {% if facet.options %}
+            <li class="list-group-item">
+              <div class="d-flex justify-content-between mb-2">
+                <strong>{{ facet.label }}</strong>
+                <div class="d-flex align-items-center">
+                  <a href="{{ request.path }}"
+                     hx-get="{{ request.path }}"
+                     hx-params="not {{ facet_name }}"
+                     hx-push-url="true">Clear</a>
```
longhotsummer, line 47: translateable

peachjam/templates/peachjam/_document_table_form.html
```diff
@@ -0,0 +1,132 @@
...
+        <li class="position-relative list-group-item bg-light d-flex justify-content-between align-items-center">
+          <strong id="document-list-filters-offcanvas-title">{% trans 'Filters' %}</strong>
+        </li>
+        {% for facet_name, facet in facet_data.items %}
+          {% if facet.options %}
+            <li class="list-group-item">
+              <div class="d-flex justify-content-between mb-2">
+                <strong>{{ facet.label }}</strong>
+                <div class="d-flex align-items-center">
+                  <a href="{{ request.path }}"
```
longhotsummer, line 44: Can we have this only if `facet.values` isn't empty?

longhotsummer: same applies for the other "Clear" options

peachjam/templates/peachjam/_document_table_form.html
```diff
@@ -0,0 +1,132 @@
...
+    <div class="offcanvas-body"></div>
+  </div>
+  <div class="row">
+    <div class="col-lg-3 d-none d-lg-block document-list-facets-wrapper">
+      <div>
+        {% block desktop-taxonomy-list %}{% endblock %}
+      </div>
+      <ul class="list-group document-list-facets pocketlaw-hidden">
+        <li class="position-relative list-group-item bg-light d-flex justify-content-between align-items-center">
+          <strong id="document-list-filters-offcanvas-title">{% trans 'Filters' %}</strong>
```
longhotsummer, line 36: Can you add "Clear all" back, which can just be `request.path` so ALL facets and filters are cleared.

peachjam/templates/peachjam/_document_table_form.html
actlikewill: done

peachjam/templates/peachjam/_document_table_form.html
actlikewill: added

longhotsummer (dismissed):
Looks great!

longhotsummer (approved):

PR #1922, opened 2024-07-19
Locality legislation fixes

actlikewill:
image

image


actlikewill:
@longhotsummer I've put the alerts into their own block, to make the templates a little more standardized

longhotsummer (approved):

Issue #1885, opened 2024-07-04
Create login and logout views


niiroobiro:

actlikewill:
needs to be configurable across sites

PR #1932, opened 2024-07-24
Auth updates

actlikewill:
updates to auth pages layouts 
check for authenticated user before saving documents
image
image

closes https://github.com/laws-africa/peachjam/issues/1885


longhotsummer:
Don't forget to mark all strings as translateable.

PR #1936, opened 2024-07-26
Add taxonomies to court listings

actlikewill:
This adds taxonomies to the court listing pages
Also some small refactors to eliminate some repitition in the facets code.

image

closes https://github.com/laws-africa/kenyalaw-pj/issues/132

### DIFF
--- a/peachjam/views/courts.py
+++ b/peachjam/views/courts.py
@@ -38,18 +38,14 @@ class FilteredJudgmentView(FilteredDocumentListView):
 
         context["doc_type"] = "Judgment"
         context["page_title"] = self.page_title()
-        context["labels"].update({"judge": Judge.model_label_plural})
         context["doc_table_show_jurisdiction"] = False
 
         self.populate_years(context)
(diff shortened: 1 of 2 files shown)


actlikewill:
@longhotsummer since the we are faceting on the taxonomies already on the document list, I dont think we need to limit to a particular taxonomy tree

longhotsummer:
> @longhotsummer since the we are faceting on the taxonomies already on the document list, I dont think we need to limit to a particular taxonomy tree

I agree - let's go with this as it is.

peachjam/views/courts.py
```diff
@@ -105,6 +101,24 @@ def populate_facets(self, context):
                 "values": self.request.GET.getlist("attorneys"),
             }
 
+        if "taxonomy" not in self.exclude_facets:
+            taxonomies = list(
```
longhotsummer, line 105: can you sort these alphabetically at the last minute?

longhotsummer (approved):

PR #1938, opened 2024-07-26
Django 4.2

longhotsummer:
django 4.2 and python 3.10

i'm going with python 3.10 for now because that's what I've been running locally and it's only EOL in 2026 (see [here](https://devguide.python.org/versions/)), so the possible breakage is minimal.

I'm going with django 4.2 for because that's the current long-term support version, until 5.2 comes out.


goose-life (approved):

longhotsummer:
This deploys cleanly and is good to go.

After merging this, developers will need to clear their virtualenvs and re-create them from scratch.

actlikewill (approved):
works okay on my end

PR #1942, opened 2024-07-29
fix alphabet filter

actlikewill:
The typed multiple choice field returns a list.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
       run: |
         sudo apt-get -y update
         sudo apt-get install -y libreoffice poppler-utils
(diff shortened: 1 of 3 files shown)


actlikewill:
@longhotsummer I'm now thinkinig that the main reason we had that bug was because we were using the params directly without cleaning, so adding the clean method fixed the original bug

longhotsummer (approved):

longhotsummer (dismissed):

longhotsummer:
Please sanity check this works for `alphabet=%00`

actlikewill:
> Please sanity check this works for `alphabet=%00`

it works okay. the CharField cleans up the nulls

longhotsummer (approved):

PR #1947, opened 2024-07-31
adjust build workflow to limit unnecessary changes to js

longhotsummer:
the sentry webpack plugin always updates compiled js by adding a sentry debug and release id. This means that the file changes every time we merge into this branch (and then deploy), which duplicates deployments and means our users are forced to re-download JS that hasn't actually cached.

So, we now:

1. run webpack
2. strip the sentry stuff from the newly compiled code AND a copy in git at HEAD
3. diff those two versions
4. only commit the compiled javascript if the stripped versions differ


actlikewill (approved):
interesting

PR #1948, opened 2024-08-01
Accessibility improvements

actlikewill:
https://www.loom.com/share/20931ea7ccd944ff86b850482c20ce50?sid=510213df-d4ce-407d-beca-fc78fb23d294

closes https://github.com/laws-africa/kenyalaw-pj/issues/63


actlikewill:
@longhotsummer @niiroobiro I've linked a small loom for this pr above

peachjam/templates/peachjam/_header.html
```diff
@@ -1,7 +1,14 @@
 {% load static i18n %}
 <header>
+  {% block skip-links %}
+    <div id="skip-links">
+      <a class="bg-white" href="#top">Go to page content</a>
```
longhotsummer, line 5: Translateable

longhotsummer:
This looks great. The updated colour choices make sense.

The visited colour for lawlibrary isn't the right shade now that the primary colour has changed.

The visited link colour would be better to be declared as a SCSS variable in `variables.scss` near where the primary colour is declared, so that they're nearby to each other.

niiroobiro:
@actlikewill I love the colour changes. Consider adding IDs to "Recent Judgments","Recent legislation" and "Gazettes" This will give users the flexibility to navigate directly to sections that are most relevant to them. Especially when you are using a screen reader.

africanlii/static/stylesheets/africanlii.scss
```diff
@@ -31,3 +31,7 @@ iframe.video {
     display: block;
   }
 }
+
+a:visited {
```
longhotsummer, line 35: is this necessary if peachjam already has it?

peachjam/templates/peachjam/_header.html
actlikewill: done

africanlii/static/stylesheets/africanlii.scss
actlikewill: fixed, I also had to set the button color for africanlii to black like we do in other liis

actlikewill:
> @actlikewill I love the colour changes. Consider adding IDs to "Recent Judgments","Recent legislation" and "Gazettes" This will give users the flexibility to navigate directly to sections that are most relevant to them. Especially when you are using a screen reader.

@niiroobiro I was strongly considering this, but at the same time, I'm wary of adding too many links at in the skip links section. They should be few brief in order to be useful, to allow the user to quickly jump to the part of the page they are interested in. Lets wait a bit and we can come back to this when, we do some more focused accessibility stuff.

longhotsummer (approved):

Issue #1971, opened 2024-08-19
Fix sublegislation drop down on current legislation listing page


actlikewill:
Dropdown for subleg is not working for current legislation/acts in force

PR #1972, opened 2024-08-20
Fix sublegislation dropdown

actlikewill:
The children are not present on the qs when we do the variant filter on the get_base_queryset method. This moves the variant filter to the get queryset method leaving the base_queryset intact so we can fetch the children.

closes https://github.com/laws-africa/peachjam/issues/1971

### DIFF
--- a/liiweb/views/legislation.py
+++ b/liiweb/views/legislation.py
@@ -26,8 +26,8 @@ class LegislationListView(FilteredDocumentListView):
             self.form_defaults = {"sort": "-date"}
         return super().get_form()
 
-    def get_queryset(self):
-        qs = super().get_queryset()
+    def get_base_queryset(self, *args, **kwargs):
+        qs = super().get_base_queryset(*args, *kwargs)
         qs = self.get_variant_queryset(qs)
         return qs
 
@@ -87,7 +87,7 @@ class LegislationListView(FilteredDocumentListView):
         )
 
         children = defaultdict(list)
-        children_qs = self.get_base_queryset().filter(
+        children_qs = Legislation.objects.filter(
             parent_work_id__in=parents, repealed=False, metadata_json__principal=True
         )
         # group children by parent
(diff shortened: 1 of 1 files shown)


liiweb/views/legislation.py
```diff
@@ -26,8 +26,8 @@ def get_form(self):
             self.form_defaults = {"sort": "-date"}
         return super().get_form()
 
-    def get_base_queryset(self, *args, **kwargs):
-        qs = super().get_base_queryset(*args, **kwargs)
+    def get_queryset(self):
+        qs = super().get_queryset()
```
longhotsummer, line 30: I don't think this works, because facet data will be calculated on the non-variant queryset, which isn't correct.

`add_children` is a bit of a weird edge case because it actually should bypass the main queryset completely. I think it's reasonable for `add_children` to explicit load its own queryset since it has explicit requirements about it; it basically almost does it already.

actlikewill: I've changed the queryset to use the unfiltered Legislation queryset.

liiweb/views/legislation.py
```diff
@@ -27,7 +27,7 @@ def get_form(self):
         return super().get_form()
 
     def get_base_queryset(self, *args, **kwargs):
-        qs = super().get_base_queryset(*args, **kwargs)
+        qs = super().get_base_queryset(*args, *kwargs)
```
longhotsummer, line 30:
```suggestion
        qs = super().get_base_queryset(*args, **kwargs)
```

liiweb/views/legislation.py
```diff
@@ -87,7 +87,7 @@ def add_children(self, queryset):
         )
 
         children = defaultdict(list)
-        children_qs = self.get_base_queryset().filter(
+        children_qs = Legislation.objects.filter(
```
longhotsummer, line 90: Also add in

```python
children_qs.preferred_language(get_language(self.request))
```

so that the best language is used where possible.

actlikewill: right, added

longhotsummer (approved):

Issue #1970, opened 2024-08-19
Deploy survey for MalawiLII and ZambiaLII


niiroobiro:
Can we add the following surveys? They need to be placed somewhere on the site where they are visible to users so that they can participate.
Zambia: https://forms.gle/dETZD858WR4hMuAR8
Malawi: https://forms.gle/3KQZrVQZNvt7xJj78

PR #1976, opened 2024-08-27
Survey banner

actlikewill:
image

image
closes https://github.com/laws-africa/peachjam/issues/1970


actlikewill:
@niiroobiro here are the banners for the surveys. Please feel free to suggest some changes, especially to the wording used

longhotsummer (approved):
let's see what the uptake is

longhotsummer:
if the homepage isn't enough, we may need a banner on all pages

PR #1980, opened 2024-08-28
Fix sublegislation ordering

actlikewill:
This fixes sublegislation ordering, on both the drop down and the listing page

image
closes https://github.com/laws-africa/kenyalaw-pj/issues/201

### DIFF
--- a/liiweb/views/legislation.py
+++ b/liiweb/views/legislation.py
@@ -22,7 +22,7 @@ class LegislationListView(FilteredDocumentListView):
 
     def get_form(self):
         self.form_defaults = {"sort": "title"}
-        if self.variant == "recent":
+        if self.variant in ["recent", "subleg"]:
             self.form_defaults = {"sort": "-date"}
         return super().get_form()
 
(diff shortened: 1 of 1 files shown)


liiweb/views/legislation.py
```diff
@@ -89,7 +89,7 @@ def add_children(self, queryset):
         children = defaultdict(list)
         children_qs = Legislation.objects.filter(
             parent_work_id__in=parents, repealed=False, metadata_json__principal=True
-        )
+        ).order_by("-date")
```
longhotsummer, line 92: This is interesting... this only impacts the legislation views which are by default sorted by title. So maybe we leave this out here?

Basically, we don't currently sort the children based on what the main listing is sorting by.

actlikewill: @longhotsummer I think the request was that the children be sorted according to latest first,

longhotsummer: ok, let's do this globally then. All good.

Issue #815, opened 2023-01-16
The filters document type and document nature return identical filter categories


NotoriousMBB:
See for example here: https://agp.africanlii.org/search/?q=refugee

Is this a result of incorrect data capturing/model or what?

longhotsummer:
this is fixed

PR #1991, opened 2024-08-29
Document nature is now required

longhotsummer:
This ensures that all documents have a descriptive nature, even those without a subtype. For example, all Legislation objects now default to "Act". This means we can rely on the nature as a database-indexed description of the type of every document.

This makes it possible to facet on the nature. Before, we had to facet on a weird mixture of doc_type and nature.

In future, nature will be preferred over doc_type in search results. But we need to backfill the nature field in ES for that to work.

Other:

* makes faceting code support `(value, label)` tuples so that labels can be different to values and can take translation into account
* pulls countries out of `fixtures/documents/sample_documents.json` so it's easier to manage

Example by-law listing:

image

Example general legislation listing:

image

## Outstanding

* [x] if possible, a mechanism to backfill elasticsearch documents to ensure they all have a nature, so we can stop using `doc_type` with search
* [x] filter facets must be able to have different values and labels, so that the label can be "Act" but the value is "act"


longhotsummer:
Closes https://github.com/laws-africa/peachjam/issues/815

longhotsummer:
@actlikewill this is ready for review

peachjam/views/courts.py
```diff
@@ -65,25 +73,23 @@ def add_facets(self, context):
...
-                .values_list("outcomes__name", flat=True)
+                .values_list("outcomes__id", flat=True)
                 .distinct()
-                if outcome
             )
             context["facet_data"]["outcomes"] = {
                 "label": _("Outcomes"),
                 "type": "checkbox",
-                "options": outcomes,
+                "options": [(outcome.code, outcome.name) for outcome in outcomes],
```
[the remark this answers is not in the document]
longhotsummer: i've reverted this to outcome.name. We should add the code for outcome and other facets, like matter type, so that we can facet across multiple languages. I'm not going to include that change in this PR.

peachjam/migrations/0152_backfill_natures.py
```diff
@@ -0,0 +1,40 @@
...
+        ("judgment", "judgment", "Judgment"),
+        ("legal_instrument", "legal-instrument", "Legal Instrument"),
+        ("legislation", "act", "Act"),
+        ("book", "book", "Book"),
+        ("journal", "journal", "Journal"),
+    ]
+    for doc_type, nature, name in DOC_TYPES:
+        nature = DocumentNature.objects.get_or_create(
+            code=nature, defaults={"name": name}
+        )[0]
```
actlikewill, line 23: image
@longhotsummer I'm not sure what happened but when I ran the migration, the names were not filled in, I didn't check it before I ran the migration, so I may have had bad data already I'm not sure

actlikewill: maybe get_or_create doesn't work in migrations

longhotsummer: i've changed the migration to set the name explicitly

peachjam/migrations/0152_backfill_natures.py
```diff
@@ -0,0 +1,42 @@
...
+        ("legislation", "act", "Act"),
+        ("book", "book", "Book"),
+        ("journal", "journal", "Journal"),
+    ]
+    for doc_type, nature, name in DOC_TYPES:
+        nature = DocumentNature.objects.get_or_create(code=nature)[0]
+        # get_or_create with defaults={name: name} doesn't seem to work
+        if not nature.name:
+            nature.name = name
+            nature.save()
```
actlikewill, line 25: @longhotsummer this is still not working for me :thinking:

longhotsummer: I wonder if it's related to the translated names functionality on those fields? Try setting `nature.name_en = name` as well?

peachjam/migrations/0152_backfill_natures.py
```diff
@@ -22,6 +22,7 @@ def forwards(apps, schema_editor):
         # get_or_create with defaults={name: name} doesn't seem to work
         if not nature.name:
             nature.name = name
+            nature.name_en = name
```
actlikewill, line 25: @longhotsummer I've added it

actlikewill (approved):

peachjam/migrations/0152_backfill_natures.py
longhotsummer: I've done a test deployment and this works.

PR #1992, opened 2024-08-30
pin whitenoise 5.3.0

longhotsummer:
This may help solve our intermittent database connection issues

See
https://forum.djangoproject.com/t/interfaceerror-and-transactionmanagementerror-postgresql/16924/11 and https://github.com/evansd/whitenoise/issues/556


longhotsummer:
I've tested this on civlii and static assets are served as normal. We'll need it deployed to the bigger sites to see if it has an impact. We currently see about 3-6 issues every hour across all our sites, so it's not a common issue.

actlikewill (approved):

PR #1994, opened 2024-08-30
Allow iframes in ckeditor

actlikewill:
The problem occurred from two places, the check for removing whitespaces got rid of the iframes, and the ckeditor config also was hiding the iframe content. Now you can paste the iframe into the source and it should work. This also enables the Iframe plugin so you can enter the url and other details manually. 

image

closes https://github.com/laws-africa/peachjam/issues/1944


longhotsummer (approved):
nice

Issue #1953, opened 2024-08-05
Include regional courts of justice block on the AU page


longhotsummer:
We have this on the AfricanLII homepage, but not on the AU page.

PR #1995, opened 2024-08-30
Add regional courts to AU page

actlikewill:
image

closes https://github.com/laws-africa/peachjam/issues/1953


actlikewill:
@longhotsummer do we need to layout this in a special way eg, to have multiple columns for multiple court classes? The Regional Courts are saved as court classes

longhotsummer:
I don't think we need to do anything fancy -- as long as the details are roughly the same as the homepage.

longhotsummer (approved):

PR #2001, opened 2024-09-04
Gazette tweaks

actlikewill:
- breadcrumbs were not visible on the listing page because of the place check
- title case special issue


actlikewill:
@longhotsummer we could also consider changing links in the liis to use the jurisdiction based listing view, so that they can see the title of their jurisdiction in their gazette page

longhotsummer (approved):

PR #2004, opened 2024-09-05
Gazette special facet

actlikewill:
- adds "special" filter facet for gazettes
- allows facets options to have separate name and value
image


peachjam/views/gazette.py
```diff
@@ -122,6 +122,10 @@ class GazetteYearView(YearMixin, FilteredDocumentListView):
     navbar_link = "gazettes"
     locality = None
     group_by_date = "month-year"
+    special_facet_options = [
+        {"name": _("Special Issue"), "value": "special"},
```
longhotsummer, line 126: the `_` must be the lazy variant, so that the translation is done at the last minute.

actlikewill: is it okay to change it from the import thereby changing it everywhere in the file?

actlikewill: I've seen we do it elsewhere so i guess its okay

longhotsummer: > I've seen we do it elsewhere so i guess its okay

yes it shouldn't hurt

longhotsummer (approved):

actlikewill:
I've added supplement to the facets 
image

longhotsummer (approved):

Issue #1979, opened 2024-08-28
Uncommenced national legislation view is erroneously including local legislation


longhotsummer:
https://lawlibrary.org.za/legislation/uncommenced

![image](https://github.com/user-attachments/assets/ff0520bc-7979-45c1-9316-5562e509f181)

The by-law indicated has a locality and should not be included in the national view.

PR #2008, opened 2024-09-09
Fix locality filters for Lawlibrary

actlikewill:
This fixes the base view that is used for the national legislation listing views.
The base view should filter out documents that have localites

closes https://github.com/laws-africa/peachjam/issues/1979

### DIFF
--- a/lawlibrary/urls.py
+++ b/lawlibrary/urls.py
@@ -20,6 +20,21 @@ urlpatterns = [
         name="legislation_list_all",
     ),
     path(
+        "legislation/uncommenced",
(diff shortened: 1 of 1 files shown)


longhotsummer (approved):
sneaky bug

actlikewill:
maybe we should consider supporting the locality vs national setup from peachjam itself,

longhotsummer:
> maybe we should consider supporting the locality vs national setup from peachjam itself,

Yes, ultimately we should. We can do that when the second LII with local legislation comes online.

PR #2009, opened 2024-09-10
PublicationFile

goose-life:
requires https://github.com/laws-africa/indigo/pull/2231

https://www.loom.com/share/b5b4fd2b06ea444d86863db9ea36cfa4
https://www.loom.com/share/97d83c0f00d64b6091206d00196abe2e
https://www.loom.com/share/e7bc1f2b59c548569e9e4cdb74da190c


peachjam/adapters/indigo.py
```diff
@@ -510,6 +519,71 @@ def download_source_file(self, url, doc, title):
                 },
             )
 
+    def create_publication_file(self, publication_document, doc, title, stub=False):
+        from peachjam.models import PublicationFile
+
+        logger.info(f"Creating / updating a publication file for {title}")
+
+        # first delete any existing PublicationFile file: a new one will be saved if needed
+        # TODO: only delete it if something has changed? (how to know that a file has changed?)
```
longhotsummer, line 528: always delete, it's much safer

peachjam/models/core_document_model.py
```diff
@@ -896,6 +896,87 @@ def save(self, *args, **kwargs):
             # first save, set the download filename
             self.set_download_filename()
 
+    def delete(self, *args, **kwargs):
+        if (
+            hasattr(self.document, "publication_file")
+            and self.document.publication_file.use_source_file
+        ):
+            raise ValidationError(
```
longhotsummer, line 904: I don't think we should do this, because it could potentially break general deletion of a document, such as during ingestion.

peachjam/adapters/indigo.py
```diff
@@ -510,6 +519,71 @@ def download_source_file(self, url, doc, title):
...
+                    f"  Deleting existing PublicationFile file on {doc.work_frbr_uri}"
+                )
+                doc.publication_file.file.delete()
+
+        if stub:
+            if hasattr(doc, "source_file"):
+                logger.info("  Stub: Using the source file")
+                PublicationFile.objects.update_or_create(
+                    document=doc,
+                    defaults={
```
longhotsummer, line 541: I think you also want to set `url` to None, otherwise it could have an old value (even though it is not used).

longhotsummer: oh, `clean()` will sort it out.

peachjam/models/core_document_model.py
```diff
@@ -896,6 +896,87 @@ def save(self, *args, **kwargs):
...
+        # final check: we need at least one of these
+        if not (self.use_source_file or self.url or self.file):
+            raise ValidationError(
+                _(
+                    "You need to provide a URL or a file, or if you checked 'Use source file' there must be one to use."
+                )
+            )
+
+    def save(self, *args, **kwargs):
+        self.clean()
```
longhotsummer, line 975: I suspect this may cause an issue later, but let's go with it. Usually `clean()` is only called by a view or a form that is validating inputs, so that they can be corrected. Calling it on every save() is not always what you want. But I can't think of an example right now where this could go wrong, so let's go with it.

peachjam/views/documents.py
```diff
@@ -126,10 +126,35 @@ def render_to_response(self, context, **response_kwargs):
...
                     )
         raise Http404()
 
 
+class DocumentPublicationView(DocumentSourceView):
+    def render_to_response(self, context, **response_kwargs):
+        if hasattr(self.object, "publication_file"):
+            publication_file = self.object.publication_file
+            if publication_file.use_source_file:
+                # TODO: change document_source to display inline?
```
longhotsummer, line 139: See https://github.com/laws-africa/peachjam/commit/3df4b19079e4e10d7b1c1c98570cf2c7df30487c

using `inline` means the browser disregards the filenames. I think we want attachment in all cases.

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -191,6 +191,17 @@ <h1>{{ document.title }}</h1>
                           </dd>
                         {% endif %}
                       {% endblock %}
+                      {% block document-metadata-content-publication %}
```
longhotsummer, line 194: I think this should be near, and be displayed very similarly to, the attached files. So it would be something like:

```
Original publication: Download PDF (32KB)
```

longhotsummer: See the `document-metadata-media-summary-download` block in document_detail.html

peachjam/models/core_document_model.py
goose-life: I'd also be happy to move all the logic back into `save()` where I had it originally — I moved the bulk into `clean()` to be able to test and demonstrate the logic in the admin form

goose-life:
updated detail view

image

peachjam/views/documents.py
longhotsummer: for PDF files, it works like this:

* if `content-disposition` is `inline`, then the browser will download the PDF and try to show it in the browser. If the user then clicks the "save" button in the PDF viewer, the browser chooses a filename based on the URL and ignores the `filename` part of `content-disposition` -- it sometimes seems to re-download the file
* if the `content-disposition` is `attachment`, then the browser will download the file and save it using the `filename` part of the header. The user must then open the file in their downloads to view the actual file

So for source files, because we want nice filenames, we changed to `attachment` from `inline` so that the filename was actually used.

goose-life:
image

goose-life:
I finally managed to test ingestion so I've moved all the logic there and off of the model.

peachjam/models/core_document_model.py
goose-life: this immediately caused an issue 😅 when we delete the file before doing the update (or create)

peachjam/adapters/indigo.py
```diff
@@ -551,33 +552,45 @@ def create_publication_file(self, publication_document, doc, title, stub=False):
...
-                        document=doc,
-                        defaults={
-                            "filename": filename,
-                            "mimetype": mimetype,
-                            "size": f.size,
-                            "url": url,
-                            "use_source_file": False,
-                        },
+            filename = publication_document.get(
+                "filename", f"{slugify('Publication: ' + title)}.pdf"
```
longhotsummer, line 556: the `filename` entry will always be there, but may be None. So you want something like this instead:

```python
filename = publication_document.get("filename") or slugify(...)
```

peachjam/adapters/indigo.py
```diff
@@ -551,33 +552,45 @@ def create_publication_file(self, publication_document, doc, title, stub=False):
...
+            )
+            if publication_document.get("has_trusted_url"):
+                logger.info(f"  Using publication file from trusted URL {url}")
+                mimetype = publication_document.get("mime_type", "application/pdf")
```
longhotsummer, line 560: similarly here; this only works if `mime_type` isn't in the json, but the api will always include it

peachjam/adapters/indigo.py
goose-life: thanks! how embarrassing, I thought the second argument would be returned if get() returned None 🙈

peachjam/adapters/indigo.py
goose-life: all fixed now, I think!

longhotsummer (approved):

PR #2019, opened 2024-09-11
Move metadata field to coredocument model

actlikewill:
This moves the metadata_json field into the core document model


actlikewill:
@longhotsummer This keeps the old metadata json field because this is a potentially destructive change. We can remove it right after if everything works okay.

longhotsummer (approved):
Looks good.

longhotsummer (approved):

Issue #2018, opened 2024-09-11
Changes to TanzLII


niiroobiro:
1. Add Finding Aids under Research Hub: https://tanzlii.org/taxonomy/finding-aids
2. On a judgment display, move the order field to the end of the display:https://tanzlii.org/akn/tz/judgment/tzca/2024/155/eng@2024-03-05
![Image](https://github.com/user-attachments/assets/a5fee3ec-e03b-43b3-9dad-1d897c60b8b2)

actlikewill:
@longhotsummer There was a request to have better visual separation of the various blocks in this section. The request was to separate them with different backgrounds rather than using <hr/>. I'm curious on your thoughts on this

actlikewill:
@niiroobiro for the first Finding Aids issue, that is a change to the taxonomy content and can be handled in the admin.

PR #2029, opened 2024-09-12
Changes order on judgment metadata

actlikewill:
image

image

closes https://github.com/laws-africa/peachjam/issues/2018


actlikewill:
> I think this is an improvement. Does it look okay on mobile, or should the headings collapse above the content on small screens?

It didn't look right on small screens, I've reworked the css a bit to push the title above on small screens

PR #2030, opened 2024-09-13
Repeat schedule for ingestors

actlikewill:


peachjam/models/ingestors.py
```diff
@@ -26,13 +27,29 @@ def __str__(self):
...
+    )
+
     adapter = models.CharField(_("adapter"), max_length=2048)
     last_refreshed_at = models.DateTimeField(
         _("last refreshed at"), null=True, blank=True
     )
     name = models.CharField(_("name"), max_length=255)
     enabled = models.BooleanField(_("enabled"), default=True)
 
+    repeat = models.BigIntegerField(choices=INGESTOR_REPEAT_CHOICES, default=Task.NEVER)
```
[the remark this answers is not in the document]
actlikewill: we don't strictly need both, but I think it may be useful to control both independently

peachjam/tasks.py
```diff
@@ -113,7 +113,9 @@ def run_ingestors():
 
     for ingestor in Ingestor.objects.all():
         if ingestor.enabled:
-            run_ingestor(ingestor.pk)
+            run_ingestor(
```
longhotsummer, line 116: this should be also be done when the ingestor object is saved (eg. via admin) -- I think it's reasonable to have a method on the ingestor that is called in the `save()` method, and can be called here too.

longhotsummer: Because we want the scheduled task to update when the ingestor is updated in the admin area.

actlikewill: good idea, added

longhotsummer: to be picky: this method AND the the `.save()` method both run the ingestor in the same way -- logic is duplicated. Additionally, this method has to know how to use the internal details of the ingestor to know how to run it.

It's more object-oriented and better encapsulation to put the invocation logic into the Ingestor object. So `Ingestor.save()` calls `Ingestor.queue_task()`, and so does this. The Ingestor itself is responsible for knowing how to queue up its task given its state.

longhotsummer: Actually, it's done in THREE places now -- in the admin, too. There should be one method on the Ingestor that handles queuing up the task.

peachjam/models/ingestors.py
```diff
@@ -26,13 +27,29 @@ def __str__(self):
...
+    schedule = models.BigIntegerField(
```
longhotsummer, line 49: Now that we know that these two must be different, it's good to have them separate. So we can configure `schedule` to be a small value like 10 minutes, and then repeat can be daily. That will ensure it's run once and then repeats. It may run more often than we need, but that's better than not running.

actlikewill: yeah, I've updated this with the new queue_task method

peachjam/models/ingestors.py
```diff
@@ -74,3 +91,11 @@ def get_adapter(self):
...
         settings = {s.name: s.value for s in ingestor_settings}
         return klass(self, settings)
+
+    def queue_task(self):
+        run_ingestor(self.id, repeat=self.repeat, schedule=self.schedule)
+
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
+        if self.enabled:
+            self.queue_task()
```
actlikewill, line 101: @longhotsummer we have an infinite loop here. we call queue_task on save, but when we run the task, we update the last_refreshed, which calls save, entering a loop. We can schedule the task to run five minutes out, but the loop is there

actlikewill: that works. done

longhotsummer (approved):

longhotsummer:
Once this is deployed, the defaults will be `never`. Will you go through each ingestor on each lii to update it? Or do you want to set a default of daily?

peachjam/models/ingestors.py
```diff
@@ -26,13 +27,29 @@ def __str__(self):
...
     adapter = models.CharField(_("adapter"), max_length=2048)
     last_refreshed_at = models.DateTimeField(
         _("last refreshed at"), null=True, blank=True
     )
     name = models.CharField(_("name"), max_length=255)
     enabled = models.BooleanField(_("enabled"), default=True)
 
+    repeat = models.BigIntegerField(choices=INGESTOR_REPEAT_CHOICES, default=Task.DAILY)
+    schedule = models.BigIntegerField(
+        choices=INGESTOR_REPEAT_CHOICES, default=Task.HOURLY
```
actlikewill, line 50: @longhotsummer defaults set to one hour for schedule and daily for repeat

longhotsummer (approved):

PR #2031, opened 2024-09-13
use outcomes__name not outcomes__code

longhotsummer:
later we must give outcomes a slug/code so that we support translations

fixes https://lawsafrica.sentry.io/issues/5804276287/

image


longhotsummer:
fixes https://github.com/laws-africa/peachjam/issues/2026

actlikewill (approved):

PR #2033, opened 2024-09-13
search result sorts document type by count (desc)

### LINKED ISSUE #1965: When viewing filters, I want them ordered by number of hits rather than alphabetically

longhotsummer:
This makes it clearer what the best matching document types are

The other fields sort by alpha because they are often long and alpha makes it easier to browser through the multiple values

closes #1965

image


actlikewill (approved):
looks good,

PR #2034, opened 2024-09-14
add main content type buttons for mobile on lawlibrary

longhotsummer:
image

this makes it easier to jump to the primary content types on mobile, when the menu is collapsed


actlikewill (approved):
looks great, we should consider this for the other liis, depending on feedback

PR #2040, opened 2024-09-16
validate year for year mixins

longhotsummer:
fixes https://github.com/laws-africa/peachjam/issues/1990


actlikewill (approved):
Looks good, we could also have created a year url converter similar to the iso date converter we use elsewhere.

PR #2041, opened 2024-09-16
tolerate multiple attachments with the same name

longhotsummer:
see https://github.com/laws-africa/peachjam/issues/2024

it doesn't fix the root cause, but it means that views will succeed for now until we fix the underlying data

### DIFF
--- a/peachjam/views/documents.py
+++ b/peachjam/views/documents.py
@@ -1,6 +1,6 @@
 from cobalt import FrbrUri
 from django.http import Http404, HttpResponse
-from django.shortcuts import get_object_or_404, redirect, reverse
+from django.shortcuts import get_list_or_404, redirect, reverse
 from django.utils.decorators import method_decorator
 from django.utils.translation import get_language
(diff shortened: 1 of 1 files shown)


peachjam/views/documents.py
```diff
@@ -142,7 +142,9 @@ class DocumentMediaView(DetailView):
     slug_url_kwarg = "frbr_uri"
 
     def render_to_response(self, context, **response_kwargs):
-        img = get_object_or_404(self.object.images, filename=self.kwargs["filename"])
+        # there should only be one, but until we enforce uniqueness of filenames, get the first in the list
+        # see https://github.com/laws-africa/peachjam/issues/2024
+        img = get_list_or_404(self.object.images, filename=self.kwargs["filename"])[0]
```
actlikewill, line 147: could this cause an index error if list is empty?

longhotsummer: `get_list_or_404` raises a 404 if the list is empty

actlikewill (approved):

PR #2043, opened 2024-09-16
highligh best search result if it's score is better than the 2nd result

### LINKED ISSUE #1964: As a user, I want the highest ranked document in search results highlighted
The idea here is to call out the top search result when its search score is notably higher than the other results, as this indicates this is very likely the result the user is looking for.

For example, in this search, the top result is scored significantly higher than the next one:

image

It's useful to call that out to the user, because they don't see internal scores.

longhotsummer:
Closes #1964 

After discussion with Thirusha, we thought a consistent gold colour was better than the primary colour.

image


actlikewill (approved):
love it

PR #2048, opened 2024-09-20
Add convert html to judgment

actlikewill:
This adds the convert html to pdf method to the judgment model

### DIFF
--- a/peachjam/models/judgment.py
+++ b/peachjam/models/judgment.py
@@ -1,8 +1,5 @@
-from tempfile import NamedTemporaryFile
-
 from countries_plus.models import Country
 from django.contrib.contenttypes.fields import GenericRelation
-from django.core.files import File
 from django.db import models
 from django.db.models import Max
 from django.template.defaultfilters import date as format_date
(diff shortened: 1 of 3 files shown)


peachjam/models/judgment.py
```diff
@@ -441,6 +445,22 @@ def pre_save(self):
             self.assign_title()
         super().pre_save()
 
+    def convert_html_to_pdf(self):
+        with NamedTemporaryFile(suffix=".html") as html_file:
+            html_file.write(self.content_html.encode("utf-8"))
```
longhotsummer, line 450: I suggest you do something like `html_file.write(self.prepare_content_html_for_pdf())` so that subclasses can override the HTML that is used to generate the PDF, such as by adding a logo at the top.

Also, I think this can safely all be done on the CoreDocument object.

actlikewill: ok, done

peachjam/tasks.py
```diff
@@ -206,6 +206,19 @@ def convert_source_file_to_pdf(source_file_id):
     log.info("Conversion to PDF done")
 
 
+@background(queue="peachjam", remove_existing_tasks=True)
+def convert_html_to_pdf(judgment_id):
+    from peachjam.models import Judgment
```
longhotsummer, line 211: I think this should work for any file, not just judgment.

actlikewill: done

longhotsummer (approved):

PR #2056, opened 2024-09-23
correctly schedule regular tasks

longhotsummer:
Before this change, the daily and weekly tasks were scheduled to run "in 1 day" and "in one week", and then daily/weekly after that.

The problem is that every time we deployed an update, it re-scheduled the tasks. So the weekly task in particular just never got run, because it was continually pushed out.

Instead, we now schedule as "run on this date" and then repeat daily/weekly.


actlikewill (approved):
looks good,

PR #2062, opened 2024-09-25
clamp pagination at 10 pages

longhotsummer:
This prevents a major performance issue with very large datasets (> 50,000) where crawlers can page through the entire dataset, hurting PostgreSQL because `OFFSET` pagination in psql is inefficient.

See https://readyset.io/blog/optimizing-sql-pagination-in-postgres

https://www.loom.com/share/973f54014ae94bb4b18bb2bfd5b6266e

### DIFF
--- a/peachjam/templatetags/peachjam.py
+++ b/peachjam/templatetags/peachjam.py
@@ -2,7 +2,6 @@ import datetime
 import json
 
 from django import template
-from django.core.paginator import Paginator
 from django.http import QueryDict
 from django.urls import reverse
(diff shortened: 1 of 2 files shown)


actlikewill (approved):
great, i'm very curious if users will notice

longhotsummer:
> great, i'm very curious if users will notice

Me too. If necessary, we can tweak the page size, or move to cursor-based pagination.

longhotsummer:
An alternative later is https://github.com/photocrowd/django-cursor-pagination

PR #2067, opened 2024-09-27
ratification counts; move ratification templates into africanlii

### LINKED ISSUE #1778: Ratifications count incorrect on AfricanLII
When importing ratification information, we report on all 55 states. Those that have not signed and/or ratified a document, are still imported but the signature date is left blank. Yet the counter on the Ratifications tab counts each entry.

Could the counter please count only the actual Ratification Date that is not null entries?

longhotsummer:
fixes #1778 

image


actlikewill (approved):
looks good

PR #2070, opened 2024-09-27
get_update_or_create method

actlikewill:


longhotsummer:
does it need to be used by other fields that would cause the document to be re-indexed for search?

actlikewill:
> does it need to be used by other fields that would cause the document to be re-indexed for search?

there is actually no other place it is needed in the adapter, the only other that do update_or_create are source files and publication files

Issue #1808, opened 2024-04-30
Handle long document lists and citation lists


actlikewill:
We need a better way of handling long lists of documents.

This is a problem for both long lists of citations in the document detail view and the legislation view.

We may want to split incoming and outgoing citations into separate tabs, and make them actual URLs so they are loaded separately (or via HTMX)

longhotsummer:
eg. see the expanded citation list for https://tanzlii.org/akn/tz/act/ord/1930/11/eng@2019-11-30

PR #2071, opened 2024-09-27
Citations speed

longhotsummer:
https://www.loom.com/share/5584208b940e476b9662234b5af9a8b7

image

This changes how we render citation lists, so that documents that are heavily cited don't have very long page load times.

* only show the first 10 of each group
* use htmx to load the next ten, etc.

Also changes to link to the cited works using the full expression FRBR URI, not the work uri.

Also reduces py-4 to py-3 since our base font size is a  bit bigger.

### DIFF
--- a/peachjam/models/citations.py
+++ b/peachjam/models/citations.py
@@ -3,6 +3,8 @@ from datetime import timedelta
 from random import randint
 
 from django.db import models
+from django.db.models import F, Window
+from django.db.models.functions import RowNumber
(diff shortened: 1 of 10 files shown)


longhotsummer:
Closes #1808

actlikewill (approved):
awesome, htmx is perfect for this experience

longhotsummer:
> awesome, htmx is perfect for this experience

absolutely -- i :heart: htmx

PR #2080, opened 2024-09-30
Check for duplicate images in adapter

actlikewill:
check for duplicate image


longhotsummer:
the `filenames` set already does this?

longhotsummer:
https://github.com/laws-africa/peachjam/pull/2065 already fixed this in theory?

longhotsummer (approved):
happy to include this to make it more robust

Issue #2076, opened 2024-09-27
As a user on AfricanLII, I want to see judgment date under date on the listing page not datestamp


niiroobiro:
![Image](https://github.com/user-attachments/assets/bf2e17d1-97f1-4bfd-a9de-07c4292d7298)
Also date format should be DD Month YYYY
https://africanlii.org/indexes/case-indexes/case-indexes-human-rights?jurisdictions=African%20Regional%20Bodies&q=&sort=-date

PR #2083, opened 2024-10-01
fix date value for index listing page

actlikewill:
fixes https://github.com/laws-africa/peachjam/issues/2076


longhotsummer (approved):
let's hope it's never null (I don't think it's possible to be null)

PR #2087, opened 2024-10-03
Order subsidiary legislation by frbr_uri number 

actlikewill:
- This adds the frbr uri number as a secondary sort parameter for subleg
- The ordering needs to be reversed if the order date sorting is reversed

image

closes https://github.com/laws-africa/kenyalaw-pj/issues/115

### DIFF
--- a/liiweb/views/legislation.py
+++ b/liiweb/views/legislation.py
@@ -20,7 +20,7 @@ class LegislationListView(BaseLegislationListView):
     def get_form(self):
         self.form_defaults = {"sort": "title"}
         if self.variant in ["recent", "subleg"]:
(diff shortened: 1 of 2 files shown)


actlikewill:
@longhotsummer this is still less than perfect, because we sort primarily by date. First some amendments from earlier years can show up in the current year. Second, some of them are not releaseed chronologically eg. in the screenshot, legal notices 63 and 64 have a date earlier than 55, and the rest following which puts them out of order

longhotsummer:
I think this is a reasonable start, particularly for recent legislation.

longhotsummer (approved):

Issue #528, opened 2022-10-07
As an admin I want to be able to force an ingestor to update a specific document or set of documents


longhotsummer:

longhotsummer:
We can't do this yet, because a document doesn't know which ingestor to use.

PR #2098, opened 2024-10-10
Edit with ingestor

longhotsummer:
* link documents to an ingestor
* default the edit button to a link provided by the ingestor, rather than in the admin area
* provide admin area edit is a dropdown option
* make mobile buttons smaller so they don't wrap

This also sets us up to use the link between a document and an ingestor to be able to do on-demand background updates for a selection of documents #528   

image

Mobile:
image


actlikewill (approved):
we will need to add the function for tausi judgments as well

Issue #2104, opened 2024-10-15
Support ratification on all LIIs


longhotsummer:
* [x] move ratification models into africanlii
* [x] rename db table in production for africanlii
* [x] ratification api
* [x] adapter to fetch per-country info from the liis
* [x] taxonomy entry

longhotsummer:
# Proposal

The only aspect that makes treaties different to other documents is their ratification information (date, country, place, etc.). AfricanLII already has the concept of Ratification data, which is linked to a Work.

1. treaties are marked up in Indigo or uploaded manually as a document
2. the Ratification class is moved into peachjam (from AfricanLII) and can be linked to any Work
3. Ratification information is populated for each treaty work
4. treaties are tagged with a particular taxonomy tag
5. a new listing page is built that uses that tag to list treaties
6. the detail page for a document includes ratification information

In terms of integrating with AfricanLII, it will work as follows:

1. all the details can be captured on AfricanLII if necessary
2. ratification information can also be captured on a per-country basis within a LII
3. a new ingestor on AfricanLII will fetch ratification details from each LII, for a particular country, and add that to its local database (it will overwrite existing ratification details for that country, if present)

This way, AfricanLII has a global view of ratification information from the LIIs, and each LII can manage its own ratification data. Ratifications are linked by Work FRBR URI.

PR #2105, opened 2024-10-15
Ratifications

longhotsummer:
* ratification model moved into peachjam to be used by all liis
* ratification API for use by other LIIs needing to fetch this information (AfricanLII)
* ratification adapter for use by africanlii
* ratifications are shown in document detail view, if available

image

Ref https://github.com/laws-africa/peachjam/issues/2104

This now means that we can configure any LII to fetch ratification information from another LII. So AfricanLII can be the source of ratification details for some LIIs, and it can also pull ratification details from other LIIs.


peachjam/adapters/base.py
```diff
@@ -41,3 +48,16 @@ def get_edit_url(self, document):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.client = requests.session()
+        self.api_url = self.settings["api_url"]
+
+    def client_get(self, url, **kwargs):
+        logger.debug(f"GET {url} kwargs={kwargs}")
+        r = self.client.get(url, **kwargs)
+        r.raise_for_status()
+        return r
```
actlikewill, line 63: I wanted to do this when I worked on the deletion adapter, because we use a post method on that one, I wanted a generic client. but this is fine since we don't have a need for then post method generally

actlikewill:
@longhotsummer looks good, the migrations need to be updated though

longhotsummer:
Migrations updated

PR #2107, opened 2024-10-16
prompt user to try using "quotes" in search

longhotsummer:
For a simple search, if the search string doesn't have quotes, suggest the user wraps the search in quotes to find more exact matches. The user can click the click to do so.

Closes https://github.com/laws-africa/peachjam/issues/1963

image


actlikewill (approved):
neat,

PR #2127, opened 2024-10-29
Show only courts that have causelists

actlikewill:
closes https://github.com/laws-africa/kenyalaw-pj/issues/260

### DIFF
--- a/peachjam/views/causelist.py
+++ b/peachjam/views/causelist.py
@@ -1,6 +1,7 @@
 from functools import cached_property
 from math import ceil
 
+from django.db.models import Prefetch
 from django.shortcuts import get_object_or_404
 from django.urls import reverse
 from django.utils.text import gettext_lazy as _
@@ -40,8 +41,18 @@ class CauseListListView(TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
+        # get only court classes and courts that have causelists
+        context["court_classes"] = (
+            CourtClass.objects.filter(courts__causelists__isnull=False)
+            .prefetch_related(
+                Prefetch(
+                    "courts",
+                    queryset=Court.objects.filter(causelists__isnull=False).distinct(),
+                )
+            )
+            .distinct()
+        )
 
-        context["court_classes"] = CourtClass.objects.prefetch_related("courts")
         context["recent_causelists"] = (
             CauseList.objects.select_related("work")
             .prefetch_related("labels")
--- a/peachjam/views/causelist.py
+++ b/peachjam/views/causelist.py
@@ -23,6 +23,8 @@ from peachjam.views import BaseDocumentDetailView, FilteredDocumentListView
 from peachjam.views.courts import MonthMixin, RegistryMixin
 from peachjam.views.generic_views import YearMixin
 
+CACHE_DURATION = 60 * 60 * 24
+
 
 @registry.register_doc_type("causelist")
 class CauseListDetailView(BaseDocumentDetailView):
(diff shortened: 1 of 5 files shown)


peachjam/views/causelist.py
```diff
@@ -40,8 +41,18 @@ class CauseListListView(TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
+        # get only court classes and courts that have causelists
+        context["court_classes"] = (
```
longhotsummer, line 45: This is potentially expensive, and doesn't change a lot, just like court years. It'll be good to establish a caching pattern for this.

Maybe something like:

```
# this could also be on the CourtClass object
def get_court_classes_with_causelists():
  return CourtClass.objects.filter(courts__causelists__isnull=False).values_list('pk', flat=True)

...

def get_context_data(...):
  ...
  court_classes_with_causelists = cache.get_or_set('court_classes_with_causelists', get_court_classes_with_causelists)
  # similar for court pks
  return CourtClass.objects.filter(pk__in=court_classes_with_causelists).prefetch(...)
```

actlikewill: changed, praise the caching

peachjam/views/causelist.py
```diff
@@ -40,8 +41,12 @@ class CauseListListView(TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-
-        context["court_classes"] = CourtClass.objects.prefetch_related("courts")
+        # get only court classes and courts that have cause lists, and cache the result
+        court_classes_with_cause_lists = cache.get_or_set(
+            "court_classes_with_causelists",
+            CourtClass.get_court_classes_with_cause_lists(),
+        )
```
actlikewill, line 48: do we need a timeout?

longhotsummer: yes

peachjam/views/causelist.py
```diff
@@ -40,8 +41,12 @@ class CauseListListView(TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-
-        context["court_classes"] = CourtClass.objects.prefetch_related("courts")
+        # get only court classes and courts that have cause lists, and cache the result
+        court_classes_with_cause_lists = cache.get_or_set(
```
longhotsummer, line 45: remember, we use the file system as the cache. so the cached values have to be serialisable. i don't think it will work to try to cache a queryset.

actlikewill: ok, so we need to fully build out the object along with the corresponding courts?

actlikewill: what about casting the queryset to a list, is that enough or should I also break down the individual courtclasses  and related courts

longhotsummer: It looks like you can cache a queryset, no idea how that works: https://docs.djangoproject.com/en/4.2/ref/models/querysets/#pickling-querysets

peachjam/views/causelist.py
longhotsummer: set a default timeout (30  mins?) on the default cache in settings.py

peachjam/views/causelist.py
```diff
@@ -22,6 +23,8 @@
 from peachjam.views.courts import MonthMixin, RegistryMixin
 from peachjam.views.generic_views import YearMixin
 
+CACHE_DURATION = 60 * 60 * 24
```
longhotsummer, line 26: I think this is too high -- it means it will take a day before a new court will show up -- this call isn't THAT expensive! 30 minutes should be sufficient and  matches what we cache other pages for.

actlikewill: i've set the default timeout on settings.py

longhotsummer (approved):

PR #2128, opened 2024-10-30
bootstrap 5 must not conflict with bootstrap 4 in admin

longhotsummer:
This makes modals in admin area work again

image

also extra logging for ingestors


actlikewill (approved):
great

PR #2130, opened 2024-10-30
support google custom search engine

longhotsummer:
Via https://programmablesearchengine.google.com/

image

Also:

* show facet buttons when there are no search results, to let users remove them
* show facets above the filters, so they're more visible
* don't cramp the "Sort by" label

https://www.loom.com/share/6064480c05804c31b11af637b5ac8cde


actlikewill (approved):
nice, are we going to deploy google search on another lii?

longhotsummer:
> nice, are we going to deploy google search on another lii?

possibly, yes

PR #2132, opened 2024-10-31
Add years to cache

actlikewill:
This caches the years for the court and causelist views. We use the base view name for the cache key which is typically the court or registry name.
Fixes https://github.com/laws-africa/peachjam/issues/2115


longhotsummer (approved):

longhotsummer:
i don't think slugify is strictly necessary?

actlikewill:
> i don't think slugify is strictly necessary?

we get console warnings if we dont slugify, it doesn't like spaces

PR #2134, opened 2024-10-31
Add article attachment link

actlikewill:
image


peachjam/views/article.py
```diff
@@ -79,6 +80,23 @@ def get_context_data(self, **kwargs):
...
+            ArticleAttachment,
+            document__date=self.kwargs["date"],
+            document__slug=self.kwargs["slug"],
+            document__author__username=self.kwargs["author"],
+            pk=self.kwargs["pk"],
+            filename=self.kwargs["filename"],
+        )
+
+    def render_to_response(self, context, **response_kwargs):
+        return FileResponse(self.object.file, filename=self.object.filename)
```
longhotsummer, line 97: can you add a long-lived caching header hear? eg 1 year, like our other static files?

actlikewill: added

peachjam/admin.py
```diff
@@ -1153,6 +1153,39 @@ class ArticleAttachmentInline(BaseAttachmentFileInline):
     model = ArticleAttachment
     extra = 1
 
+    def attachment_url(self, obj):
+        return reverse(
```
longhotsummer, line 1157: It strikes me that this is the `get_absolute_url` for this attachment object.

longhotsummer:
Looks good. When we render the attachments in a list at the end of the article, do we need to use the new url?

peachjam/admin.py
actlikewill: right. I've added it as such and changed the links including in the template

longhotsummer (approved):

PR #2135, opened 2024-10-31
Do a phrase-match search on all basic fields

longhotsummer:
also:

* apply most_recent_first on the server automatically
* calculate click score when writing to the db
* serialise filters as a string for readability


peachjam_search/views.py
```diff
@@ -192,6 +192,14 @@ def build_basic_queries(self, request, view):
             for field in query_fields
         ]
 
+        if " " in self.query:
+            # do optimistic match-phrase queries for multi-word queries
+            for field, options in view.search_fields.items():
+                query = {"query": self.query, "slop": 0}
+                if "boost" in (options or {}):
+                    query["boost"] = options["boost"]
+                queries.append(MatchPhrase(**{field: query}))
```
actlikewill, line 201: is this similar to adding quotes on searches? but we now enforce it on the backend

actlikewill (approved):

peachjam_search/views.py
longhotsummer: yes, it's saying : try this search as a phrase and boost if it matches, but don't fail if it doesn't

PR #2144, opened 2024-11-05
All judgments view, labels facet

longhotsummer:
* new view for judgments from all courts
* link to all courts from top of judgments page
* short court groupings as cards -- the big win here is that the "all judgments" link can go into the card footer
* labels faceted, for exploring reported judgments

image

image

image

### DIFF
--- a/peachjam/views/courts.py
+++ b/peachjam/views/courts.py
@@ -177,28 +177,33 @@ class CourtDetailView(FilteredJudgmentView):
 
     @cached_property
     def court(self):
+        if self.kwargs.get("code") == "all":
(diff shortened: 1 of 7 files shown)


peachjam/views/courts.py
```diff
@@ -79,6 +79,24 @@ def add_facets(self, context):
                 "values": self.request.GET.getlist("judges"),
             }
 
+        if "labels" not in self.exclude_facets:
```
actlikewill, line 82: we need to discuss how to add this in KL

actlikewill (approved):
looks good. the courts page needed some attention.

PR #2145, opened 2024-11-05
Filter for article tags in admin

actlikewill:
image

This filters for article tags in the admin.

### DIFF
--- a/peachjam/admin.py
+++ b/peachjam/admin.py
@@ -1188,6 +1188,12 @@ class ArticleAttachmentInline(BaseAttachmentFileInline):
 class ArticleForm(forms.ModelForm):
     body = forms.CharField(widget=CKEditorWidget())
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


actlikewill:
@longhotsummer Not all liis have the tag in their db. Should we run a migration to add it?

peachjam/admin.py
```diff
@@ -1188,6 +1188,12 @@ def new_init(self, *args, **kwargs):
 class ArticleForm(forms.ModelForm):
     body = forms.CharField(widget=CKEditorWidget())
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        article_tags = Taxonomy.objects.filter(name="Article tags").first()
```
longhotsummer, line 1193: There's a helper method on taxonomy that will do this for you

longhotsummer: Sorry, it's on Article. And it'll create if necessary.

actlikewill: added

longhotsummer (approved):

Issue #2093, opened 2024-10-09
Subleg listing tab on legislation detail page must show repealed label, and be ordered.


longhotsummer:
Otherwise the listing under the principal Act on the listing page (which only shows in force) is different to the subleg listing page.

Replace the current table with a normal document table.

Sort by date (desc), and add the date column: so title, citation, date.

longhotsummer:
eg. https://lawlibrary.org.za/akn/za/act/1990/94/eng@2003-08-05

Issue #2090, opened 2024-10-07
Don't nest subleg in the all legislation view


longhotsummer:
It's confusing and creates double entries.

eg https://lawlibrary.org.za/legislation/all?q=&sort=title

Issue #2088, opened 2024-10-03
When filtering legislation, I want to be able to filter in the text box by citation


longhotsummer:
eg. "Cap 23". Note that in some jurisdictions they use "Cap. 23" and we should handle that.

The filtering field only searches the title - expand that to the citation field too.

I also suggest we place wildcards at spaces, so `foo bar` would search for `foo%bar` so it doesn't have to be an exact prefix match.

PR #2148, opened 2024-11-06
Legislation listing changes

actlikewill:
image

image

This has small changes to the legislation listing and detail views

closes https://github.com/laws-africa/peachjam/issues/2093
closes https://github.com/laws-africa/peachjam/issues/2088
closes https://github.com/laws-africa/peachjam/issues/2090

### DIFF
--- a/peachjam/forms.py
+++ b/peachjam/forms.py
@@ -224,7 +224,7 @@ class BaseDocumentFilterForm(forms.Form):
             queryset = queryset.filter(taxonomies__topic__slug__in=taxonomies)
 
         if filter_q and q and exclude != "q":
-            queryset = queryset.filter(title__icontains=q)
+            queryset = queryset.filter(Q(title__icontains=q) | Q(citation__icontains=q))
 
         return queryset
 
--- a/peachjam/forms.py
+++ b/peachjam/forms.py
@@ -224,7 +224,11 @@ class BaseDocumentFilterForm(forms.Form):
             queryset = queryset.filter(taxonomies__topic__slug__in=taxonomies)
 
         if filter_q and q and exclude != "q":
-            queryset = queryset.filter(Q(title__icontains=q) | Q(citation__icontains=q))
+            terms = q.split()
+            queries = Q()
+            for term in terms:
+                queries &= Q(Q(title__icontains=term) | Q(citation__icontains=term))
+            queryset = queryset.filter(queries)
 
         return queryset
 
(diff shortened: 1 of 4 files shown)


peachjam/forms.py
```diff
@@ -224,7 +224,7 @@ def filter_queryset(self, queryset, exclude=None, filter_q=False):
             queryset = queryset.filter(taxonomies__topic__slug__in=taxonomies)
 
         if filter_q and q and exclude != "q":
-            queryset = queryset.filter(title__icontains=q)
+            queryset = queryset.filter(Q(title__icontains=q) | Q(citation__icontains=q))
```
longhotsummer, line 227: If the field is "Cap. 7" and I search for "cap 7", it won't match. I think we should separate search terms with `%`, chatgpt suggests:

```python
from django.db.models import Q

def search_name(queryset, search_string):
    terms = search_string.split()
    q_objects = Q()
    for term in terms:
        q_objects &= Q(name__icontains=term)
    return queryset.filter(q_objects)
```

peachjam/templates/peachjam/_child_documents.html
```diff
@@ -1,22 +1,3 @@
...
-          <td>
-            <a href="{{ document.get_absolute_url }}">{{ document.title }}</a>
-          </td>
-          <td>{{ document.citation }}</td>
-        </tr>
-      {% endfor %}
-    </tbody>
-  </table>
-</div>
+{% include "peachjam/_document_table.html" with documents=child_documents doc_table_citations=True doc_table_title_label="Title" doc_table_date_label="Date" %}
```
longhotsummer, line 3: can you sanity check the sorting for the subleg? It should be date descending, with the same hoop-jumping you do for gazettes so that the numbering works correctly for documents with the same date

peachjam/forms.py
actlikewill: added

peachjam/views/legislation.py
```diff
@@ -277,12 +277,14 @@ def get_timeline(self):
...
+        docs_ids = (
             self.model.objects.filter(parent_work=self.object.work)
             .distinct("work_frbr_uri")
-            .order_by("work_frbr_uri", "-date")
+            .order_by("work_frbr_uri")
         )
         # now sort by title
+        docs = self.model.objects.filter(pk__in=docs_ids).order_by(
+            "-date", "-frbr_uri_number"
+        )
```
actlikewill, line 288: added this for the sorting. It was not working well because of the distinct work_frbr_uri. Its still not very good. as you can see from the screenshot. I think we need a better solution just for sorting.

actlikewill: sorting by work__frbr_uri_date is much better
image

longhotsummer (approved):

PR #2157, opened 2024-11-13
ordering on document nature; default number for generic doc to title slug

longhotsummer:
This simplifies imports and updates


actlikewill (dismissed):
ran into this yesterday

PR #2160, opened 2024-11-13
Bills

longhotsummer:
* bills
* custom properties (key-value pairs) for all documents
* set frbr_uri_number from title for all documents
* push prefetches for listing views right up to the top parent class, so that they are consistent and easily configurable

image

image

image


actlikewill (approved):
looks good

Issue #2113, opened 2024-10-23
Saved searches


actlikewill:

longhotsummer:
# New pages and URLs

## Search page

When a search is run, the Vue search app puts the search alert box into the page and uses HTMX (ajax?) to load the content of the box. This allows the server to check if a search alert for this search already exists.

* `GET /search/saved-searches/check?q=...`  renders the search alert box, customised if the user is logged in and if they have already saved this search

If the user is not logged in, that html has a normal bootstrap modal with the login prompt, no special HTMX or JS needed.

If the user is logged in, and they have not saved this search, the button is part of a form which POSTs to a new endpoint that is submitted with HTMX and the box is updated with the text for a newly created saved search alert.

* `POST /search/saved-searches`

If the user is logged in and they have saved this search before, the box has an Unsubscribe experience that deletes the saved search.

* `POST /search/saved-searches/<pk>/delete`

PR #2161, opened 2024-11-15
Search alerts

actlikewill:
closes https://github.com/laws-africa/peachjam/issues/2113


peachjam_search/models.py
```diff
@@ -77,3 +84,56 @@ class Meta:
...
+            self.last_alert = now()
+            self.save()
+
+    def find_new_hits(self):
+        from peachjam_search.views import DocumentSearchViewSet
+
+        factory = APIRequestFactory()
+        request = factory.get("/search/api/documents/")
+        request.user = self.user
+        params = f"q={self.q}&{self.filters}&created_at__gte={self.last_alert}"
```
longhotsummer, line 113: this won't encode self.q properly. Instead, build up the query dict by setting individual parameters directly so that they get encoded.

peachjam_search/models.py
```diff
@@ -77,3 +84,56 @@ class Meta:
...
+        hits = hits[:10]
+        context = {
+            "hits": hits,
+            "saved_search": self,
+            "site": Site.objects.get_current(),
+        }
+        html = render_to_string("peachjam_search/emails/search_alert.html", context)
+        plain_txt = render_to_string("peachjam_search/emails/search_alert.txt", context)
+
+        subject = settings.EMAIL_SUBJECT_PREFIX + _("New hits for search ") + self.q
```
longhotsummer, line 131:
```suggestion
        subject = settings.EMAIL_SUBJECT_PREFIX + _("New matches for search: ") + self.q
```

longhotsummer: The translation is an interesting consideration here, because we don't have a request to know what language to use. Ideally, we should store (and use) the user's preferred language, but we don't have that currently.

I guess this will use the default site language for now, which is fine (we only have English sites any way).

peachjam_search/templates/peachjam_search/emails/search_alert.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8"/>
+    <title>Search Alert</title>
```
longhotsummer, line 5: These should all be translateable.

peachjam_search/views.py
```diff
@@ -710,3 +713,11 @@ def get(self, request, *args, **kwargs):
 
     def has_permission(self):
         return self.request.user.is_authenticated and self.request.user.is_staff
+
+
+class SavedSearchCreateView(CreateAPIView):
+    permission_classes = (IsAuthenticated,)
```
longhotsummer, line 719: Please include the usual django model perms, too.

longhotsummer: Actually, because we're using htmx for the UI, this probably won't be required, since the forms-powered htmx view will handle this.

longhotsummer:
Looking good, I've added some comments for this draft phase.

peachjam_search/models.py
```diff
@@ -78,3 +85,59 @@ class Meta:
...
+            self.save()
+
+    def find_new_hits(self):
+        from peachjam_search.views import DocumentSearchViewSet
+
+        factory = APIRequestFactory()
+        request = factory.get("/search/api/documents/")
+        request.user = self.user
+        params = f"q={self.q}&{self.filters}&created_at__gte={self.last_alert.replace(tzinfo=None).isoformat()}"
+        request.GET = QueryDict(params)
```
longhotsummer, line 117: bulid up piece by piece

peachjam_search/models.py
```diff
@@ -78,3 +85,59 @@ class Meta:
...
+        request.id = "none"
```
longhotsummer, line 118: set to something meaningful like `search-alert-{pk}`, so that DocumentSearchViewSet can filter those out and not save them as a search trace.

peachjam_search/serializers.py
```diff
@@ -158,3 +159,9 @@ class SearchClickSerializer(ModelSerializer):
     class Meta:
         model = SearchClick
         fields = ("frbr_uri", "search_trace", "portion", "position")
+
+
+class SavedSearchSerializer(ModelSerializer):
```
longhotsummer, line 164: can fall away, using htmx instead

peachjam_search/templates/peachjam_search/saved_search_modal.html
```diff
@@ -0,0 +1,31 @@
...
+  <div class="modal-content">
+    <div class="modal-header">
+      <h1 id="savedSearchModalLabel" class="modal-title fs-5">{% trans 'Search alert' %}</h1>
+      <button type="button"
+              class="btn-close"
+              data-bs-dismiss="modal"
+              aria-label="Close"></button>
+    </div>
+    <div hx-swap-oob="true" id="search-modal-body" class="modal-body">
+      You are not logged in. Please log in to save your search and receive alerts.
```
longhotsummer, line 20: translateable

peachjam_search/views.py
```diff
@@ -753,3 +764,67 @@ def get(self, request, *args, **kwargs):
...
+            saved_search = SavedSearch.objects.filter(
+                user=self.request.user, q=q, filters=filters
+            ).first()
+            if saved_search:
+                return HttpResponseRedirect(
+                    reverse(
+                        "search:saved_search_update", kwargs={"pk": saved_search.pk}
+                    )
+                )
+            return HttpResponseRedirect(reverse("search:saved_search_create"))
```
longhotsummer, line 786: with `?foo=bar&q=baz`

peachjam_search/models.py
actlikewill: done, we use the get_filters_dict method to get the initial dict and then add the other items

peachjam_search/templates/peachjam_search/saved_search_modal.html
actlikewill: done

peachjam_search/models.py
actlikewill: updated

peachjam_search/models.py
actlikewill: noted, subject updated

peachjam_search/views.py
actlikewill: updated to use the model permissions only

peachjam_search/models.py
actlikewill: done, the viewset has also been updated to check  for the id and abort the save if it is a search alert

peachjam_search/serializers.py
actlikewill: removed

actlikewill:
@longhotsummer I've updated this with the requested changed. The only thing pending is the advanced search.

peachjam_search/models.py
```diff
@@ -78,3 +85,86 @@ class Meta:
...
+
+
+class SavedSearch(models.Model):
+    q = models.CharField(max_length=4098)
+    filters = models.CharField(max_length=4098, null=True, blank=True)
+    note = models.TextField(null=True, blank=True)
+    user = models.ForeignKey(
+        User, on_delete=models.CASCADE, related_name="saved_searches"
+    )
+    created_at = models.DateField(auto_now_add=True)
```
longhotsummer, line 97: We usually use a datetime field for `created_at`

peachjam_search/views.py
```diff
@@ -753,3 +770,99 @@ def get(self, request, *args, **kwargs):
...
+
+    def get_template_names(self):
+        if self.request.htmx:
+            return ["peachjam_search/_saved_search_list.html"]
+        return ["peachjam_search/saved_search_list.html"]
+
+    def post(self, *args, **kwargs):
+        self.object = self.get_object()
+        self.object.delete()
+        response = HttpResponse("saved search deleted")
```
longhotsummer, line 866: I think it's simpler not to use htmx at all here, there's no benefit, reloading the whole page is quick and deleting isn't a common operation. Less fuss.

actlikewill: we can maintain htmx if we plan on having a delete button on the search page modal,

longhotsummer (dismissed):

peachjam_search/views.py
```diff
@@ -766,4 +782,104 @@ def get(self, request, *args, **kwargs):
         return super().get(request, *args, **kwargs)
 
     def has_permission(self):
-        return self.request.user.is_authenticated and self.request.user.is_staff
+        return self.request.user.is_authenticated and self.request.user.is_staffo
```
longhotsummer, line 785: xx

peachjam_search/views.py
```diff
@@ -766,4 +782,104 @@ def get(self, request, *args, **kwargs):
...
+
+
+class AllowSavedSearchesMixin:
+    def dispatch(self, *args, **kwargs):
+        if not pj_settings().allow_save_searches:
+
```
longhotsummer, line 791:
```suggestion
```

peachjam_search/views.py
longhotsummer:
```suggestion
        return self.request.user.is_authenticated and self.request.user.is_staff
```

longhotsummer (approved):

peachjam/settings.py
actlikewill, line 1: @longhotsummer added the config,

longhotsummer (approved):

PR #2164, opened 2024-11-16
Certain topics in document listing views

longhotsummer:
Show selected taxonomy topics in listing views. This opens the door to showing summaries in listing views.

* adds a new `show_in_document_listing` boolean to topics; if that is set, then that topic will be included in document listings. To make configuration easier, changing this cascades down through descendants.
* pre-calculate a `path_name` field for topics, which is the names of each ancestor (except the first) down to the topic, for use in listing views
* inject topic path names into search results

image

image


peachjam/models/taxonomies.py
longhotsummer, line 1: line endings were corrected from CRLF

actlikewill (approved):
neat

PR #2166, opened 2024-11-18
fix lookups on tag name when language is not EN

longhotsummer:
otherwise, this tries to lookup, for example, name_fr__iexact=...

fixes https://lawsafrica.sentry.io/issues/6035652754/


actlikewill (approved):
this was a tricky one

longhotsummer:
> this was a tricky one

ja, I only realised it when looking through the SQL breadcrumbs.

PR #2169, opened 2024-11-18
articles can have embedded styles

longhotsummer:
also fix default ordering for articles

also:
* fix homepage inheritance so that LIIWeb's homepage builds on peachjam's homepage
* only count docs in AfricanLII, since it's expensive and not used elsewhere


peachjam/settings.py
```diff
@@ -569,7 +569,32 @@ def before_send(event, hint):
             ["Smiley", "SpecialChar", "LaAkn"],
             ["Source"],
         ],
-    }
+    },
+    "article": {
+        "removePlugins": ["iframe"],
+        "extraAllowedContent": "*[id];*{*}",
```
longhotsummer, line 575: this means that any tag can have any style details: `*{*}`

actlikewill (approved):
looks good. I'm going through the articles to see which ones have broken styling

PR #2174, opened 2024-11-20
Locality legislation

longhotsummer:
* liiweb now supports locality-based legislation views
* locality-based legislation views require a full place code, and redirect if necessary for backwards compatibility
* fix chunking so that we get columns of alphabetical chunks that collapse well on mobile
* OBL and by-law microsites now use `/localities` to standardise URLs

https://www.loom.com/share/12401742a2b34a82b03e69a47de6b0af


actlikewill (approved):
works well.

PR #2180, opened 2024-11-21
prevent creation of core documents in admin, and better errors if one…

longhotsummer:
… exists

otherwise, if a superuser creates a coredocument, it actually can't be seen anywhere.

also prevent creation of PeachJamSettings object since it's a singleton


actlikewill (approved):
great. this just came up today

PR #2188, opened 2024-11-22
added IDs to group titles

Sandravaphilips:
This is for https://github.com/laws-africa/kenyalaw-pj/issues/198
https://github.com/user-attachments/assets/46348ef7-616a-4662-80a7-b3d7820cbeb8

### DIFF
--- a/peachjam/views/generic_views.py
+++ b/peachjam/views/generic_views.py
@@ -33,8 +33,6 @@ from peachjam_api.serializers import (
     RelationshipSerializer,
 )
 
-months_reverse = {v: k for k, v in MONTHS.items()}
-
 
 class ClampedPaginator(Paginator):
     """A paginator that clamps the maximum number of pages. This is useful when dealing with very large datasets,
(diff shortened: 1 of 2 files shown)


peachjam/views/generic_views.py
```diff
@@ -109,6 +111,9 @@ class Group:
             def __init__(self, title):
                 self.title = title
 
+                if group_by == "date":
+                    self.group_id = months_reverse[title.split()[0]]
```
actlikewill, line 115: @Sandravaphilips I've altered the logic a bit here. the previous was giving an incorrect index, eg. it gave November the id of 10, instead of 11

actlikewill (approved):

peachjam/views/generic_views.py
longhotsummer: does this work in multiple languages?

actlikewill: modified this to allow the months to be translatable

longhotsummer (approved):

Issue #2173, opened 2024-11-20
As a user on ZanzibarLII, I would like to see the Court of Appeal of Tanzania - Zanzibar registry on ZanzibarLII


NotoriousMBB:
The collection needs to be displayed on both the Judgments listings page: https://zanzibarlii.org/judgments/
and in search results.

longhotsummer:
Currently judgments for this registry are on TanzLII, which we can consider authoritative. We need a mechanism to pull them into zanzibarlii, too.

longhotsummer:
# Proposal

A peachjam-to-peachjam judgment ingestion adapter. We already have the ratification adapter.

* the ingestor must be able to accept a court code to filter on
* the judgment API must be able to filter by court code
* the court is configured in ZanzibarLII
* the ingestor is configured on ZanzibarLII to fetch judgments into the court

The TanzLII website would remain authoritative for the court's judgments, and they would be mirrored to ZanzibarLII.

PR #2205, opened 2024-12-04
Judgment adapter

actlikewill:
closes https://github.com/laws-africa/peachjam/issues/2173


peachjam_api/serializers.py
```diff
@@ -160,12 +176,15 @@ class JudgmentSerializer(BaseSerializerMixin, serializers.ModelSerializer):
     court = CourtSerializer(read_only=True)
     url = serializers.SerializerMethodField()
     judges = serializers.SerializerMethodField()
-    case_numbers = serializers.SerializerMethodField()
+    case_numbers = CaseNumbersSerializer(many=True, read_only=True)
```
longhotsummer, line 179: pls sanity check that these three related fields are loaded with prefetch_related / select_releated

actlikewill: I've added the related lookups for case_numbers, judges and taxonomies

peachjam/adapters/judgments.py
```diff
@@ -0,0 +1,309 @@
...
+        for pair in (self.settings.get("filters") or "").split():
+            if "=" in pair:
+                key, value = pair.split("=")
+                self.filters[key] = value
+        self.court_codes = self.settings["court_code"].split()
+        self.filters["court__code"] = []
+        for code in self.court_codes:
+            self.filters["court__code"].append(code)
+
+    def client_request(self, method, url, **kwargs):
```
longhotsummer, line 50: this feels like it could either go on RequestsAdapter, or is repeating something already there

peachjam/adapters/judgments.py
```diff
@@ -0,0 +1,309 @@
...
+                raise e
+
+
+@plugins.register("ingestor-adapter")
+class JudgmentDeleteAdapter(BaseJudgmentAdapter):
+    def check_for_updates(self, last_refreshed):
+        from peachjam.tasks import get_deleted_documents
+
+        # kick of bg tasks in batches
+        judgments_count = Judgment.objects.count()
```
longhotsummer, line 259: shouldn't this be filtered by the court code? perhaps create a queryset once and then pass it into get_deleted_documents

peachjam/adapters/judgments.py
```diff
@@ -0,0 +1,309 @@
...
+            .values_list("expression_frbr_uri", flat=True)[range_start:range_end]
+        )
+        url = f"{self.api_url}/validate-expression-frbr-uris"
+        result = self.client_request(
+            "post", url, json={"expression_frbr_uris": list(expression_frbr_uris)}
+        ).json()
+        invalid_frbr_uris = result["invalid_expression_frbr_uris"]
+        log.info(f"found {len(invalid_frbr_uris)} documents to delete")
+
+        qs = Judgment.objects.filter(court__code__in=self.court_codes).values_list(
```
longhotsummer, line 287: this seems unnecessary -- why not just call delete_document with the invalid uris? If you really want to validate them to ensure the server didn't send you something back you didn't ask about, you can compare it with the submitted list, rather than going back to the db.

peachjam/adapters/judgments.py
```diff
@@ -0,0 +1,309 @@
...
+            "expression_frbr_uri", flat=True
+        )
+        # find invalid_frbr_uris that exist in the queryset
+        existing = set(invalid_frbr_uris) & set(qs)
+
+        for frbr_uri in list(existing):
+            delete_document(self.ingestor.id, frbr_uri)
+
+    def delete_document(self, expression_frbr_uri):
```
longhotsummer, line 296: this looks duplicated, can it go into the parent class and be re-used across both classes?

peachjam/adapters/judgments.py
longhotsummer: Oh, I see `get_deleted_documents` is the background task. This qs should still be filtered though, right?

peachjam_api/public_views.py
```diff
@@ -120,3 +160,31 @@ class RatificationsViewSet(viewsets.ReadOnlyModelViewSet):
         "updated_at": ["exact", "gte", "lte"],
     }
     ordering = ["updated_at"]
+
+
+class ValidateExpressionFrbrUrisView(APIView):
+    """Validates a list of decision expression FRBR URIs and returns those that aren't valid."""
+
+    permission_classes = (IsAuthenticated,)
```
longhotsummer, line 168: Is there an additional permission needed? Perhaps any of the per-content-type permissions?

actlikewill: hmm, i don't think we need any additional perms for this view, its a readonly view that returns a list of frbr uris

peachjam/adapters/judgments.py
actlikewill: done

peachjam/adapters/judgments.py
actlikewill: added the filter

peachjam/adapters/judgments.py
actlikewill: ok, i've moved the delete function to the base adapter

longhotsummer (approved):

PR #2222, opened 2025-01-10
explore court class footer

actlikewill:
slight tweak to the footer template to allow it to work with the kl tribunals hack

### DIFF
--- a/peachjam/templates/peachjam/_court_list.html
+++ b/peachjam/templates/peachjam/_court_list.html
@@ -23,7 +23,7 @@
           {% if court_class.show_listing_page %}
             <div class="card-footer">
               {% block court_class_link %}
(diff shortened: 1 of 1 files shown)


peachjam/templates/peachjam/_court_list.html
```diff
@@ -23,7 +23,7 @@ <h4 class="mb-2">{{ court_class.name }}</h4>
           {% if court_class.show_listing_page %}
             <div class="card-footer">
               {% block court_class_link %}
-                <a href="{% url 'court_class' court_class.slug %}">{% blocktrans with court=court_class %}Explore {{ court }}{% endblocktrans %} →</a>
```
longhotsummer, line 26: If you make this `court=court_class.name` then it should Just Work without the below.

longhotsummer (approved):

PR #2225, opened 2025-01-14
added email share

### LINKED ISSUE #2223: Add option to share documents via email
Currently, users have multiple document sharing options (WhatsApp, X, Facebook, LinkedIn). However, there is no option to share documents directly via email. This limitation may inconvenience users who rely on email for communication.

Image

- Add an "Email" sharing option alongside the existing sharing buttons.
- Clicking the email icon should open the user's default email client or display a pop-up form to send the document directly via email.

Sandravaphilips:
- Closes #2223 
- This is a draft to show how it currently works. If you'd like to change something, kindly let me know.
- There are also share buttons on the article detail page and share selection modal. Should I add it there as well?
- You can also mention other places you'd like me to add the email share to.

Screenshot 2025-01-14 at 10 20 13
Screenshot 2025-01-14 at 10 20 38

### DIFF
--- a/peachjam/js/components/DocumentContent/ShareSelectionModal.vue
+++ b/peachjam/js/components/DocumentContent/ShareSelectionModal.vue
@@ -53,6 +53,13 @@
               @click="modal.hide()"
             ><i class="bi bi-linkedin linkedin-forecolor share-icon" />
             </a>
+            <a
+              :href="`mailto:?subject=${$t('Take a look at this text from') + ' ' + truncatedText}&body=${text}%0A%0ASee more at ${encodeURIComponent(url) }`"
+              class="btn btn-link"
+              target="_blank"
+              @click="modal.hide()"
+            ><i class="bi bi-envelope-at-fill envelope-at-fill-forecolor share-icon" />
+            </a>
           </div>
         </div>
       </div>
@@ -69,6 +76,10 @@ export default {
   computed: {
     combined () {
       return encodeURIComponent(`${this.text} ${this.url}`)
+    },
+    truncatedText () {
+      const title = document.querySelector('.document-content')?.dataset?.title || '';
+      return title.split(' ').splice(0, 8).join(' ') + '...';
     }
   },
 
--- a/peachjam/js/components/DocumentContent/ShareSelectionModal.vue
+++ b/peachjam/js/components/DocumentContent/ShareSelectionModal.vue
@@ -54,7 +54,7 @@
             ><i class="bi bi-linkedin linkedin-forecolor share-icon" />
             </a>
             <a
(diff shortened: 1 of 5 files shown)


peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -51,6 +51,11 @@ <h1>{{ document.title }}</h1>
                  target="_blank">
                 <i class="bi bi-linkedin linkedin-forecolor share-icon"></i>
               </a>
+              <a href="mailto:example@exapmle.com?subject=Check out this document&body={{ request.build_absolute_uri }}"
```
longhotsummer, line 54:
```suggestion
              <a href="mailto:?subject=Take a look at this document from {{ APP_NAME }}: {{ document.title|truncatewords:7|escape }}&body={{ request.build_absolute_uri }}"
```

Please also make that "Take a look at this document from {{ APP_NAME }}" bit translateable.

peachjam/js/components/DocumentContent/ShareSelectionModal.vue
```diff
@@ -53,6 +53,13 @@
               @click="modal.hide()"
             ><i class="bi bi-linkedin linkedin-forecolor share-icon" />
             </a>
+            <a
+              :href="`mailto:?subject=${$t('Take a look at this text from') + ' ' + truncatedText}&body=${text}%0A%0ASee more at ${encodeURIComponent(url) }`"
```
longhotsummer, line 57: Nice. The `See more at` must be translateable.

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -51,6 +51,11 @@ <h1>{{ document.title }}</h1>
                  target="_blank">
                 <i class="bi bi-linkedin linkedin-forecolor share-icon"></i>
               </a>
+              <a href="mailto:?subject={% trans 'Take a look at this document from' %} {{ APP_NAME }}: {{ document.title|truncatewords:7|escape }}&body={{ request.build_absolute_uri }}"
```
longhotsummer, line 54: What if, in addition to the href, you put the subject in as `data-subject=...` and the body similarly, and then the vue code just fetches those attributes, rather than having to re-compute it in Vue? That way the translation all happens through Django.

Sandravaphilips: This is a much better way, thank you. Also, I don't think we'll need to add the body as an attribute since it's different in both cases. Right now, I'm using the _combined_ computed value as the body for the **ShareSelectionModal**
This is what it looks like now
Screenshot 2025-01-17 at 11 28 57

longhotsummer (approved):

longhotsummer:
Great, thank you!

PR #2259, opened 2025-02-03
Floating titlebar above document

longhotsummer:
This adds a floating titlebar at the top of a document that is shown when the user scrolls past the title of the document. This means the title is always visible, and gives us a place to put key action buttons (for now, just "save").

The navigation toggle in mobile moves into the title bar and takes up less room.

The save button is adjust so that it uses htmx's oob-swap functionality to replace all `.save-document-button` instances on the page, not just the one in the main toolbar.

The "save document" button text is adjusted to support smaller widths.

image

image

https://www.loom.com/share/019b164cde2242e7a624ab1bbb6a13cb

### DIFF
--- a/peachjam/js/components/floating-header.ts
+++ b/peachjam/js/components/floating-header.ts
@@ -11,7 +11,7 @@ export default class FloatingHeader {
   }
 
   init () {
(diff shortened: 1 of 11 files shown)


actlikewill (approved):
Looks good and a good enhancement

PR #2265, opened 2025-02-05
Copy citation button for floating toolbar

longhotsummer:
image

* the button is only visible on desktop, because there isn't enough room on mobile and the need for copying the citation on mobile is much less


actlikewill (approved):
looks good

PR #2267, opened 2025-02-05
Refactors facets

actlikewill:


peachjam/views/courts.py
```diff
@@ -125,40 +122,22 @@ def add_facets(self, context):
...
-        if "alphabet" not in self.exclude_facets:
-            context["facet_data"]["alphabet"] = {
-                "label": _("Alphabet"),
-                "type": "radio",
-                "options": [(a, a) for a in lowercase_alphabet()],
-                "values": self.request.GET.get("alphabet"),
-            }
+    def add_facets(self, context):
+        context["facet_data"] = {}
+        self.get_judges_facet(context)
```
longhotsummer, line 135: Minor: Use `add_...` rather than `get_...` because `get` usually means that there is no side-effect and the method returns a result.

longhotsummer (approved):

peachjam/tests/test_forms.py
```diff
@@ -9,8 +9,8 @@ def test_years_filter_with_single_year(self):
         response = self.client.get("/legislation/?years=2007")
         self.assertEqual(response.status_code, 200)
         self.assertEqual(
-            response.context["facet_data"]["years"]["options"],
```
longhotsummer, line 12: I don't understand why this would change so much.

peachjam/tests/test_views.py
```diff
@@ -170,8 +170,8 @@ def test_generic_document_listing(self):
             documents,
         )
         self.assertEqual(
-            [("2017", 2017)],
```
longhotsummer, line 173: this makes more sense :)

longhotsummer (approved):

PR #2273, opened 2025-02-07
Decorators

longhotsummer:
Decorators moves the site-specific label decoration into a mini plugin class, which can be overridden/changed by other sites.

Also make it possible to override the alternative names label in the document detail page.


actlikewill (approved):
cool. this will make customizing models much easier :100:

Issue #2149, opened 2024-11-06
Make private annotations/comments on a document feature


niiroobiro:
Find the write-up for this here:
https://docs.google.com/document/d/1CADWeX6xdeS_DkagKHZ5iZ6nsMpSLvCM9UVrfmuclnI/edit?usp=sharing
### **Make private annotations/comments on a document**
**Introduction:**
This personalization feature enables users to add private annotations and comments directly to documents (judgments, legislation) within the platform. These annotations and comments will be visible only to the user who created them, ensuring confidentiality and a personalized study or research experience. Annotations will also be accessible through both the document view and a central list of annotated documents
**_Target Audience:_**

- Legal researchers/journalists
- Students
- Candidate attorneys
- Legal practitioners
- Judges

### **User Need:**
As a user, I want to be able to add private annotations to documents to enhance my understanding, analysis, and future reference. These annotations should be private and accessible, allowing me to personalize my experience and capture my thoughts without sharing them publicly.
As a user with multiple annotated documents, I want a clear, intuitive view of my documents with annotation details so that I can quickly understand my annotation activity
### Proposed Solution:

- When users hover over text in a document, Text changes color and an annotation icon appears in the right sidebar, allowing them to add private annotations/notes.
- Documents are organized into folders
- Each document displays an annotation-specific icon
- A visible comment count appears next to each document and there is a clear visual indication of annotation presence and volume
- Access and modify annotations in-document at any time like we already have in Google Docs where users can edit and delete annotations within the document and retain full context and original document integrity
- **Export /Download annotations as a document or other suitable format for offline reference.( With comments embedded)-This will be implemented at a later time**
- Users should be able to search for annotations.

**Enhancements to Consider:**
_List View for Annotated Documents:_
Provide users with a dedicated section in their profile to view all documents they have annotated. This list view will allow users to quickly find and navigate to their annotated documents, enhancing usability for frequent researchers.
_Integration with Saved Documents:_
Annotated documents should automatically appear in the user's "Saved Documents" section. This ensures that annotations are linked to saved documents for streamlined access and organization.
_Helping a user realise they can add a comment_
This can be done by showing a pop-up when a user hovers over a paragraph / portion of the document to let them know they can add private notes.
### User Journey 1: Logged-In Users

- The user browses the platform and identifies a document of interest.
- When a user hovers over a section of text, Text changes color and a floating button appears on the right margin which follows the user as they scroll through the document with buttons to add an annotation and save the document.
[...]

PR #2278, opened 2025-02-11
Annotations

actlikewill:
https://www.loom.com/share/ead508d1821a4d8296819194dcb30e9e?sid=1aa95dcc-3a9e-4bc3-b229-81276b7f7380

closes https://github.com/laws-africa/peachjam/issues/2149

pending: 

- [x] delete annotations when document is unsaved, with warning
- [x] add warning when deleting annotations from document view  
- [x] add prefix and suffix to highlighted text
- [x] add username to new comment

https://www.loom.com/share/740a90284ab54fedb21a6f92dfac98c8?sid=fb775748-c4e0-4495-b884-3e0cd3bb281a


longhotsummer:
Looking good. I'm surprised we need separate AddAnnotation and Annotation components. I would think that Annotation can handle both a new and a saved annotation.

peachjam_api/views.py
```diff
@@ -29,6 +31,11 @@ class RelationshipViewSet(viewsets.ModelViewSet):
     serializer_class = RelationshipSerializer
 
 
+class AnnotationViewSet(viewsets.ModelViewSet):
```
longhotsummer, line 34: this will need to be restricted to the user making the api call.

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -478,5 +478,6 @@ <h5>
     {% if perms.peachjam.add_relationship %}{{ predicates_json|json_script:"predicates" }}{% endif %}
     {{ citation_links|json_script:"citation-links" }}
     {{ provision_relationships|json_script:"provision-relationships" }}
+    {{ annotations|json_script:"annotations" }}
```
longhotsummer, line 481: this will have to by loaded dynamically so it doesn't break caching

peachjam/models/annotation.py
```diff
@@ -0,0 +1,23 @@
...
+        "peachjam.CoreDocument",
+        related_name="annotations",
+        on_delete=models.CASCADE,
+        verbose_name=_("document"),
+    )
+    text = models.TextField(_("text"), null=False, blank=False)
+    target_id = models.CharField(_("target id"), max_length=1024, null=True, blank=True)
+    target_selectors = models.JSONField(
+        verbose_name=_("target selectors"), null=True, blank=True
+    )
```
longhotsummer, line 16: need a FK to the user, and a created_at timestamp.

peachjam_api/urls.py
```diff
@@ -12,6 +12,7 @@
 router.register(r"relationships", views.RelationshipViewSet, basename="relationships")
 router.register(r"works", views.WorksViewSet, basename="works")
 router.register(r"citation-links", views.CitationLinkViewSet, basename="citation-links")
+router.register(r"annotations", views.AnnotationViewSet, basename="annotations")
```
longhotsummer, line 15: this will have to be scoped to the document so the annotations can be loaded dynamically.

eg. `documents/<pk>/annotations`

peachjam/js/components/Annotations/AnnotationList.vue
```diff
@@ -0,0 +1,68 @@
...
+            throw new Error('Failed to fetch annotations');
+          }
+          return resp.json();
+        })
+        .then((annotations) => {
+          this.items = annotations.results;})
+        .catch((err) => {
+          console.error(err);
+        });
+    },
```
actlikewill, line 50: rather than loading the annoations from the document detail view, we can fetch them directly from the api endpoint, which solves the caching problem

peachjam/models/annotation.py
actlikewill: done

peachjam_api/views.py
actlikewill: done

peachjam_api/urls.py
longhotsummer: See how we do it in indigo - it's a clean and understandable URL pattern for nested REST resources: https://github.com/laws-africa/indigo/blob/master/indigo_api/urls.py#L14

peachjam_api/views.py
```diff
@@ -29,6 +31,24 @@ class RelationshipViewSet(viewsets.ModelViewSet):
...
+    queryset = Annotation.objects.all()
+    serializer_class = AnnotationSerializer
```
longhotsummer, line 36: It's good to always be explicit about permissions for these views. In this case, DjangoModelPermissions should suffice - it enforces authentication.

peachjam/js/components/Annotations/Annotation.vue
```diff
@@ -0,0 +1,149 @@
...
+        <p v-else>
+          {{ annotation.text }}
+        </p>
+        <div class="mt-2 text-end">
+          <button
+            v-if="editing"
+            class="ms-1 btn btn-sm btn-secondary"
+            @click="editing = false"
+          >
+            Cancel
```
longhotsummer, line 40: if the annotation is new, emit a delete event so that the list can delete it completely.

peachjam/js/components/Annotations/Annotation.vue
```diff
@@ -0,0 +1,149 @@
...
+          </button>
+
+          <button
+            v-if="editing"
+            class="ms-1 btn btn-sm btn-primary"
+            @click="saveAnnotation"
```
longhotsummer, line 46: Either do validation manually, or wrap this in a form and rely on client-side form validation but putting the event handler on the form's `submit` event.

peachjam_api/serializers.py
```diff
@@ -73,6 +74,12 @@ class Meta:
         ]
 
 
+class AnnotationSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Annotation
+        fields = ("id", "document", "text", "target_id", "target_selectors")
```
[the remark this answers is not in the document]
actlikewill: added

peachjam/js/components/Annotations/Annotation.vue
actlikewill: done, we now emit an event to remove the annotation fromt the list

peachjam/js/components/Annotations/Annotation.vue
actlikewill: added a form with validation and attached the function to the submit event

peachjam/js/components/Annotations/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+      });
+      this.manager.addProvider(this);
+    }
+
+    getButton (target: IRangeTarget): HTMLButtonElement | null {
+      const btn = document.createElement('button');
+      btn.className = 'btn btn-outline-secondary';
+      btn.type = 'button';
+      btn.innerText = 'Add annotation...';
```
longhotsummer, line 41: Perhaps `Comment...` with the comment icon? Also must be translatable.

longhotsummer (approved):

PR #2288, opened 2025-02-13
check for nature in context before removing for bills

actlikewill:
fixes https://github.com/laws-africa/peachjam/issues/2286

### DIFF
--- a/peachjam/views/bills.py
+++ b/peachjam/views/bills.py
@@ -28,7 +28,8 @@ class BillListView(FilteredDocumentListView):
         context["doc_table_show_court"] = False
         context["doc_table_show_jurisdiction"] = False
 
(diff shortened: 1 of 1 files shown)


peachjam/views/bills.py
```diff
@@ -28,7 +28,8 @@ def get_context_data(self, **kwargs):
         context["doc_table_show_court"] = False
         context["doc_table_show_jurisdiction"] = False
 
-        del context["facet_data"]["natures"]
+        if "nature" in context["facet_data"]:
```
longhotsummer, line 31: Let's just remove this completely -- IIRC it was only there so that "bill" wasn't the only one in the list, which is now fixed elsewhere.

longhotsummer (approved):

Issue #2231, opened 2025-01-16
Reconsider date used for legislation in listing views


longhotsummer:
The current date used is the expression date. To some this is confusing, because an Act from 1960s, say, can show up at the top of the list.

What date should be used instead - the publication date / commencement date (and then fall back to expression date if neither of those are available)?

goose-life:
I think the FRBR URI date / year is a good candidate, since all works have one and it's usually the date people associate them with — at least for legislation.

longhotsummer:
The work date is 99% of the time just a year, so we'd need to support that (or use 1999-01-01, for example). We should probably also consider whether we show a date in the legislation listing views at all.

longhotsummer:
The year facet would also need to be updated.

longhotsummer:
This is done.

PR #2295, opened 2025-02-18
Remove dates from legislation listing views

longhotsummer:
Where necessary, use the work year for facets and ordering.

https://www.loom.com/share/05b52faa01eb4d95b4ab7472b0bc2e67

* make filtering in the filter form more customisable
* use the year portion of `frbr_uri_date` rather than the `date` field in many legislation listing views
* add overrides for filtering and grouping in legislation list views
* facet by frbr_uri_date
* update `.doc-table` CSS so that it handles lack of a date column (previously this was always assumed to be present)
* bonus: add a tooltip and a group title for "Subsidiary legislation" in the expanded children view

image

image

### DIFF
--- a/peachjam/forms.py
+++ b/peachjam/forms.py
@@ -11,6 +11,7 @@ from django.core.exceptions import ValidationError
 from django.core.files import File
 from django.core.mail import send_mail
 from django.db.models import Q
+from django.db.models.functions.text import Substr
 from django.http import QueryDict
 from django.template.loader import render_to_string
 from django.utils.translation import gettext as _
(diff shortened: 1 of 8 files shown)


longhotsummer:
Close #2231

longhotsummer:
An alternative which would make this a little easier would to be add a new field called `frbr_uri_date_year` which is derived (at save time) from the `frbr_uri_date` and stored as an integer. That would make some of the faceting and queries a bit easier, but I don't think would have a huge performance impact.

goose-life (dismissed):

PR #2297, opened 2025-02-19
Replace django-elasticsearch-dsl-drf with vanilla django-elasticsearch-dsl implementation

longhotsummer:
The django-elasticsearch-dsl-drf library makes it pretty difficult for us to explicitly craft the elasticsearch query. In particular, it's impossible for use to add ES version 8 features like retrievers and RRF (hybrid) searches.

This moves the core logic (which we had largely implemented outside of elasticsearch-dsl-drf already) into a SearchEngine class.

This has the benefit that we can craft and call searches outside of DRF.

### DIFF
--- a/peachjam/settings.py
+++ b/peachjam/settings.py
@@ -58,7 +58,6 @@ INSTALLED_APPS = [
     "rest_framework.authtoken",
     "django_filters",
     "django_elasticsearch_dsl",
-    "django_elasticsearch_dsl_drf",
     "jazzmin",
     "dal",
     "dal_select2",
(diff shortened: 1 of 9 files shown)


actlikewill (approved):
everything looks in order. i'm kind of happy for this change. the magic that DRF was doing is now laid bare to see :smile:

longhotsummer:
> everything looks in order. i'm kind of happy for this change. the magic that DRF was doing is now laid bare to see 😄

Agreed! It turns out it really wasn't doing much for us anyway, and was a very frustrating library to build upon.

PR #2303, opened 2025-02-21
js strings extraction for translation

longhotsummer:


peachjam/js/locale/fr/translation.json
actlikewill, line 1: no translations?, i guess this should have been part of the pull request

longhotsummer: this is just the extraction part; once merged into main it goes to crowdin to be translated.

PR #2320, opened 2025-02-27
revert document citations view

actlikewill:


actlikewill:
@longhotsummer this simply reverts the view, I'm working out how to add the treatments according to what we discussed,

longhotsummer (approved):
If the view works correctly with this code (without treatments), let's merge it and deploy it, and then work out how to do it correctly with treatments.

PR #2322, opened 2025-02-28
Citation treatments

actlikewill:
This adds the citation treatments back. We fetch the treatments at the last minute as discussed.
It adds about 4 additional queries, this should scale better.

image


longhotsummer (approved):
nice!

Issue #2242, opened 2025-01-22
Alerts for new documents for courts/judges/province/municipality/subject areas you’re interested in


niiroobiro:
The idea behind this feature is that a user can indicate their interest in a variety of different things, and we'll notify them when a new document is added that is linked to that thing.

Conceptually, there are two aspects:

1. the user can choose to follow (or unfollow) a thing
2. we periodically check for documents with a "creation" date after the last alert was sent for the "thing", and send emails accordingly

Models that can be followed:

* Author
* Country (jurisdiction)
* Locality
* CourtClass
* Court
* CourtRegistry
* TaxonomyTopic

More can readily be added later.

# Proposal

### Following

On the detail page for each thing to be followed, a new button is added `(bell) Follow XXX` or `(bell) Unfollow XXX`.

The button is loaded dynamically by htmx just like the "save" button for documents, so that caching isn't impacted.

* Endpoint: `GET /account/following/button?taxonomy=10` (where the query params have the thing being followed and the PK

If the user is logged in, it checks if the user is following the thing and returns a follow or unfollow button as appropriate.

If the user is NOT logged in, it shows a follow button that shows the login modal (similar to search alerts).

Clicking the follow button sends a POST that sets up the following and redirects to the button url, to change the button to unfollow. 

* Endpoint: `POST /account/following/follow?taxonomy=10`
* Endpoint: `POST /account/following/unfollow?taxonomy=10` 

Because the buttons are standalone, we can later add them to any part of any page. For example, a judgment detail page could have a "Follow" button alongside the court name.

#### User profile

A new tab is added to the User profile page called "Following", that lists the things that the user is following, and includes an unfollow button. The behaviour of the button can be exactly the same as above. This would mean if the user accidentally unfollows, the entry is still there but now with a Follow button. If they refresh the page, it's gone.

* Endpoint: `GET /account/following`

### Emails

Sending one email per day reduces the number of emails sent. It could look a bit like this:

```
Dear Greg,

We've found new documents for areas that you're following.

High Court
----------

1. Foo
2. Bar
3. Baz

Education
---------

1. Boom
2. Baz

Manage what you're following (link)

Best,
The LII team
```

## Backend

A new UserFollowing model:

* user: FK
* court: FK
* taxonomy_topic: FK
* etc (see below)
* created_at: timestamp
* last_alerted_at: timestamp

[...]

PR #2351, opened 2025-03-12
Document alerts

actlikewill:
image

image

image

https://www.loom.com/share/09804cc30e5845c5a2339548ca31d9ae?sid=46c37f2a-8a90-4977-aafb-ade51f5180b9

closes https://github.com/laws-africa/peachjam/issues/2242


peachjam/models/user_following.py
```diff
@@ -0,0 +1,244 @@
...
+        on_delete=models.CASCADE,
+        related_name="followers",
+        verbose_name=_("taxonomy topic"),
+    )
+    last_alerted_at = models.DateTimeField(_("last alerted at"), null=True, blank=True)
+
+    created_at = models.DateTimeField(_("created at"), auto_now_add=True)
+
+    class Meta:
+        constraints = [
```
longhotsummer, line 82: Perhaps a unique constraint across all fields?

actlikewill: each followed object has to be unique together with a user. unless i'm not getting what you mean?

longhotsummer: the constraint `(user, court, author, court_class, ...)` etc. should do that for you.

actlikewill: hmm this wont work perfectly, since it will enforce uniqueness across all values combined.
eg (userA, courtA, authorA) and (userA, courtA, courtB) will be allowed.

longhotsummer: True, but we will never create objects like that. What we want to guard against is the user clicking "subscribe" very quickly (perhaps by mistake) and us creating two entries for the same court. Only one of the FK items will ever be non-null.

actlikewill: id rather leave the constraints explicit like this. it better reflects how we want the model to behave

peachjam/models/user_following.py
```diff
@@ -0,0 +1,255 @@
...
+            "user": user,
+            "site": Site.objects.get_current(),
+        }
+        html = render_to_string(
+            "peachjam/emails/user_following_alert_email.html", context
+        )
+        plain_txt = render_to_string(
+            "peachjam/emails/user_following_alert_email.txt", context
+        )
+        subject = settings.EMAIL_SUBJECT_PREFIX + _("New documents have been published")
```
longhotsummer, line 247: how do we handle translations for other alerts? because we don't have a request here to determine the correct language to use (I know we chatted about this before, I don't recall the outcome). Perhaps now is the time to have a "preferred language" setting on the user object?

peachjam/models/user_following.py
```diff
@@ -0,0 +1,255 @@
...
+        send_mail(
+            subject=subject,
+            message=plain_txt,
+            from_email=settings.DEFAULT_FROM_EMAIL,
+            recipient_list=[user.email],
+            fail_silently=False,
```
longhotsummer, line 253: I wonder if we should queue up background tasks for each user? Otherwise, if for some reason there's a dodgy email address for a user that causes `send_mail` to fail, then we'll never get past that user.

peachjam/views/courts.py
```diff
@@ -168,6 +169,12 @@ def get_base_queryset(self, exclude=None):
             qs = qs.filter(court=self.court)
         return qs
 
+    def get_user_follow_form(self):
+        form = UserFollowingForm(
+            initial={"court": self.court.pk, "user": self.request.user.pk}
```
longhotsummer, line 174: user must not be a part of the form

peachjam/views/user_following.py
```diff
@@ -0,0 +1,76 @@
...
+class UserFollowingButtonView(TemplateView):
+    template_name = "peachjam/user_following_button.html"
+
+    def get(self, *args, **kwargs):
+        form = UserFollowingForm(self.request.GET)
+        if form.is_valid():
+            follow = UserFollowing.objects.filter(
+                **form.cleaned_data, user=self.request.user
+            ).first()
+            if self.request.user.is_authenticated:
```
longhotsummer, line 38: check this before calling the db with self.request.user

peachjam/views/user_following.py
```diff
@@ -0,0 +1,76 @@
...
+            + f"?{self.request.GET.urlencode()}"
+        )
+
+
+class UserFollowingDeleteView(DeleteView):
+    model = UserFollowing
+    template_name = "peachjam/user_following_delete.html"
+
+    def get_success_url(self):
+        return reverse("user_following_button") + f"?{self.request.GET.urlencode()}"
```
longhotsummer, line 76: do URL params for delete matter?

peachjam/templates/peachjam/court_class_detail.html
```diff
@@ -6,3 +6,7 @@
 {% block month_list %}
   {% include "peachjam/_court_class_month_list.html" %}
 {% endblock %}
+{% block follow-button %}
```
longhotsummer, line 9: I think an include would be better than a block (or a block that wraps an include)

peachjam/templates/peachjam/user_following_button.html
```diff
@@ -0,0 +1,29 @@
...
+</button>
+<div class="modal" id="followModal" tabindex="-1">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title">{% trans 'Follow' %}</h5>
+        <button type="button"
+                class="btn-close"
+                data-bs-dismiss="modal"
+                aria-label="Close"></button>
```
longhotsummer, line 16: translateable

peachjam/templates/peachjam/user_following_button.html
```diff
@@ -0,0 +1,29 @@
...
+      </div>
+      <div class="modal-body">
+        <p>{% trans 'You are not logged in.' %}</p>
+        <p>{% trans 'To follow, you need to be logged in.' %}</p>
+      </div>
+      <div class="modal-footer">
+        <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">{% trans 'Close' %}</button>
+        <a href="{% url 'account_login' %}?next={{ request.path }}"
```
longhotsummer, line 24:
```suggestion
        <a href="{% url 'account_login' %}?next={{ request.get_full_path }}"
```

so we don't lose the query string, if any

peachjam/templates/peachjam/user_following_delete.html
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
+<form hx-post="{% url 'user_following_delete' object.pk %}?{{ request.GET.urlencode }}">
```
longhotsummer, line 2: is the query string part necessary?

actlikewill: yes, we use that to reload the button after unfollow.

peachjam/views/user_following.py
longhotsummer: Also, you need to filter the queryset to ensure that `UserFollowing.user == self.request.user`

peachjam/templates/peachjam/user_following_delete.html
longhotsummer: of course! 🙏

peachjam/models/user_following.py
actlikewill: added the preferred language to the user profile model, the override context allows us to set the language while rendering the email. also added the preferred language to the queryset.

peachjam/models/user_following.py
actlikewill: hmm we could do that, we could also set "fail silently" to True

peachjam/views/courts.py
actlikewill: we don't need the form in the view, removed

peachjam/views/user_following.py
actlikewill: done

peachjam/models/user_following.py
actlikewill: yes, lets queue up the tasks per user.

actlikewill: done

peachjam/templates/user_account/_tabs.html
```diff
@@ -18,6 +18,14 @@
         </a>
       </li>
     {% endif %}
+    {% if perms.peachjam.view_userfollowing %}
```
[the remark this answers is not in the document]
longhotsummer: Yes, I agree. I was also wondering how we decide whether to show a feature for an anonymous user, if we don't know that they'll have the perm once they're logged in.

peachjam/apps.py
```diff
@@ -8,17 +8,21 @@ class PeachJamConfig(AppConfig):
...
         import jazzmin.settings
+        from countries_plus.models import Country
 
         import peachjam.adapters  # noqa
         import peachjam.signals  # noqa
+        from peachjam.helpers import get_country_absolute_url
 
         jazzmin.settings.THEMES["peachjam"] = "stylesheets/peachjam-jazzmin.css"
 
+        Country.get_absolute_url = get_country_absolute_url
```
longhotsummer, line 19: sneaky

peachjam/models/core_document.py
```diff
@@ -111,6 +112,9 @@ def place_code(self):
     def __str__(self):
         return self.name
 
+    def get_absolute_url(self):
```
longhotsummer, line 115: I think we don't use these enough - good addition.

peachjam/templates/peachjam/user_following_list.html
```diff
@@ -0,0 +1,31 @@
...
+{% endblock %}
+{% block account-tabs %}
+  {% include 'user_account/_tabs.html' with active='following' %}
+{% endblock %}
+{% block account-content %}
+  <div class="mb-4">
+    <h2>{% trans 'Following' %}</h2>
+    <p>{% trans 'Receive email alerts when new documents are added to the following areas.' %}</p>
+    {% if not object_list|length %}
+      <p>{% trans 'You are not following any subjects' %}</p>
```
longhotsummer, line 14:
```suggestion
      <p>{% trans 'You are not following anything. Click the Follow button on pages across the site to start following.' %}</p>
```

actlikewill: done

peachjam/models/user_following.py
```diff
@@ -0,0 +1,268 @@
...
+        blank=True,
+        on_delete=models.CASCADE,
+        related_name="followers",
+        verbose_name=_("taxonomy"),
+    )
+    last_alerted_at = models.DateTimeField(
+        _("last alerted at"), null=True, blank=True, auto_now_add=True
+    )
+
+    created_at = models.DateTimeField(_("created at"), auto_now_add=True)
```
actlikewill, line 87: added auto_now_add=True to set the initial timestamp

longhotsummer (approved):

PR #2356, opened 2025-03-13
account view puts content first, move tabs to left

longhotsummer:
* adjust account page: this makes it easier to add and group tabs, and puts the user's content first
* move user dropdown into secondary nav, to the left of the search box. This means it has more room to expand to the right, doesn't take up room in the navbar, and is visible on mobile
* change wording to "My account" so it's a consistent width
* include user's login details in the dropdown

# navbar

image

Screenshot 2025-03-17 at 12 56 59

# my account

image

# my account mobile

On mobile the nav falls away and is replaced by links.

image


peachjam/models/settings.py
```diff
@@ -72,6 +72,11 @@ class PeachJamSettings(SingletonModel):
         default=False,
         help_text=_("Allow signups via social accounts"),
     )
+    allow_signups = models.BooleanField(
+        verbose_name=_("allow signups"),
+        default=True,
+        help_text=_("Allow users to create accounts"),
+    )
```
actlikewill, line 79: i wish we could disambiguate this and the other setting above. the two are doing similar things that could be rolled into one setting.

peachjam/templates/user_account/_tabs.html
```diff
@@ -28,23 +57,6 @@
...
-        <a href="{% url 'search:saved_search_list' %}"
-           class="nav-link {% if active == 'saved_searches' %}active{% endif %}">
-          <i class="bi bi-bell"></i>
-          {% trans "Search alerts" %}
-        </a>
-      </li>
-    {% endif %}
     <li class="nav-item">
       <a href="{% url 'document_access_group_list' %}"
          class="nav-link {% if active == 'doc_access_groups' %}active{% endif %}">
```
actlikewill, line 62: i think we could move this to the account group rather than settings?

peachjam/models/settings.py
longhotsummer: ja, I tried to find a combined setting in django-allauth, but couldn't find one. I've pushed a change to actually enforce this (based on https://docs.allauth.org/en/dev/account/adapter.html#allauth.account.adapter.DefaultAccountAdapter.is_open_for_signup )

PR #2360, opened 2025-03-17
Subscriptions support

longhotsummer:
This adds optional subscription support.

The subscriptions work as follows:

1. A feature is a permission that is included in a product. A feature is usually associated with multiple Django
   permissions which enable access to the feature.
2. A product is a collection of features, such as "Team Plan".
3. A pricing plan is a way to charge for a product, such as X per month.
4. A product offering is a product with a pricing plan. The same product may be offered under different pricing plans
   to reflect discounts, etc.
5. A subscription is a user's access to a product offering.

A default subscription can be configured that all new users get.

This does not define a mechanism for billing. This only allows an admin to define the products, their features and permissions, and link them to users.

## Screenshots

image


actlikewill (approved):
awesome, we will need to update docs with what features need what permissions typically

PR #2371, opened 2025-03-20
Add save document buttons to search results

longhotsummer:
* Make multiple save-doc buttons on one page possible
* All the modals and buttons are now identified by document id.
* The search view loads saved document buttons in bulk and injects them into the page.
* Streamline the saved document button view so that it doesn't redirect, which is slow. Instead, the button view decides which template to use.
* Combine the "not logged in" template into the "create" template.
* Add a "download" button directly to the search results page
* "Unsave" view doesn't force a full refresh, it simply re-renders the "create" button

https://www.loom.com/share/9096748d55354f08a65905300cf29292

## Screenshots

image


peachjam/js/components/FindDocuments/index.vue
```diff
@@ -927,6 +931,18 @@ export default {
       }
     },
 
+    loadSaveDocumentButtons () {
+      // use htmx to load and inject save-document buttons
+      // get document ids
+      const ids = this.searchInfo.results.map(result => result.id);
+      if (ids.length) {
+        const el = document.createElement('div');
```
longhotsummer, line 939: htmx wants an element to inject content into, so we create this. in reality, all the content is swapped into place by htmx, but i add this to the end of the document just to be safe.

peachjam/templates/peachjam/saved_document/_create.html
longhotsummer, line 1: This is copied from the old file - the only major changes are to handle non-authenticated users, and the same addition of document ids to various elements as in `_update.html`

peachjam/views/save_document.py
```diff
@@ -142,8 +188,8 @@ def get_success_url(self):
         )
 
 
-class SavedDocumentCreateView(BaseSavedDocumentFormView, CreateView):
-    template_name = "peachjam/saved_document_create.html"
+class SavedDocumentCreateView(SavedDocumentFormMixin, CreateView):
```
actlikewill, line 191: i noticed that we set the permission_required attribute but we forgot to add the permission required mixin for these views,

longhotsummer: good catch, added.

actlikewill (approved):
great addition

Issue #2368, opened 2025-03-20
Set user prefered language


actlikewill:
Allow the user to set their preferred language
- when changing the language on the site
- from their user profile

PR #2375, opened 2025-03-21
user preferred language

actlikewill:
closes https://github.com/laws-africa/peachjam/issues/2368


peachjam/forms.py
```diff
@@ -519,10 +521,35 @@ class PeachjamLoginForm(LoginForm):
...
 
-class UserForm(forms.ModelForm):
+class UserProfileForm(forms.Form):
+    first_name = forms.CharField(max_length=255, required=False)
+    last_name = forms.CharField(max_length=255, required=False)
+    preferred_language = forms.ModelChoiceField(
+        queryset=Language.objects.filter(iso_639_1__in=get_languages())
+    )
+
     class Meta:
```
longhotsummer, line 531: If it's no longer a ModelForm, then Meta does nothing and can be dropped completely.

peachjam/templates/user_account/edit.html
```diff
@@ -38,6 +38,24 @@ <h2>{% trans "Profile" %}</h2>
             </div>
           {% endif %}
         </div>
+        <div class="mb-3">
+          <label for="{{ form.preferred_language.id_for_label }}">{{ form.preferred_language.label }}</label>
+          <select class="form-control"
+                  name="{{ form.preferred_language.name }}"
+                  id="{{ form.preferred_language.id_for_label }}">
+            {% for language in form.preferred_language.field.choices %}
```
longhotsummer, line 46: You can usually just iterate like this:

```
{% for opt in form.preferred_language %}
  {{ opt }}
{% endfor %}
```

peachjam/views/accounts.py
```diff
@@ -16,17 +18,37 @@ class AccountsHomeView(LoginRequiredMixin, TemplateView):
     template_name = "user_account/home.html"
 
 
-class EditAccountView(LoginRequiredMixin, UpdateView):
+class EditAccountView(LoginRequiredMixin, FormView):
     authentication_required = True
     model = User
```
longhotsummer, line 23: This can fall away

peachjam/forms.py
actlikewill: done

peachjam/templates/user_account/edit.html
actlikewill: right, done

peachjam/views/accounts.py
actlikewill: done

longhotsummer (approved):

longhotsummer:
We should also set  the language when the user signs in.

peachjam/middleware.py
```diff
@@ -48,6 +49,14 @@ def process_request(self, request):
             del request.META["HTTP_ACCEPT_LANGUAGE"]
 
 
+class SetPreferredLanguageMiddleware(MiddlewareMixin):
+    def process_response(self, request, response):
+        if hasattr(request, "set_language"):
+            request.POST = {"language": request.set_language}
+            response = set_language(request)
+        return response
```
actlikewill, line 57: @longhotsummer i've added the middleware here to set the language when logging in. we make use of the set_language view. it does a redirect and checks the next parameter. this is better than setting the cookie manually

actlikewill: this redirect is breaking some tests :(

peachjam/middleware.py
```diff
@@ -48,6 +51,15 @@ def process_request(self, request):
             del request.META["HTTP_ACCEPT_LANGUAGE"]
 
 
+class SetPreferredLanguageMiddleware(MiddlewareMixin):
+    def process_response(self, request, response):
+
+        if hasattr(request, "set_language") and "test" not in sys.argv:
+            request.POST = {"language": request.set_language}
+            response = set_language(request)
```
longhotsummer, line 59: The only important thing that `set_language` does is set the cookie. We have a response object here which we can add a cookie to, without going through a weird redirect.

Why not just replace this code with:

```python
            response.set_cookie(
                settings.LANGUAGE_COOKIE_NAME,
                lang_code,
                max_age=settings.LANGUAGE_COOKIE_AGE,
                path=settings.LANGUAGE_COOKIE_PATH,
                domain=settings.LANGUAGE_COOKIE_DOMAIN,
                secure=settings.LANGUAGE_COOKIE_SECURE,
                httponly=settings.LANGUAGE_COOKIE_HTTPONLY,
                samesite=settings.LANGUAGE_COOKIE_SAMESITE,
            )
```

peachjam/views/accounts.py
```diff
@@ -16,17 +17,28 @@ class AccountsHomeView(LoginRequiredMixin, TemplateView):
...
     def get_success_url(self):
         return reverse("edit_account")
 
-    def get_object(self, queryset=None):
-        return self.request.user
+    def form_valid(self, form):
+        if form.is_valid():
+            user = form.save()
+            language_code = user.userprofile.preferred_language.iso_639_1
+            self.request.POST = {
```
longhotsummer, line 36: use your middleware here?

actlikewill: updated.

peachjam/middleware.py
```diff
@@ -48,6 +49,25 @@ def process_request(self, request):
             del request.META["HTTP_ACCEPT_LANGUAGE"]
 
 
+class SetPreferredLanguageMiddleware(MiddlewareMixin):
+    def process_response(self, request, response):
+        if hasattr(request, "set_language") or hasattr(response, "set_language"):
+            lang_code = getattr(request, "set_language", None) or getattr(
+                response, "set_language", None
+            )
```
actlikewill, line 57: i wish this was cleaner but here we have to check both request and response. the signal only has the request and the form valid method only has the response :smiling_face_with_tear:

actlikewill: changed this to only use request. it works okay

longhotsummer (approved):
great

Issue #2229, opened 2025-01-14
When sharing a document on social media I want an image coverpage with rich metadata


longhotsummer:
This would make links on social media much more visually engaging. It would be similar to what github generates for a repo: https://github.blog/open-source/git/framework-building-open-graph-images/

longhotsummer:
Done

PR #2377, opened 2025-03-24
Document social media image

longhotsummer:
Adds a view for a social media sharing image for a document. The generated image is similar to the document popup, with key details and the site URL and logo:

* place
* title
* citation
* date
* taxonomy topics

It is cached and re-used unless the template changes (which will also happen if any of the key document details change)

social-image


longhotsummer:
Closes https://github.com/laws-africa/peachjam/issues/2229

actlikewill (dismissed):
great

PR #2378, opened 2025-03-25
Store embeddings data in postgres

longhotsummer:
* Refactor embeddings into new `peachjam_ml` django app, which will house all the machine learning stuff.
* models to store content chunks and embeddings in psql, rather than treating elasticsearch as the main store
* DocumentEmbedding model calculated from the average of all the content chunks for the document
* temporary support to backfill embeddings from those already stored in elasticsearch

This builds the foundation for document similarity.


actlikewill (dismissed):
looks good, cant wait to see this in action

Issue #2362, opened 2025-03-18
Reading lists and course-specific collections


niiroobiro:

longhotsummer:
## Practical example of checking what documents to show from descendant topics

```
Courses (public)
└── My School (public)
    ├── Class A1 (restricted: "view_class_a1")
    ├── Class A2 (public)
    └── Class A3 (restricted: "view_class_a3")
        └── Math (public; inherits restriction from Class A3)
```

When the user views My School:

* Check “Class A1”: Has explicit restriction; user lacks permission → exclude.
* Check “Class A2”: No restriction → include.
* Check “Class A3”: Has explicit restriction; user lacks permission → exclude (and descendants also excluded implicitly).
* Check “Class A3 > Math”: Parent (“Class A3”) restricted → exclude implicitly.

Thus, user sees only:
 
* Documents directly tagged to My School.
* Documents tagged under visible descendants (Class A2).

actlikewill:
@longhotsummer here is a stab at some of the details

**Restricted flag** 

If we need to be able to filter out restricted document from lists, we need to add a flag to filter with. We already have a `restricted` flag that we can reuse. We can add extra logic in the detail view to check whether user has access for the taxonomy to allow access in the detail view. 
For listing views, we can filter out documents with the restricted flag. This would be a new change that would need to be communicated. 
 
**Restricted Taxonomies**

PR #2379, opened 2025-03-25
Restricted taxonomies

actlikewill:
https://www.loom.com/share/1eeb17c69ccf4af6ba9d3923e6e87ca9?sid=7955053e-5d31-4a12-8851-c013feaea6ba

closes https://github.com/laws-africa/peachjam/issues/2362


peachjam/migrations/0181_taxonomy_path_name.py
```diff
@@ -2,12 +2,10 @@
...
 from django.db import migrations, models
 
-from peachjam.models import Taxonomy
-
 
 def backfill_path_name(apps, schema_editor):
-    # deliberately use the actual Taxonomy model
-    for taxonomy in Taxonomy.get_root_nodes():
+    Taxonomy = apps.get_model("peachjam", "Taxonomy")
+    for taxonomy in Taxonomy.objects.iterator(chunk_size=100):
```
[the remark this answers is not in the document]
longhotsummer: At this point all our projects have had this migration applied. So we can safely make this whole migration a no-op.

peachjam/views/documents.py
```diff
@@ -69,16 +68,9 @@ def dispatch(self, request, *args, **kwargs):
...
+            if not request.user.has_perm("peachjam.view_document", obj):
                 return restricted_view.dispatch(request, *args, **kwargs)
 
-            if not request.user.is_superuser or not request.user.is_staff:
-                # check if user belongs to a group that has access to the document
-                doc_groups = get_groups_with_perms(obj)
-                user_groups = request.user.groups.all()
-                if not doc_groups.intersection(user_groups):
-                    return restricted_view.dispatch(request, *args, **kwargs)
-
```
actlikewill, line 81: changed this to use the object permissions directly, anonymous users don't have the permissions so we don't have to check for authentication

peachjam/views/taxonomy.py
```diff
@@ -11,32 +12,49 @@ class TaxonomyListView(TemplateView):
...
-class TaxonomyFirstLevelView(DetailView):
+class AllowedTaxonomyMixin:
+    def dispatch(self, request, *args, **kwargs):
+        self.taxonomy = self.get_taxonomy()
+        is_ancestor_restricted = self.taxonomy.get_ancestors().values_list(
+            "restricted", flat=True
+        )
+        if self.taxonomy.restricted or any(is_ancestor_restricted):
+            if not request.user.has_perm("peachjam.view_taxonomy", self.taxonomy):
+                return HttpResponseForbidden()
```
longhotsummer, line 28: This response indicates "this exists, but you can't see it". From a security standpoint, there's an argument that we shouldn't even indicate to someone that the thing exists.

So I propose that we just return a 404 here.

longhotsummer: Actually, I think that may cause more confusion than it's worth. Let's keep it like this.

peachjam/migrations/0181_taxonomy_path_name.py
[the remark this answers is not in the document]
actlikewill: great, i've removed the code entirely

PR #2387, opened 2025-03-28
Better AI-based judgment upload view

longhotsummer:
This moves the AI-extraction into the normal "add jugment" django admin view, and does the extraction via AJAX. This means that the user doesn't need to first save the judgment to see the results of the AI extraction. In the case that the extraction fails, the user can proceed as per usual.

To implement this I could have implemented a new form view (eg. with some JS and some HTMX), or I could use JS to manipulate the existing form view. I chose to go with the latter so it's less confusing for the user, even though it's a bit of hoop jumping.

The way it works is:

1. the uploaded file is submitted via ajax
2. the extraction is run
3. re-render the impacted parts of the form (the inputs and formsets) and send those to the client
4. JS on the client injects the updated inputs into the page (special casing for select2 selects)
5. JS also injects updated formsets into the page and triggers django's  formset management code

Also:

* removes the old upload view
* adds sha256 fingerprinting of source files to very quickly identify if a file upload is a duplicate. This applies to ALL files, not just judgments 

https://www.loom.com/share/42ead5ea939d434ab8ff15da0043fb09

image


peachjam/models/attachments.py
```diff
@@ -110,6 +110,9 @@ class SourceFile(AttachmentAbstractModel):
         null=True,
         blank=True,
     )
+    sha256 = models.CharField(
+        "SHA 256", max_length=64, null=True, blank=True, db_index=True
```
actlikewill, line 114: could we set this as unique and let db validation handle it?

actlikewill (approved):
looks good. i love that we have this feature!

peachjam/models/attachments.py
longhotsummer: That's a good suggestion. Maybe we can follow up once this is deployed.

PR #2390, opened 2025-03-28
Standardise user login/signup/reset layouts

longhotsummer:
* standardise the layouts of the various account creation/login flows
* pull in the site logo and the app name
* for the login flow, move create account close to login button, so that it's more obvious
* move password reset link closer to the password field
* signup form drops username and includes first and last names

Also:

* create UUID as tracking id for users
* push anonymised user id to matomo

# screenshots

image

image

image


lawlibrary/static/images/logo.png
longhotsummer, line 1: this is a symlink

actlikewill (approved):
looks good

Issue #2283, opened 2025-02-12
Add court division to judgment model


longhotsummer:
Add as a facet in search and listing pages.

Issue #2282, opened 2025-02-12
Add case verdict type to judgments


longhotsummer:
AKA Case Action.

Make a filter in the search and listing views.

PR #2406, opened 2025-04-04
Case verdicts and court divisions

actlikewill:
This adds case verdicts and court divisions to the judgment model. 
Both are added to listing page filters and the search facets.

image

image

closes https://github.com/laws-africa/peachjam/issues/2282
closes https://github.com/laws-africa/peachjam/issues/2283


peachjam/js/components/FindDocuments/index.vue
```diff
@@ -404,6 +413,13 @@ export default {
         value: [],
         options: []
       },
+      {
+        title: this.$t('Case verdicts'),
```
longhotsummer, line 417:
```suggestion
        title: this.$t('Case verdict'),
```

longhotsummer:
> @longhotsummer did we setup something to update search index mapping with single fields?

Let me check, it may break with the vector fields used for search since they're only enabled in some places.

peachjam_search/migrations/0015_es_mapping_add_division_action.py
```diff
@@ -0,0 +1,35 @@
...
+            print(f"Adding case action and court division mapping for {ix._name}")
+            ix.connection.indices.put_mapping(
+                index=ix._name,
+                body={
+                    "properties": {
+                        "case_action": ix._mapping["case_action"].to_dict(),
+                        "division": ix._mapping["division"].to_dict(),
+                    }
+                },
+            )
```
actlikewill, line 24: @longhotsummer i've added this. It updates the mapping but i still get an error in search because the individual documents don't have the fields updated. Do we still have to update the indexed documents?

image

longhotsummer: Backfilling is complicated in ES. Adjust the field on the serializer - you can use `allow_null=True`, provide a default, or set `required=False` and it should skip missing attributes.

longhotsummer (approved):

PR #2414, opened 2025-04-08
Follow button 

actlikewill:
image

image

closes https://github.com/laws-africa/peachjam/issues/2409


actlikewill:
@longhotsummer @niiroobiro This places the follow button below the title of the page which should make it more visible. We can compare engagement with the follow feature after this change. I wish there was a way we could do some A/B testing :smiling_face_with_tear:

longhotsummer (approved):
great, simpler and easier.

longhotsummer (approved):

Issue #2411, opened 2025-04-07
Saved search alerts can alert on very poor search matches


actlikewill:
Saved searches sometimes return poor results. We can explore changing how the search is performed to get better results

longhotsummer:
Example search: Border Management Authority Act, 2020

Search result alerted: https://lawlibrary.org.za/akn/za-gp/judgment/zagpjhc/2025/189/eng@2025-02-18 (doesn't appear in the first 10 pages)

longhotsummer:
What can happen is the following:

1. user creates a saved search that is fairly broad
2. a document is added that matches the search, but with a low score; normally the document would show up on page 2+ of the search results
3. when the search is run to find new matches, we only use the first page of results. However, we include a `created_at__gte=xxx` field in the search, so even the first page may contain a small number of poor matches
4. we send a search alert for a match that isn't a very good match

longhotsummer:
I propose that we remove the `created_at__gte` filter on the search, and instead check the `created_at` date of items in the first 2 pages of search results.

That should limit alerts to vaguely reasonable hits.

PR #2415, opened 2025-04-08
remove Saved search created_at filter

actlikewill:
This removes the created at filter from the saved search query. We now should get better search alerts. 
closes https://github.com/laws-africa/peachjam/issues/2411


peachjam_search/models.py
```diff
@@ -149,15 +149,23 @@ def find_new_hits(self):
...
                 params[key] = values
         params["search"] = self.q
-        params["created_at__gte"] = self.last_alerted_at.isoformat()
 
         engine = SearchEngine()
         form = SearchForm(params)
         form.is_valid()
         form.configure_engine(engine)
         es_response = engine.execute()
+        hits = es_response.hits[:20]
```
longhotsummer, line 158: If we want the first two pages then you'll have to override the default page size on the search engine object before executing the search.

longhotsummer (changes requested):

peachjam_search/models.py
actlikewill: ah right. changed

longhotsummer (approved):

PR #2417, opened 2025-04-09
unique constraint for relationships

actlikewill:
closes https://github.com/laws-africa/peachjam/issues/2385

We already had a unique constraint on the model. The problem arose with the null values.
Django 5+ has `nulls_distinct=True` but since we are on 4.2 we can remove the nulls and set the default to an empty string.

https://docs.djangoproject.com/en/5.2/ref/models/constraints/#nulls-distinct


longhotsummer (approved):
Interesting. Do you need to backfill anything?

actlikewill:
> Interesting. Do you need to backfill anything?

no. the migration will set the default value

Issue #2279, opened 2025-02-12
Change citing of Law Library to LawLibrary across the site


niiroobiro:
Feedback from external engagement is that citing it as a single word distinguishes it as a product name.

PR #2428, opened 2025-04-14
tanzlii and lawlibrary logo improvements

longhotsummer:
lawlibrary: transparent, bigger; better for embedding
tanzlii: narrower


longhotsummer:
Closes https://github.com/laws-africa/peachjam/issues/2279

Issue #2408, opened 2025-04-07
Find similar documents to the one I am reading


niiroobiro:
https://www.notion.so/Find-similar-documents-to-the-one-I-am-reading-120233d841b181d381a8d628f6578840?pvs=4

actlikewill:
@longhotsummer @niiroobiro See the above proposal for similar documents.

longhotsummer:
This is a good start. Some thoughts:

1. to avoid showing poor matches when there aren't any good ones, we can adjust the sql query to have a minimum score threshold (eg. if it ranges from 0 to 1, then the score should be at least 0.5)
2. I'm not sure an indicator of similarity is very useful to the user
3. We order documents by similarity score descending, right?
4. this may be slow to calculate; if so, we can populate the tab with htmx when it's clicked for the first time

actlikewill:
ui mockup: https://www.figma.com/proto/5lBHEej3KHw56thBGFXzcK/Similar-documents?node-id=1-2&p=f&t=6XrnUZ0qtlAfh490-1&scaling=scale-down&content-scaling=fixed&page-id=0%3A1

longhotsummer:
https://github.com/laws-africa/peachjam/pull/2461 adds an authority score which can be used to re-rank the top 20 similar documents.

PR #2439, opened 2025-04-22
Similar documents

actlikewill:
https://www.loom.com/share/98ecdbafdf4c4091af56bedc55b905aa?sid=8569e3e3-0f99-4503-9bfd-4025d96e4f19

https://github.com/laws-africa/peachjam/issues/2408


peachjam/views/documents.py
```diff
@@ -250,6 +253,36 @@ def get_context_data(self, **kwargs):
...
+class SimilarDocumentsView(DetailView):
+    template_name = "peachjam/_similar_documents.html"
+    slug_url_kwarg = "frbr_uri"
+    slug_field = "expression_frbr_uri"
+    model = CoreDocument
+
+    def get_similar_documents(self):
+        embedding = get_object_or_404(DocumentEmbedding, document_id=self.object.pk)
+        similar_docs = (
+            DocumentEmbedding.objects.exclude(document_id=self.object.pk)
```
longhotsummer, line 266: exclude this document's work, as well, to avoid getting historical versions of the current document.

lawlibrary/settings.py
```diff
@@ -3,7 +3,6 @@
 INSTALLED_APPS = [
     "lawlibrary.apps.LawlibraryConfig",
     "peachjam_subs",
-    "peachjam_ml",
```
longhotsummer, line 6: I had this as opt-in because it requires the database itself to have specific plugins enabled. We currently only have two websites that support this functionality.

peachjam/views/documents.py
```diff
@@ -250,6 +253,36 @@ def get_context_data(self, **kwargs):
         return context
 
 
+@method_decorator(add_slash_to_frbr_uri(), name="setup")
+class SimilarDocumentsView(DetailView):
+    template_name = "peachjam/_similar_documents.html"
+    slug_url_kwarg = "frbr_uri"
+    slug_field = "expression_frbr_uri"
+    model = CoreDocument
```
longhotsummer, line 261: Wrap this view in a permission (perhaps `peachjam_ml.view_documentembedding`) so that we can deploy it and limit access.

lawlibrary/settings.py
actlikewill: should we then also have the view itself live in this app? tests will fail otherwise

longhotsummer: yes, I think so for now. I'm not sure how complex that will get as we integrate similarity into other features, but let's start like that for now. It's not too complex to move later if necessary.

peachjam/views/documents.py
actlikewill: done

lawlibrary/settings.py
actlikewill: moved to peachjam ml

peachjam/views/documents.py
actlikewill: done

peachjam_ml/apps.py
```diff
@@ -7,3 +7,8 @@ class PeachjamMLConfig(AppConfig):
 
     def ready(self):
         import peachjam_ml.signals  # noqa
+        from peachjam.views import BaseDocumentDetailView
+
+        from .handler import modify_document_detail_context
+
+        BaseDocumentDetailView.modify_context.connect(modify_document_detail_context)
```
longhotsummer, line 14: this is fine for now, but usually this would be connected in signals.py with the usual decorators

peachjam_ml/templates/peachjam/layouts/document_detail.html
longhotsummer, line 1: I don't think this file is necessary?

peachjam_ml/templates/peachjam/_similar_documents.html
```diff
@@ -0,0 +1,11 @@
+{% load i18n %}
+<h4>{% trans 'Similar documents' %}</h4>
```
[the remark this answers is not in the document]
longhotsummer: The team felt pretty strongly about this

longhotsummer (changes requested):
minor changes :)

longhotsummer (approved):

PR #2442, opened 2025-04-22
templated emails

longhotsummer:
this makes it easier to apply common templates to emails, using django-templated-emails

text portions are generated automatically using html2text


longhotsummer:
This will allow us to create a Customer.io TemplateBackend which calls customer.io to send the email (if configured), rather than sending it directly.

Issue #2261, opened 2025-02-03
As an administrator I want to ensure that original file uploads are preserved (but private) when a judgment is anonymised


longhotsummer:
We need a record of the original, even if it's not publicly visible. There may be more than one file.

longhotsummer:
This is now correctly handled for anonymised files:

1. if the original document is not anonymised, a flag can be set on the `SourceFile` object and the file will be preserved, but not displayed to users
2. if the original document is anonymised, then it can be shown to users safely

To finish this, we need to add a `private` flag to the `AttachedFiles` model. That will control whether it is displayed or not, and will allow arbitrary hidden files for archiving purposes.

## Proposal

Add `private` boolean flag to `AttachedFiles` and don't show private files outside of the admin view at all.

longhotsummer:
@actlikewill this is good to go

PR #2453, opened 2025-04-24
Private flag for source files

actlikewill:
This adds the private flag to source files.
We hide the links on the detail page if a file is marked as private.
We return 404 for download urls if the file is private

closes https://github.com/laws-africa/peachjam/issues/2261


longhotsummer:
The flag only needs to be added to the AttachedFiles (sic) model. The SourceFile model has its own way of handling this, and it doesn't make sense for images or other abstracte files

actlikewill:
> The flag only needs to be added to the AttachedFiles (sic) model. The SourceFile model has its own way of handling this, and it doesn't make sense for images or other abstracte files

@longhotsummer I've set the private flag on the AttachedFile model only. The file is still accessible directly through its medial url though. Should we handle this as well?

longhotsummer:
> > The flag only needs to be added to the AttachedFiles (sic) model. The SourceFile model has its own way of handling this, and it doesn't make sense for images or other abstracte files
> 
> @longhotsummer I've set the private flag on the AttachedFile model only. The file is still accessible directly through its media url though. `media/doctype/attachments/pk/` Should we handle this as well?

Yes, private files must not be downloadable.

peachjam/views/documents.py
```diff
@@ -229,6 +229,24 @@ def render_to_response(self, context, **response_kwargs):
...
+            filename=self.kwargs["filename"]
+        ).first()
+        if file and not file.private:
+            if getattr(file.file.storage, "custom_domain", None):
+                # use the storage's custom domain to serve the file
+                return redirect(file.file.url)
+
+            file_bytes = file.file.open().read()
+            response = HttpResponse(file_bytes, content_type=file.mimetype)
+            response["Content-Disposition"] = f"attachment; filename={file.filename}"
```
longhotsummer, line 244: I have learnt the hard way that the filenames can be horribly dirty. Clean it up before sending it:

```python
filename = re.sub(r'[^A-Za-z0-9._-]', '', file.filename)
```

longhotsummer (changes requested):
great, minor change

longhotsummer (approved):
Is there a checkbox for the new flag in the admin area?

actlikewill:
> Is there a checkbox for the new flag in the admin area?

yes there is

PR #2461, opened 2025-04-28
ranking -> pagerank; calculate authority score

longhotsummer:
the authority score is a weighted combination of the normalised pagerank, incoming citations and recency.  this is useful when ordering document recommendations

also:

* renames Work.ranking to Work.pagerank
* moves ranker into peachjam.analysis


longhotsummer:
We can use this in combination with similarity scores when showing similar documents, which will help bubble authoritative documents to the top:

```
final_score = 0.7 * similarity_score + 0.3 * authority_score
```

actlikewill (approved):
great, curious on how you came up with this solution. Is it a standard approach on calculating page authority?

longhotsummer:
> great, curious on how you came up with this solution. Is it a standard approach on calculating page authority?

A weighted average is a common way to combine different scores. ChatGPT did the heavy lifting :)

PR #2465, opened 2025-04-29
use authority to re-rank similar documents and order citations

longhotsummer:
Get the top 100 documents by similarity, then re-rank using authority, and show the top 10.

Motivation for re-ranking, rather than folding the authority into the score during querying:

1.	Semantic similarity is the primary signal — the user expects the results to be on-topic first.
2.	Authority is a tie-breaker — we want to show the most valuable among the relevant.
3.	We avoid accidentally discarding high-similarity docs with lower authority.

Also:

1. sort citation lists by authority, rather than just citations


actlikewill (approved):
great,

Issue #2444, opened 2025-04-23
PJ refactor and maintenance


actlikewill:
We can use this issue to keep track of some smaller refactors and code maintenance issues come up so we can prioritize appropriately. They can be moved to their own issues if necessary. 

- Filtered documents view need a concept of a "filtered object", this will allow us to simplify some tempates 
- Move Base document view to documents.py from generic_views.py

PR #2468, opened 2025-05-01
split out urls.py into separate files

longhotsummer:
No URLs have been changed.


actlikewill (approved):
great, some sanity in the urls structure. I was compiling such maintenance tasks into an issue as we identify them. https://github.com/laws-africa/peachjam/issues/2444

PR #2471, opened 2025-05-02
Similar documents for folders

actlikewill:
This adds similar document recommendations to saved folders.

https://www.loom.com/share/af397a0d2da54c6faf53fcdb62fbe2d2?sid=ee2d8215-47c4-4d4e-83fa-76894d3f20b1

image

closes https://github.com/laws-africa/peachjam/issues/2408


longhotsummer:
@actlikewill screenshots or loom, pls

peachjam_ml/templates/peachjam/_similar_documents.html
```diff
@@ -1,5 +1,5 @@
 {% load i18n %}
-<h4>{% trans 'More documents like this one' %}</h4>
+<h4>{% trans 'Recommended documents' %}</h4>
```
actlikewill, line 2: we needed a more general title for these similar documents since they now appear in multiple places. I went with "Recommended documents". If this term is going to be used in other contexts we can change it

peachjam/templates/peachjam/saved_document/_folders_list.html
```diff
@@ -15,15 +15,56 @@ <h2>{% trans 'Saved documents' %}</h2>
...
+        <div class="card-header">
+          <ul class="nav nav-tabs card-header-tabs">
+            <li class="nav-item">
+              <button class="nav-link active"
+                      id="saved-doc-tab-{{ folder.pk }}"
+                      data-bs-toggle="tab"
+                      data-bs-target="#saved-docs-{{ folder.pk }}"
+                      type="button"
+                      role="tab"
+                      aria-controls="saved-doc-tab-{{ folder.pk }}"
```
longhotsummer, line 27: this and the data-bs-target should be the same element AFAIK

peachjam/templates/peachjam/saved_document/_folders_list.html
```diff
@@ -15,15 +15,56 @@ <h2>{% trans 'Saved documents' %}</h2>
...
-                {% include "peachjam/saved_document/_folder_list_item.html" %}
-              {% endfor %}
-            {% else %}
-              <p>{% trans 'No saved documents.' %}</p>
-            {% endif %}
+          <div class="tab-content">
+            <div class="tab-pane fade show active"
+                 id="saved-docs-{{ folder.pk }}"
+                 role="tabpanel"
+                 aria-labelledby="saved-doc-tab">
```
longhotsummer, line 53: needs to correspond to the ID of the tab

peachjam_ml/templates/peachjam/_similar_documents.html
longhotsummer: Another option is to move the heading out of this template so it can be customised where there's enough context to know how to phrase it

peachjam_ml/views.py
```diff
@@ -1,45 +1,30 @@
...
-            .filter(similarity__gt=self.similarity_threshold)
-            .values("title", "expression_frbr_uri", "similarity", "authority_score")
-            .order_by("-similarity")
+        doc_ids = self.get_doc_ids()
+        similar_docs = DocumentEmbedding.get_similar_documents(
+            doc_ids,
+            threshold=self.similarity_threshold,
         )[: self.top_k]
 
         # re-rank based on a weighted average of similarity and authority score, and keep the top 10
```
longhotsummer, line 30: I think you should push all of this into the document embedding class - that's the class that knows about ranking, scores, thresholds etc. The views don't need that and aren't customising it

longhotsummer: We're going to need it all later when sending recommendation emails, and it's domain logic not view logic

peachjam_ml/models.py
```diff
@@ -147,6 +147,50 @@ def refresh_for_document(cls, document):
 
         return doc_embedding
 
+    @classmethod
+    def get_average_embedding(cls, pks):
+        """Get the average embedding for a set of documents."""
+        if not pks:
+            return None
+
+        if len(pks) == 1:
```
longhotsummer, line 156: This is a premature optimisation. If there's only one, the db will handle it during Avg() etc

peachjam_ml/models.py
```diff
@@ -147,6 +147,50 @@ def refresh_for_document(cls, document):
...
+        return avg
+
+    @classmethod
+    def get_similar_documents(cls, doc_ids, threshold=0.0):
+        # always exclude the documents we are comparing against
+        exclude_ids = doc_ids
+        avg_embedding = cls.get_average_embedding(doc_ids)
+
+        similar_docs = (
+            DocumentEmbedding.objects.exclude(document__work__documents__in=exclude_ids)
```
longhotsummer, line 179: I don't think this works correctly? I think you need a sub query to find the works to exclude

peachjam/templates/peachjam/saved_document/_folders_list.html
```diff
@@ -15,15 +15,56 @@ <h2>{% trans 'Saved documents' %}</h2>
...
-              {% endfor %}
-            {% else %}
-              <p>{% trans 'No saved documents.' %}</p>
-            {% endif %}
+          <div class="tab-content">
+            <div class="tab-pane fade show active"
+                 id="saved-docs-{{ folder.pk }}"
+                 role="tabpanel"
+                 aria-labelledby="saved-doc-tab-{{ folder.pk }}">
+              {% if folder.saved_documents.all %}
```
longhotsummer, line 54: The .all goes to the db on this line and again on the next. Instead use a for loop and an empty tag at the end.

peachjam_ml/models.py
```diff
@@ -147,6 +147,50 @@ def refresh_for_document(cls, document):
...
+            doc_embedding = cls.objects.filter(document__pk=pks[0]).first()
+            if doc_embedding:
+                return doc_embedding.text_embedding
+
+        avg = (
+            DocumentEmbedding.objects.filter(document__in=pks)
+            .aggregate(avg=Avg("text_embedding"))
+            .get("avg")
+        )
+        if avg is not None and len(avg):
```
longhotsummer, line 167: This is the same as:

If avg:

peachjam/templates/peachjam/saved_document/_folders_list.html
actlikewill: done

peachjam/templates/peachjam/saved_document/_folders_list.html
actlikewill: done

peachjam_ml/views.py
actlikewill: I've moved everything to the DocumentEmbedding class. I've put all the configurations in the method, with some defaults that can overriden in the method call

peachjam_ml/templates/peachjam/_similar_documents.html
actlikewill: since we are loading via htmx, we'd need to have separate templates so we can handle empty responses properly. Let me know if we need to do that. Using a common title is easier for now.

peachjam_ml/models.py
actlikewill: i think the more common case is the document detail view which will have only one embedding. we don't need to be calculating the average for one document each time we call that view.

peachjam_ml/models.py
```diff
@@ -147,6 +147,50 @@ def refresh_for_document(cls, document):
 
         return doc_embedding
 
+    @classmethod
+    def get_average_embedding(cls, pks):
+        """Get the average embedding for a set of documents."""
+        if not pks:
```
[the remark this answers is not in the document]
actlikewill: ok, removed

peachjam_ml/models.py
actlikewill: this works as expected. it excludes all documents that we are using to calculate similarity

peachjam/templates/peachjam/saved_document/_folders_list.html
actlikewill: neat, i've never used the {% empty %} tag before

peachjam_ml/models.py
actlikewill: the value here is a numpy array, not a list. this is how we can check for truthiness for it

longhotsummer: ah of course

peachjam_ml/views.py
```diff
@@ -1,60 +1,42 @@
...
         return similar_docs
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context["similar_documents"] = self.get_similar_documents()
         return context
+
+
+@method_decorator(add_slash_to_frbr_uri(), name="setup")
+class SimilarDocumentsView(BaseSimilarDocumentsView):
```
longhotsummer, line 29: This mixin doesn't really help any more -- it's more work to get the classes to use it than just to have a single line in their `get_context_data()`:

```
        context["similar_documents"] = DocumentEmbedding.get_similar_documents([self.object.id])
        return context
```

peachjam/templates/peachjam/saved_document/_folders_list.html
```diff
@@ -15,15 +15,54 @@ <h2>{% trans 'Saved documents' %}</h2>
         <h5 class="card-header">
           <i class="bi bi-folder"></i> {{ folder.name }}
         </h5>
+        <div class="card-header">
```
longhotsummer, line 18: You've got a duplicate header - move the `ul` up into the header above and it'll look better. That's how it's designed to work.

Also, we should only show this if peachjam_ml is configured. We can either do template inheritance with overridden blocks, or we can assume that `peachjam` knows about the existence of `peachjam_ml` (like it knows about `peachjam_search`) and have something like `settings.PEACHJAM["PEACHJAM_ML"] = True` that gets injected into templates.

peachjam_ml/models.py
longhotsummer: It's not expensive to calculate the average of one, and it's all done in the DB. You can safely delete 6 lines of code -- less to test, less to go wrong.

peachjam_ml/models.py
longhotsummer: You'll have to show me -- we need to exclude all other documents that have the same works as the documents we're looking at. We don't want to include historical versions of a legislation document, for example.

peachjam_ml/views.py
actlikewill: ok, i've removed the base class. much cleaner

peachjam_ml/models.py
```diff
@@ -147,6 +148,55 @@ def refresh_for_document(cls, document):
...
+        )
+        if avg is not None and len(avg):
+            avg = normalize_vector(avg)
+
+        return avg
+
+    @classmethod
+    def get_similar_documents(cls, doc_ids, threshold=0.8, n_similar=10):
+        weight_similarity = 0.9
+        weight_authority = 0.1
```
actlikewill, line 168: adjusted the weights and threshold

peachjam/static/stylesheets/_global.scss
```diff
@@ -151,6 +151,10 @@ footer {
   font-size: 22px;
 }
 
+.text-purple-500 {
```
longhotsummer, line 154: Minor: I suggest making this semantic (eg `.text-ai`) and not the exact colour, and applying the same colour for the `pj-ai` icon, so that you can change that colour globally (eg. for ULII).

peachjam_ml/urls.py
```diff
@@ -1,14 +1,19 @@
...
 urlpatterns = [
     re_path(
         r"^(?P<frbr_uri>akn/.*)/similar-documents$",
-        cache_page(CACHE_DURATION)(SimilarDocumentsView.as_view()),
+        cache_page(CACHE_DURATION)(SimilarDocumentsDocumentDetailView.as_view()),
         name="document_similar_docs",
     ),
+    path(
+        "folder/<int:pk>/similar-documents",
+        cache_page(CACHE_DURATION)(SimilarDocumentsFolderView.as_view()),
```
longhotsummer, line 16: Do we want to cache this? If I add a document or two to my folder, then the suggestions must update.

longhotsummer: It's also not horribly expensive.

actlikewill: makes sense, removed

peachjam/static/stylesheets/_global.scss
actlikewill: agreed, changed

longhotsummer (approved):

PR #2483, opened 2025-05-08
Better PDF error handling

longhotsummer:
* show an error when the PDF fails to load, rather than letting the user watch a spinner go nowhere
* load the pdf data using fetch() so that we have visibility into what went wrong
* if the response is a non-200 response, then raise an exception so we get a sentry alert

image


actlikewill (approved):
looks good

PR #2485, opened 2025-05-09
Visual and layout tweaks

longhotsummer:
* move social login buttons to top of login and sign up pages, since they are by far the most used login mechanism
* reduce size of OLA header
* reduce padding on alerts to take up less room
* normalise margins on document title and reduce font size
* add an xxxl breakpoint at 1600px for very wide screens to take advantage of the extra width
* drop sticky-top on timeline, it looks weird

# screenshots

## login buttons

image


actlikewill (approved):
looks good

PR #2486, opened 2025-05-09
Fix bug that skewed legislation document embeddings

longhotsummer:
when we break an AKN document into chunks, we do it recursively. This
gives us good coverage and semantic search across provision boundaries
as well as accuracy when finding provision-level matches.

When we're calculating the average of all a document's chunks to find
the document-level embedding, we don't want to include the repeated
(nested) chunks because it gives them more weight than they should have.
Instead, use the average of the top-level provisions.

### DIFF
--- a/peachjam_ml/admin.py
+++ b/peachjam_ml/admin.py
@@ -1,21 +1,44 @@
 from django.utils.translation import gettext_lazy as _
 
 from peachjam.admin import DocumentAdmin
+from peachjam_ml.models import DocumentEmbedding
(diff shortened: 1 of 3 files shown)


longhotsummer:
An example of the impact would be that a judgment that quoted a few chunks of a particular section or chapter of a single piece of legislation would then be considered very similar to that legislation, because the quoted content lends an unreasonable weight to the embedding calculation for the legislation.

actlikewill (approved):
interesting

PR #2487, opened 2025-05-09
Adjust document detail page toolbar and use menu on mobile

longhotsummer:
* move Save button to the left and make it primary so it stands out
* make download button secondary
* on mobile, move share, download and report a problem into a dropdown menu to preserve space

## desktop

image

## mobile

image

image


actlikewill (approved):
great.

PR #2490, opened 2025-05-13
check for empty content before writing to file

goose-life:
resolves https://lawsafrica.sentry.io/issues/6156551493

### DIFF
--- a/peachjam/adapters/indigo.py
+++ b/peachjam/adapters/indigo.py
@@ -536,24 +536,31 @@ class IndigoAdapter(RequestsAdapter):
 
         logger.info(f"Downloading source file from {url}")
 
-        with NamedTemporaryFile() as f:
-            r = self.client_get(url)
-            try:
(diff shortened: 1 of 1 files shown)


longhotsummer:
I don't understand how this fixes the issue. Indigo should never be returning an empty file in a successful (200-level) response. It should either be a full file or a URL. Any idea what's going on?

goose-life:
https://github.com/laws-africa/indigo-lawsafrica/issues/1557 is open for the indigo part of it. I've put my findings so far there, but I don't have a solution nor do I understand what went wrong.

goose-life:
So this doesn't address the root cause, but it fixes the issue of presenting users with a link that takes them to a 500-error page.
It also gives us something to search for in the logs to identify the odd case where a publication document PDF didn't save successfully when being serialized.

longhotsummer:
This is no longer needed

PR #2494, opened 2025-05-15
As a user I want to be able to save documents to multiple folders

### LINKED ISSUE #2464: As a user I want to be able to save documents to multiple folders
Additionally, let's enforce that a document must always be in a folder.

## Proposal

1. make SavedDocument many-to-many with folders
2. when saving a document, ensure that at least one folder always exists - if necessary, create one called "My documents"
3. when saving a document (no folder is specified upon first save), default to saving the document to the folder of the most-recently-saved document
4. create a migration to ensure all users with saved docs that aren't in a folder get moved into a new "My documents" folder
5. in the saved document modal, change the folder dropdown to a list of folder names with checkboxes next to them. use a reasonable `max-height` style and scroll the overflow
6. in the saved document modal, drop the `New folder name` input label and instead use placeholder text.
7. update the saved document folder view to not have to worry about documents without a folder
8. update the folder deletion action so that it only unsaves documents that aren't saved to any other folder

Additionally,

1. in the folder listing view, move the "download", "delete" and "rename" options into a three-dots dropdown menu in the top-right of the folder, in line with the folder title

Image

Sandravaphilips:
Closes #2464 
https://www.loom.com/share/531e12d87e9a4967b313dbe2f5cacb10?sid=034737f2-59e7-47c3-a573-8da6078e2801


longhotsummer (approved):

longhotsummer:
I've pushed a final migration to remove the old folder field, and tweak some layouts.

looks great!

PR #2509, opened 2025-05-22
limit taxonomies to lowest items only

longhotsummer:
This fixes a situation like this where "TanzLII Case Digest" is repeated:

image

which is caused by this:

image

it should be this:

image


actlikewill (approved):
looks good

PR #2513, opened 2025-05-26
Add scroll css to nav tabs

actlikewill:


actlikewill:
> Cool. Can scroll-xs fall away?

yes, i've pulled the extra li item styling into nav tabs so it now works correctly.

longhotsummer (approved):

PR #2525, opened 2025-06-03
As a user I want to be encouraged to sign up to see similar documents

### LINKED ISSUE #2479: As a user I want to be encouraged to sign up to see similar documents

Sandravaphilips:
Closes #2479 
https://www.loom.com/share/1726cde8d98f45c0afd24e85426c6910?sid=a5d12138-3602-4d3c-9514-74d6650dd377

### DIFF
--- a/peachjam/templates/peachjam/_related_documents.html
+++ b/peachjam/templates/peachjam/_related_documents.html
@@ -38,4 +38,10 @@
     </div>
   {% endif %}
 </div>
-{% block extra-related-documents %}{% endblock %}
+{% if user.is_authenticated %}
+  {% block extra-related-documents %}{% endblock %}
+{% else %}
+  <p>
+    <a href="{% url 'account_signup' %}?next={{ request.get_full_path|urlencode }}">{% trans "Sign up" %}</a> or <a href="{% url 'account_login' %}?next={{ request.get_full_path|urlencode }}">{% trans "Log in" %}</a> to see similar documents.
+  </p>
+{% endif %}
--- a/peachjam_ml/apps.py
+++ b/peachjam_ml/apps.py
@@ -4,11 +4,3 @@ from django.apps import AppConfig
 class PeachjamMLConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "peachjam_ml"
-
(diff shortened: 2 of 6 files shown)


peachjam_ml/apps.py
```diff
@@ -4,11 +4,3 @@
...
     default_auto_field = "django.db.models.BigAutoField"
     name = "peachjam_ml"
-
-    def ready(self):
-        import peachjam_ml.signals  # noqa
-        from peachjam.views import BaseDocumentDetailView
-
-        from .handler import modify_document_detail_context
-
-        BaseDocumentDetailView.modify_context.connect(modify_document_detail_context)
```
actlikewill, line 14: @Sandravaphilips we don't want to remove the permissions entirely, we only want to check if the user is logged in to allow them to view similar documents

peachjam_ml/handler.py
```diff
@@ -1,6 +1,7 @@
 def modify_document_detail_context(sender, context, view, **kwargs):
     if not context["show_related_documents"]:
-        if view.request.user.has_perm("peachjam_ml.view_documentembedding") and hasattr(
-            context["document"], "embedding"
-        ):
+        if (
+            not view.request.user.is_authenticated
+            or view.request.user.has_perm("peachjam_ml.view_documentembedding")
+        ) and hasattr(context["document"], "embedding"):
```
actlikewill, line 6: is it possible to do this entirely in the template? I don't like the fact that we spreading this logic to this module,

actlikewill (approved):

peachjam/templates/peachjam/_related_documents.html
```diff
@@ -38,4 +38,10 @@ <h4>{% translate verb as verb %}{{ verb|capfirst }}</h4>
     </div>
   {% endif %}
 </div>
-{% block extra-related-documents %}{% endblock %}
+{% if user.is_authenticated %}
+  {% block extra-related-documents %}{% endblock %}
+{% else %}
+  <p>
```
longhotsummer, line 44: @Sandravaphilips this needs to be fully translatable. You'll need to store the URLs in variables and then use a blocktrans tag.

PR #2545, opened 2025-06-19
move published field to key details

actlikewill:
Editors cannot republish anonymized documents after anonymisation because they don't have access to the advanced tab. I think its safe to allow editors to have access to this field

### DIFF
--- a/peachjam/admin.py
+++ b/peachjam/admin.py
@@ -623,6 +623,7 @@ class DocumentAdmin(AccessGroupMixin, BaseAdmin):
                     "locality",
                     "date",
                     "language",
+                    "published",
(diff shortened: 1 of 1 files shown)


actlikewill:
@longhotsummer what workflow makes sense in this case? should we be setting these fields automatically after anonymisation>

longhotsummer (approved):

longhotsummer:
> @longhotsummer what workflow makes sense in this case? should we be setting these fields automatically after anonymisation?

the anonymisation tool does this automatically (it says "anonymised" and "published" when they click "publish" in the tool). But this change is NB as well.

PR #2547, opened 2025-06-20
strip remarks from akn html before indexing provisions for search

longhotsummer:
This will help to avoid annoying situations like this where the remarks are matched on searches (which is 99% of the time not wanted)

image

we will have to backfill existing content (including recalculating embeddings for sites that use them)


actlikewill (approved):

longhotsummer:
For tanzlii and lawlibrary (which use peachjam_ml), we must run:

```python
doc_ids = CoreDocument.objects.filter(content_html_is_akn=True, embedding__isnull=False).values_list("id", flat=True)

DocumentEmbedding.objects.filter(document_id__in=doc_ids).update(content_text_md5=None)

# queue background updates
for doc_id in doc_ids:
    update_document_embeddings(doc_id, schedule=5)
```

longhotsummer:
For other LIIs, all legislation must be re-indexed for search

PR #2574, opened 2025-07-02
Uncommenced provisions tweaks

goose-life:
- [x] Point to the Uncommenced provisions tab, not the Document detail tab, in the notice
- [x] ~Darker grey highlight~
- [x] Amber highlight instead — match 'warning' when active, lighter when not
- [x] Similarly for unconstitutional and 'danger'
    - [x] Resolved unconstitutional provisions should probably be ~'info'~ 'primary'
- [x] Color the gutter items too (warning / danger)
- [x] Icon: warning triangle (and for unconstitutional provisions)
- [x] Link to global list from legislation list

image

image

### DIFF
--- a/liiweb/templates/liiweb/legislation_list.html
+++ b/liiweb/templates/liiweb/legislation_list.html
@@ -24,13 +24,11 @@
               <a href="{% url "unconstitutional_provision_list" %}">{% trans "Unconstitutional provisions" %}</a>
             </li>
           {% endif %}
-          {% comment %}
           {% if show_uncommenced_provisions %}
             <li>
               <a href="{% url "uncommenced_provision_list" %}">{% trans "Uncommenced provisions" %}</a>
             </li>
           {% endif %}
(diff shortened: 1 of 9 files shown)


goose-life:
@longhotsummer I need a bit of help finalising the list above 🙏

goose-life:
One option (not committed, colours need work to match up nicely):

image

image

goose-life:
https://www.loom.com/share/6f6a54f849444df69510a207d4482165

longhotsummer:
I think this UX is really great. The colours look good and are well aligned. The mobile colour icons really stand out nicely.

Perhaps for unconstitutional elements that are resolved, since they're now informational rather than warnings:

* use an `(i)` icon (i in a circle) rather than the warning symbol
* card has a normal white background, not green -- otherwise the other informational ones should also be green (which I think is a bad idea!)

so we only use red and amber card backgrounds.

PR #2586, opened 2025-07-15
show incoming citation counts (WIP)

longhotsummer:
This is ready to merge into @actlikewill's branch with the provision citation logic.

It needs a bootstrap icon rather than the emoji, but I didn't have access to that on the plane.


actlikewill (approved):
looks good. let me merge

PR #2588, opened 2025-07-15
Show judgment summary in document sidebar

### LINKED ISSUE #2584: Show judgment summary in document sidebar
The idea here is that putting the judgment summary in the sidebar (as the default tab) is more useful to desktop users than the "Table of Contents" which is often empty for judgments.

# Proposal

1. Add a new "Summary" tab to the sidebar  for judgments only
2. The new tab comes first and is activated by default
3. Copy (not move) the summary into that tab
4. Show the flynote and blurb to everyone, show the full summary to only those with the correct perms (same behaviour as how the current summary is shown)

Example:

Image

Sandravaphilips:
Closes #2584 
Screenshot 2025-07-15 at 12 52 35


peachjam/templates/peachjam/_judgment_summary.html
longhotsummer, line 1: pls move this into `templates/peachjam/document/` so that we're slowly moving document templates there

peachjam/templates/peachjam/_document_content_sidebar.html
```diff
@@ -10,6 +10,18 @@
...
     <nav>
       <div class="nav nav-tabs" id="navigation-tab" role="tablist">
+        {% if document.doc_type == "judgment" and document.case_summary %}
+          <button class="nav-link"
+                  id="doc-summary-tab"
+                  data-bs-toggle="tab"
+                  data-bs-target="#doc-summary"
+                  type="button"
+                  role="tab"
+                  aria-controls="nav-sum"
```
longhotsummer, line 20: this must be the id of the summary tab-pane (ie `aria-controls="doc-summary"`)

the `aria-controls` of the other tabs should also match their tab pane ids

PR #2589, opened 2025-07-17
clean html fields before save

actlikewill:
This cleans the html fields before save. 
Ulii has an where case summary fields have invalid data during judgment upload. Not sure how its happening but it is preventing automatic judgment summary generation.

here is the data in metabase:
image


longhotsummer (approved):
nice, i was wondering what's going on.

so many was to have "no" content

PR #2607, opened 2025-07-25
search summary, flynote and blurb

longhotsummer:
* adds blurb and a (combined) summary field for indexing
* adds flynote, summary blurb to search fields


peachjam_search/engine.py
```diff
@@ -63,8 +63,10 @@ class SearchEngine:
             "content",
             "content_chunks",
             "case_summary",
```
actlikewill, line 65: not sure if this will affect the excludes but you removed case_summary has been removed from the fields

longhotsummer: i've changed it to a list of includes

PR #2624, opened 2025-08-08
quadruple max_length for read_in_text_html

goose-life:


longhotsummer (approved):
or make it a textfield

PR #2632, opened 2025-08-15
Timeline events

actlikewill:


actlikewill:
- send emails per user as bg task
- add edge case tests

peachjam/models/user_following.py
```diff
@@ -211,19 +218,22 @@ def get_new_followed_documents(self):
...
+
+                # add to timeline
+                timeline_event = TimelineEvent.objects.create(
+                    user_following=follow,
+                    event_type="new_documents",
+                )
+                timeline_event.subject_documents.set(new["documents"])
 
     @classmethod
     def send_alert(cls, user, followed_documents):
```
longhotsummer, line 239: so this would fall away?

peachjam/tasks.py
```diff
@@ -263,3 +263,12 @@ def generate_judgment_summary(doc_id):
         return
     log.info(f"Summarizing judgment {doc_id}")
     doc.generate_summary()
+
+
+@background(queue="peachjam", remove_existing_tasks=True)
+def send_timeline_emails():
```
longhotsummer, line 269: When does this get triggered? Daily?

As discussed, I think this should queue up separate tasks for each user that should be notified.

peachjam/views/my.py
```diff
@@ -65,13 +63,10 @@ def get_context_data(self, **kwargs):
...
-            50,
-            before_date,
+        timeline, next_before = TimelineEvent.get_events(
+            self.request.user, before=before_date
         )
-        if timeline:
-            context["before_date"] = list(timeline.keys())[-1]
 
+        context["timeline"] = timeline
+        context["next_before"] = (next_before,)
```
longhotsummer, line 71: should this be a tuple like this?

peachjam/models/timeline.py
```diff
@@ -0,0 +1,107 @@
...
+                "subject_documents__work",
+                "subject_documents__labels",
+            )
+            .annotate(event_date=TruncDate("created_at"))
+            .order_by("-event_date", "user_following__id", "-created_at")
+        )
+        # step 3: group into {date: {followed_obj: [docs...]}}
+        grouped = defaultdict(lambda: defaultdict(list))
+
+        for ev in qs:
```
longhotsummer, line 63: We do we need to group the documents together with the user_following? Why can't we use `for doc in event.subject_documents...` ?

peachjam/models/user_following.py
actlikewill: yes, removed

peachjam/models/timeline.py
```diff
@@ -0,0 +1,107 @@
...
+        max_length=256,
+        choices=EventTypes.choices,
+    )
+    created_at = models.DateTimeField(auto_now_add=True)
+    user_notified = models.BooleanField(default=False)
+
+    @classmethod
+    def get_events(cls, user, before=None, limit=1):
+        qs = cls.objects.filter(user_following__user=user).annotate(
+            event_date=TruncDate("created_at")
```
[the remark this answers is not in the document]
actlikewill: okay, we can work it out when we get to it

peachjam/tasks.py
```diff
@@ -250,7 +250,32 @@ def update_user_follows_for_user(user_id):
         return
 
     log.info(f"Updating user follows for user {user_id}")
-    UserFollowing.update_and_alert(user)
+    UserFollowing.update_timeline(user)
+
+
+@background(queue="peachjam", remove_existing_tasks=True, schedule={"priority": -1})
```
actlikewill, line 256: i've set the email tasks at a lower priority so that they run after the update follow tasks.

peachjam/models/user_following.py
```diff
@@ -201,110 +204,40 @@ def get_new_followed_documents(self):
...
         for follow in follows:
             new = follow.get_new_followed_documents()
-            if new["documents"]:
+            if new:
                 log.info(
-                    f"Found {new['documents'].count()} new documents for {new['followed_object']}"
+                    f"Found {new.count()} new documents for {follow.followed_object}"
                 )
+                today = timezone.now().date()
+                timeline_event = TimelineEvent.objects.filter(
```
longhotsummer, line 220: use objects.get_or_create to avoid race conditions.

peachjam/tasks.py
```diff
@@ -250,7 +250,32 @@ def update_user_follows_for_user(user_id):
...
+def send_timeline_email_alerts():
+    from peachjam.models import TimelineEvent
+
+    log.info("Sending timeline emails")
+    TimelineEvent.send_email_alerts()
+    log.info("Timeline emails sent")
```
longhotsummer, line 262: not really true any more :)

peachjam/management/commands/backfill_timeline.py
```diff
@@ -0,0 +1,71 @@
...
+        total_users = users.count()
+        ten_days_ago = datetime.now() - timedelta(days=10)
+        self.stdout.write(f"Found {total_users} users with follows")
+
+        for user in users:
+            self.stdout.write(f"Processing user {user.username}...")
+            with transaction.atomic():
+
+                TimelineEvent.objects.filter(user_following__user=user).delete()
+                for follow in user.following.all():
```
longhotsummer, line 27: If you refactor UserFollowing a bit you can remove all this duplicate code and do something like this:

```
for follow in user.following.all():
  follow.last_alerted_at = ten_days_ago
  documents = follow.get_new_followed_documents
  follow.create_timeline_entries(documents)
```

That will help reduce bugs in duplicated code.

peachjam/models/user_following.py
```diff
@@ -195,116 +198,53 @@ def get_documents_queryset(self):
...
+            event = TimelineEvent.objects.create(
+                user_following=self,
+                event_type=TimelineEvent.EventTypes.NEW_DOCUMENTS,
+            )
+        else:
+            log.info(f"Updating existing timeline event for {self.followed_object}")
+
+        event.subject_documents.add(*documents)
+        self.last_alerted_at = timezone.now()
+        self.save(update_fields=["last_alerted_at"])
```
longhotsummer, line 232: this isn't a good place to do this -- it's not related to creating the timeline events. Whatever method is responsible for triggering this update should, once the event is created, then update the `last_alerted_at`.

peachjam/management/commands/backfill_timeline.py
```diff
@@ -0,0 +1,62 @@
...
+                    self.stdout.write(
+                        f"Processing follow for {follow.followed_object}..."
+                    )
+
+                    # Get the documents followed by this user
+                    docs = (
```
longhotsummer, line 33: minor: you could still call `follow.get_new_followed_documents(ten_days_ago)` if you make the `limit` parameter optional (or do it outside that method, which would also make sense)

peachjam/apps.py
```diff
@@ -38,3 +42,4 @@ def ready(self):
             ).replace(hour=3, minute=0, second=0)
             rank_works(schedule=run_at, repeat=Task.WEEKLY)
             update_user_follows(schedule=Task.HOURLY, repeat=Task.DAILY)
+            send_timeline_email_alerts(schedule=Task.HOURLY, repeat=Task.DAILY)
```
longhotsummer, line 45: the irony here is that because we deploy multiple times a day sometimes, we land up sending multiple emails in a day, rather than just one.

we don't need to fix it now, but we probably want to fix it so it either sends daily exactly, or safely sends multiple times a day (eg. twice).

peachjam/models/user_following.py
```diff
@@ -195,116 +198,53 @@ def get_documents_queryset(self):
...
-        follows = UserFollowing.objects.filter(user=user)
+    def update_timeline(cls, user):
+        follows = cls.objects.filter(user=user)
         log.info(f"Found {follows.count()} follows for user {user.pk}")
-        followed_documents = []
         for follow in follows:
-            new = follow.get_new_followed_documents()
-            if new["documents"]:
+            documents = follow.get_new_followed_documents()
+            event = follow.create_timeline_event(documents)
```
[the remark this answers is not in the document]
actlikewill: done

peachjam/models/user_following.py
actlikewill: done

longhotsummer (approved):
looks great

Issue #2399, opened 2025-04-03
Notifications for saved documents


niiroobiro:
When a user saves a document we can treat it as "following" that document, rather than creating a separate concept. We can then notify the user when something interesting happens related to that document.

For each type of document, we must define what "interesting" can mean.

longhotsummer:
# Proposal

A user can also follow a single document. The user is alerted as follows:

## Legislation

1. the document is repealed, commenced or amended (we'll need to see how we do this since the repeal or commencement date may be before the last time we sent an alert)
2. there is a new child (subsidiary) document

## Judgment

1. a case history entry is created with this judgment as the historical judgment (ie. this judgment was reviewed by a superior court)

## User experience

The UI will need to be updated to show a "follow" button for the document.

We probably want to automatically follow any saved document (and unfollow when unsaved).

actlikewill:
@longhotsummer Sounds good. We can prioritize implementation.

longhotsummer:
This needs a full write up for the UI and to determine how to trigger the notifications mentioned above.

longhotsummer:
We should be able to fold multiple document notifications into one email, such as:

> Hello X, 
>
> There are updates for some of your saved documents.
>
> **Act X of 2005**
> * Repealed by XXX on D
> * Amended by XXX on D
> * [Partially] Commenced by XXX on D
> * New subsidiary legislation: XXX
>
> **A v B Judgment**
> * Cited by XXX on D
> * Mentioned by a superior court in XXX on D: upheld

For all of these, there is a related document that has been created that caused the relationship. The email is triggered when the creation date for the related document is after the last alerted date for the saved document.

actlikewill:
@longhotsummer Here is a proposal for the backend for this:

- Introduce a DocumentChangeLog. Here we track changes worth notifying users about. 
- We could use pre_save and post_save signals to track changes we are interested in or a simple approach could do it in the model save method.
- Send notifications by comparing last notified date for the SavedDocument model and the DocumentChangeLog created_at date


```python
class DocumentChangeLog 
    document = (fk) to CoreDocument
    created_at = (date) when this log notification is created
    old_values = (JSON) here we can create an object tracks the models current values in pre_save signal
    new_values = (JSON) here we save the new fields from post_save signal
    change_type = (choice field) amendment added, case history added etc,

    def get_change_type():
           # here we use a differ to check the changes and determine the change_type
           return ""


class CoreDocument:

       def get_changelog_values():
[...]

PR #2635, opened 2025-08-21
Saved search timeline

actlikewill:
image

part of https://github.com/laws-africa/peachjam/issues/2399


actlikewill:
TODO: create a userfollowing object when saved search is created, 
- what is the best way to deal with the potential circular imports (do it in a signal?)

actlikewill:
@longhotsummer i'm a bit worried because we have peachjam_search models being imported into base peachjam.

peachjam/models/user_following.py
```diff
@@ -132,11 +141,23 @@ class Meta:
...
+            ),
         ]
 
     def __str__(self):
         return f"{self.user} follows {self.followed_object}"
 
+    @property
+    def description_text(self):
+        if self.get_event_type() == TimelineEvent.EventTypes.SAVED_SEARCH:
+            return _("New matches for saved search")
```
longhotsummer, line 157:
```suggestion
            return _("New matches for search alert")
```

peachjam/models/user_following.py
```diff
@@ -160,13 +191,13 @@ def clean(self):
         if set_fields == 0:
             raise ValidationError(
                 "One of the following fields must be set: court, author, court class, court registry, country, "
-                "locality, taxonomy topic"
+                "locality, taxonomy topic saved search"
```
longhotsummer, line 194: you're violating DRY -- use self.follow_fields

longhotsummer (changes requested):
Looks great, minor changes only.

peachjam/models/user_following.py
actlikewill: done

longhotsummer (approved):

PR #2637, opened 2025-08-22
enforce soffice timeout and memory limits

longhotsummer:
otherwise sometimes soffice eats up all memory and never quits


longhotsummer:
today I learned about `prlimit`, thanks chatgpt.

PR #2641, opened 2025-08-27
prevent lxml 6.0.0

longhotsummer:


longhotsummer:
Hopefully this fixes https://lawsafrica.sentry.io/issues/6835162611/

> All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters

PR #2644, opened 2025-08-28
show only latest expression in suggested documents

### LINKED ISSUE #2542: show only latest expression in suggested documents (folder view)
We need to show only latest expressions for suggested documents. Double check also on the document detail view
Image

Sandravaphilips:
Closes #2542 

@actlikewill I'd have had to repeat the Coredocuments mapping for the folder view, so I chose to filter from the base function.

The before and after

Screenshot 2025-08-27 at 17 51 56

Screenshot 2025-08-27 at 17 51 24

### DIFF
--- a/peachjam_ml/models.py
+++ b/peachjam_ml/models.py
@@ -190,11 +190,13 @@ class DocumentEmbedding(models.Model):
         weight_authority = 0.1
         top_k = 100
         avg_embedding = cls.get_average_embedding(doc_ids)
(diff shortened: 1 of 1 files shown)


actlikewill (changes requested):
on second thought, we want to also remove them from the similar documents calculation, so we only always use embeddings for documents that are the most recent

actlikewill (approved):

PR #2661, opened 2025-09-05
Fix search email

actlikewill:
image


actlikewill:
@longhotsummer this should now work properly.

PR #2668, opened 2025-09-10
enforce folder limits

actlikewill:


longhotsummer:
Looks good. The bootstrap blue is pretty lurid. Perhaps `alert-warning` instead?

PR #2679, opened 2025-09-15
Use gunicorn conf to set gunicorn process title

longhotsummer:
* sets process title
* restart worker threads after 10,000 requests to avoid memory leak issues (just in case)

This works correctly:

image


actlikewill (approved):
👍

PR #2706, opened 2025-09-23
default product management command

actlikewill:


longhotsummer (approved):
Just be careful not to run this on tanzlii or lawlibrary

PR #2707, opened 2025-09-23
Support for better caching

longhotsummer:
* move to a public cache for cacheable pages
* don't Vary on cookie
* load dynamic content islands with htmx
* move messages to be loaded dynamically through `/user/loaded` endpoint, which runs on each pageload and also includes user information and the user account menu
* use i18n_patterns for multi-language sites
* move permission checks out of most pages


longhotsummer:
Moved search changes into https://github.com/laws-africa/peachjam/pull/2717

PR #2715, opened 2025-09-25
minor changes to mooc

Sandravaphilips:
Screenshot 2025-09-25 at 14 52 12


longhotsummer:
@Sandravaphilips if you put the `.bg-light` sections into their own `.container`, then the colour stretches across the whole page, which looks nicer. So close the top-level `.container` and then open a new one. That's a common pattern on pages where we need full-width`.bg-light`

africanlii/templates/africanlii/mooc.html
```diff
@@ -54,52 +54,56 @@ <h2 class="fw-bold">🚀 Why Enrol?</h2>
...
+                <p class="small text-muted">{% trans 'Estimated 6–8 weeks (self-paced).' %}</p>
               </div>
             </div>
           </div>
         </div>
       </div>
     </section>
-    <section id="modules" class="py-5 bg-light">
+  </section>
+  <section class="container-fluid bg-light">
```
longhotsummer, line 86:
```suggestion
  <section class="container bg-light">
```

longhotsummer: otherwise this part could be "wider" than the parts above and below.

PR #2717, opened 2025-09-29
no need to vary search on cookie

longhotsummer:
split authed vs non-authed search endpoints; all users get semantic search if enabled


longhotsummer:
This was actually necessary and I've merged it into #2707

PR #2730, opened 2025-10-01
Improve confirmation email text to include a clear call-to-action

Sandravaphilips:
Closes #2654 
Screenshot 2025-10-01 at 18 39 34

PR #2732, opened 2025-10-02
Updated emails

longhotsummer:


longhotsummer:
Thanks, I've folded it into https://github.com/laws-africa/peachjam/pull/2730

PR #2779, opened 2025-10-28
content signals

longhotsummer:
see https://contentsignals.org/


actlikewill (approved):
interesting. lets hope this is adopted more widely

PR #2791, opened 2025-11-05
use work for saved documents

actlikewill:


longhotsummer (dismissed):
looks great, just a small comment on the migration

longhotsummer (approved):

PR #2815, opened 2025-11-18
timeline subject works

actlikewill:
This replaces subject documents with subject works.

### DIFF
--- a/peachjam/tests/test_timeline.py
+++ b/peachjam/tests/test_timeline.py
@@ -55,7 +55,7 @@ class TimelineViewTest(TestCase):
         )
         subject_docs = TimelineEvent.objects.filter(
             user_following__user=self.user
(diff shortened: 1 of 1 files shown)


peachjam/models/timeline.py
```diff
@@ -3,26 +3,66 @@
...
 from django.db.models.functions import TruncDate
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
+from peachjam.models import CoreDocument
+
 log = logging.getLogger(__name__)
 
 
+class TimelineEventManager(models.Manager):
```
actlikewill: @longhotsummer i went back and forth with the best way to attach subject documents to event instances and ended up with this pattern. Its trickier than it seemed because we need to jump across multiple tables. This is the simplest and most efficent, and reusable way to get this done, that i could find

PR #2842, opened 2025-11-26
new amendment alerts

actlikewill:
image

image

### DIFF
--- a/peachjam/models/timeline.py
+++ b/peachjam/models/timeline.py
@@ -261,7 +261,6 @@ class TimelineEvent(models.Model):
             TimelineEvent.objects.attach_subject_documents(ev) for ev in events_qs
         ]
 
-        # FIX: group by (date → following → event_type)
         grouped = defaultdict(lambda: defaultdict(lambda: defaultdict(list)))
 
         for ev in events_qs:


peachjam/timeline_email_service.py
```diff
@@ -22,12 +24,20 @@
 class TimelineEmailService:
     @staticmethod
     def already_alerted_today(user, event_type):
+        q = Q()
         last_24_hrs = timezone.now() - timedelta(hours=24)
-        email_sent = TimelineEvent.objects.filter(
+
+        if type(event_type) == list:
+            q = Q(event_type__in=event_type)
+        elif type(event_type) == TimelineEvent.EventTypes:
```
longhotsummer:
```suggestion
        else:
```

Otherwise, if you have a typo somewhere and you're calling this method with a weird value for `event_type`, then you'll never apply an `event_type` filter to `q` which would be dangerous.

peachjam/timeline_email_service.py
```diff
@@ -231,3 +253,61 @@ def send_new_citation_email(user):
 
         for ev in events:
             ev.mark_as_sent()
+
+    @staticmethod
+    def send_new_relationship_email(user):
```
longhotsummer: as discussed, I think we need to separate out legislation relationship emails from judgment (overturn) relationship emails.

actlikewill: done

longhotsummer (dismissed):
looks good.

PR #2848, opened 2025-11-28
fix citation email urls

actlikewill:


peachjam/timeline_email_service.py
```diff
@@ -155,6 +156,8 @@ def send_new_citation_email(user):
                     "citing_documents": citing_documents,
                 }
             )
+        site = Site.objects.get_current()
+        context["site_domain"] = f"https://{site.domain}"
```
longhotsummer: i'm sure we already inject this somewhere...

longhotsummer: oh never mind

PR #2852, opened 2025-12-02
All AI-generated fields should be editable in the admin

### LINKED ISSUE #2712: All AI-generated fields should be editable in the admin

Sandravaphilips:
Closes #2712 
I couldn't find any other fields that haven't been included, except held and issues.

I'm also not sure about leaving the fields as a list as is. But it seems creating a custom widget might be more trouble than it's worth. If you have better ideas, please let me know.

Screenshot 2025-12-02 at 08 06 20

### DIFF
--- a/peachjam/admin.py
+++ b/peachjam/admin.py
@@ -1140,12 +1140,12 @@ class JudgmentAdminForm(DocumentForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-
-        if self.instance and self.instance.held:
-            self.initial["held"] = self.convert_to_paragraphs(self.instance.held)
-
-        if self.instance and self.instance.issues:
-            self.initial["issues"] = self.convert_to_paragraphs(self.instance.issues)
+        self.initial["held"] = self.convert_to_paragraphs(
+            self.instance.held if self.instance else []
+        )
+        self.initial["issues"] = self.convert_to_paragraphs(
+            self.instance.issues if self.instance else []
+        )
 
     def clean_held(self):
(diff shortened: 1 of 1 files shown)


longhotsummer:
Hmm. Both `held` and `issues` are lists of strings. What if you use a textarea widget, and then use `splitlines()` to split them into lists when saving?

I actually think in the long term we can get rid of them and fold them into `case_summary`.

Sandravaphilips:
It looks like this now
Screenshot 2025-12-03 at 09 34 52

peachjam/admin.py
```diff
@@ -1121,11 +1121,28 @@ def get_formset(self, request, obj=None, **kwargs):
 
 class JudgmentAdminForm(DocumentForm):
     hearing_date = forms.DateField(widget=DateSelectorWidget(), required=False)
+    held = forms.CharField(widget=CKEditorWidget(), required=False)
```
longhotsummer: These are rich HTML fields, they're just plain text. So no need for CKEditor widget.

peachjam/admin.py
```diff
@@ -1138,6 +1155,15 @@ def save(self, *args, **kwargs):
 
         return self.instance
 
+    def convert_to_paragraphs(self, value):
+        if not value:
+            return ""
+
+        if isinstance(value, list):
+            return "".join([f"{item}\n" for item in value])
```
longhotsummer:
```suggestion
            return "\n".join(value)
```

i think that's sufficient?

longhotsummer (approved):

PR #2858, opened 2025-12-03
Updates the README for all required dependencies

mugiwaramunyi:
This PR fixes the issue, when following the instructions in the README.md file, we have some hidden dependencies that cause the manage.py commands to fail.

##WorkItem
<https://github.com/orgs/laws-africa/projects/12/views/1>


README.md
```diff
@@ -6,7 +6,9 @@
 Project Peach Jam
 
 ## Prerequisites
-- postgresql
+- Python@3.10
+- libmagic
```
longhotsummer: I don't have libmagic installed and we don't install it in our Dockerfile -- perhaps a later python needed it but 3.10 doesn't?

README.md
```diff
@@ -6,7 +6,9 @@
...
+- postgresql > 14
 - pip
 - elasticsearch
```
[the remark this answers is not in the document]
Jonathan-Ngondi: Okay noted

README.md
Jonathan-Ngondi: @actlikewill I left that part out because one of the migrations takes care of this step
`peachjam_ml/migrations/0001_pgvector.py`. Should I still include it in the README?

actlikewill: @Jonathan-Ngondi We could still make a note of it in the docs and specifically mention that the database needs to have the pgvector plugin installed

PR #2866, opened 2025-12-04
add check for email alerts within last 24 hours

actlikewill:
- adds a check for emails sent within the last 24 hours
- removes old timeline events

### DIFF
--- a/peachjam/tasks.py
+++ b/peachjam/tasks.py
@@ -258,8 +258,6 @@ def send_timeline_email_alerts():
     from peachjam.timeline_email_service import TimelineEmailService
 
     log.info("Checking for pending timeline emails")
-    log.info("Temporarily disabled")
-    return
     TimelineEmailService.send_email_alerts()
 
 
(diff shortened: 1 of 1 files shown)


peachjam/timeline_email_service.py
```diff
@@ -12,10 +16,34 @@
...
+            )
+            return True
+        return False
+
     @staticmethod
     def send_email_alerts():
+        if not settings.PEACHJAM["EMAIL_ALERTS_ENABLED"]:
+            log.info("email alerts are disabled")
+            return
+
```
actlikewill: @longhotsummer i've added the setting for disabling email alerts. I considered adding a setting in the Settings model, to have more control but we can add that if or when needed

peachjam/models/save_document.py
```diff
@@ -105,6 +105,9 @@ class SavedDocument(models.Model):
     def __str__(self):
         return self.work.title
 
+    def get_absolute_url(self):
+        return self.work.frbr_uri
+
```
actlikewill: @longhotsummer this fixes the citation link in the timeline

longhotsummer: This must use reverse (like `document.get_absolute_url`) otherwise the language prefix on multilingual sites won't be included

longhotsummer (changes requested):
just work.get_absolute_url pls

longhotsummer (approved):

PR #2882, opened 2025-12-10
mark timeline events as read

actlikewill:
- this fixes marking timeline events as sent, for liis that do not have email sending enabled

### DIFF
--- a/peachjam/timeline_email_service.py
+++ b/peachjam/timeline_email_service.py
@@ -40,10 +40,6 @@ class TimelineEmailService:
 
     @staticmethod
     def send_email_alerts():
-        if not settings.PEACHJAM["EMAIL_ALERTS_ENABLED"]:
-            log.info("email alerts are disabled")
-            return
-
         events = TimelineEvent.objects.filter(email_alert_sent_at__isnull=True)
         if not events.exists():
             return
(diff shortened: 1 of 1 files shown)


peachjam/timeline_email_service.py
```diff
@@ -180,58 +175,59 @@ def send_new_citation_email(user):
...
+                        {
+                            "document": doc,
+                            "provision_citations": provision_citations,
+                        }
+                    )
+
+                context["saved_documents"].append(
                     {
-                        "document": doc,
-                        "provision_citations": provision_citations,
```
longhotsummer: should this be changing?

peachjam/timeline_email_service.py
```diff
@@ -180,58 +175,59 @@ def send_new_citation_email(user):
...
+            with override(user.userprofile.preferred_language.pk):
+                send_templated_mail(
+                    template_name="new_citation_alert",
+                    from_email=settings.DEFAULT_FROM_EMAIL,
+                    recipient_list=[user.email],
+                    context=context,
+                )
+
+        for ev in events:
+            ev.mark_as_sent()
```
Jonathan-Ngondi: Kind of a Nit: But in the event of an email service failure/crash halfway through doing it's thing, the email maybe sent, but it won't be marked. So duplicate emails will still occur. 
Maybe using guaranteed atomicity via transaction.atomic() for future feature improvements?

longhotsummer: `send_templated_email` isn't asynchronous. Once that API succeeds, the code will continue and call `mark_as_sent()`. If the API call fails, it will raise an exception and so `mark_as_sent()` won't be reached. 

If the email API returns success but can't (for some reason) send the email, then that's out of our hands.

PR #2896, opened 2026-01-12
Adds the Journal class, Volume class, and adds Journal Foreign Key to Journal Art…

mugiwaramunyi:
…icle


peachjam/models/journals_books.py
```diff
@@ -31,15 +33,68 @@ def pre_save(self):
...
+    entity_profile = GenericRelation(
+        "peachjam.EntityProfile", verbose_name=_("profile")
+    )
+
+    # "Editorial board" - Assuming text for now, but could be ManyToMany to Users
+    editorial_board = models.TextField(
+        help_text="Markdown or plain text listing board members"
+    )
+
+    peer_reviewed = models.BooleanField(default=True)
```
longhotsummer: I suggest both of these are simply captured in the entity profile.

peachjam/models/journals_books.py
```diff
@@ -31,15 +34,81 @@ def pre_save(self):
...
+
+
 class JournalArticle(CoreDocument):
 
     decorator = JournalArticleDecorator()
 
     publisher = models.CharField(max_length=2048)
     default_nature = ("journal_article", "Journal article")
+    journal = models.ForeignKey(
+        "Journal",
```
actlikewill: does this work without the peachjam appname prefix? maybe it would just be better to be explicit

Jonathan-Ngondi:
> @actlikewill Yes I was worried about this because my migration history looked messy compared to what I see in the codebase. Thanks for your help!

peachjam/models/journals_books.py
```diff
@@ -31,15 +35,146 @@ def pre_save(self):
...
+        db_index=True,  # Indexing added since this is key for sorting/faceting
+    )
+
+    class Meta:
+        ordering = ["-year", "title"]
+        verbose_name = "Volume/Issue"
+        verbose_name_plural = "Volumes/Issues"
+
+    def save(self, *args, **kwargs):
+        slug = (self.slug or "").strip().lower()
```
actlikewill: @Jonathan-Ngondi i'm not sure we need this logic to ensure unique slugs. I think its enough to set the field to unique in the database and let it enforce that. if the user adds a volume with a the same title, they can correct it themselves. i also saw some logic further down that checks that the slugs are present,  you could also set the field to be required by removing null=True,

peachjam/views/journals.py
```diff
@@ -0,0 +1,166 @@
...
+    FilteredDocumentListView,
+)
+
+
+def ensure_volume_slugs(volumes):
+    for volume in volumes:
+        slug = (volume.slug or "").strip().lower()
+        if not slug or slug in {"none", "null"}:
+            volume.slug = slugify(volume.title)
+            volume.save(update_fields=["slug"])
```
actlikewill: making slugs required will avoid this

Jonathan-Ngondi:
Screenshot 2026-01-19 at 14 48 08
 URL: `http://127.0.0.1:8000/en/journals/`
 
 
Screenshot 2026-01-19 at 15 36 25
URL: `http://127.0.0.1:8000/en/journals/test-journal/`

Screenshot 2026-01-19 at 15 36 59
URL: `http://127.0.0.1:8000/en/journals/articles/test-journal-article/`

peachjam/urls/journals.py
```diff
@@ -0,0 +1,26 @@
...
+    JournalArticleSlugDetailView,
+    JournalListView,
+    VolumeIssueDetailView,
+    VolumeIssueListView,
+)
+
+urlpatterns = [
+    path("", JournalListView.as_view(), name="journal_list"),
+    path("articles/", JournalArticleListView.as_view(), name="journal_article_list"),
+    path(
```
longhotsummer: This view isn't necessary, your `JournalArticleDetailView` class will be called when the article is viewed using its Expression FRBR URI, just like all the other documents.

That also means that the `slug` field on your JournalArticle class can fall away. The CoreDocument system will ensure a unique expression FRBR URI exists, which effectively includes a "slug" (derived from the title).

PR #2905, opened 2026-01-20
AI summaries and document chat support for SenLII

longhotsummer:
This mimics the same setup for CivLII - https://github.com/laws-africa/peachjam/blob/main/civlii/settings.py


longhotsummer:
I'm deploying because I need this set up so that I can demo it later today.

PR #2949, opened 2026-02-09
Upgrade to python 3.12

longhotsummer:
* update tests and deployments to use python 3.12
* updates pre-commit package versions for 3.12 (except djlint)
* commit results of `pre-commit run --all-files` with updated linters
* change Dockerfile to deploy with Ubuntu 24.04 as base, so we know what versions of libreoffice and poppler-utils we're getting

djlint-django should be updated, but it makes a lot of html file changes which we need to check first, and it's happy to run on python 3.12

## 3rd party apps

* LibreOffice 24.2.7.2 420(Build:2)
* pdftotext version 24.02.0
* pandoc 3.1.3

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,21 +3,17 @@ FROM ubuntu:24.04
 ENV PYTHONUNBUFFERED=1
 ENV DEBIAN_FRONTEND=noninteractive
 
-# python 3.12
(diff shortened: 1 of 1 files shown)


longhotsummer:
We will all need to re-install the pre-commit hooks to bring them up to date, since they run from a separate python venv.

PR #2950, opened 2026-02-09
Use OpenAI for document chats

longhotsummer:
* move from langgraph to OpenAI's agent framework for document chats. This is much simpler to work with and is OpenAI's preferred mechanism for running agents.
* move document chat functionality out of peachjam_ml and into peachjam -- it's not actually related to anything machine-learning related

### DIFF
--- a/peachjam/templatetags/peachjam.py
+++ b/peachjam/templatetags/peachjam.py
@@ -221,11 +221,11 @@ def recent_chats(user):
         .annotate(
             document_rank=Window(
                 expression=RowNumber(),
-                partition_by=[F("document_id")],
+                partition_by=[F("core_document_id")],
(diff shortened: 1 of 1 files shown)


peachjam/admin.py
```diff
@@ -2074,6 +2075,84 @@ class PartnerAdmin(admin.ModelAdmin):
...
+    can_delete = False
+    fields = ("updated_at", "document_link", "score")
+    readonly_fields = fields
+    show_change_link = True
+
+    def get_queryset(self, request):
+        return (
+            super()
+            .get_queryset(request)
+            .select_related("document")
```
Jonathan-Ngondi: Same as above comment `core_document`

PR #2962, opened 2026-02-13
Update favicon files and references in base.html

Dennismwangi-k:
Updated the favicon for Ghalii

before:

Screenshot 2026-02-13 at 3 18 47 PM

After:

Screenshot 2026-02-13 at 3 18 43 PM


Dennismwangi-k:
> Please use the logo that has black bars. That is the real logo, and most users will be using a light background, and the white bars don't show up.

updated

longhotsummer (approved):
great

PR #2967, opened 2026-02-17
Implement passwordless OTP login feature with email verification

Dennismwangi-k:
- Adds passwordless login so users can sign in with just their email and a one-time code
- Provides a password fallback for users who prefer traditional login
- Allows new users to sign up and optionally set a password after verifying their email
- Includes a polished login card UI with smooth toggling between code and password options

### DIFF
--- a/peachjam/urls/accounts.py
+++ b/peachjam/urls/accounts.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.urls import include, path
 from django.views.generic import RedirectView
 
@@ -10,14 +11,10 @@ from peachjam.views import (
 )
 from peachjam.views.accounts import OnboardView, UserAuthView
 
-urlpatterns = [
-    path(
-        "login/code/",
-        RedirectView.as_view(pattern_name="account_login", query_string=True),
-        name="account_request_login_code",
-    ),
+urlpatterns = []
+
+urlpatterns += [
     path("", include("allauth.urls")),
-    path("login/auth", UserAuthView.as_view(), name="account_confirm_login_code"),
     path("accept-terms/", AcceptTermsView.as_view(), name="account_accept_terms"),
     path(
         "onboard",
@@ -40,3 +37,15 @@ urlpatterns = [
         ),
     ),
 ]
+
+# OTP-specific URLs: redirect allauth's "request a code" page to login,
+# and register our custom verification view.
+if settings.PEACHJAM["AUTH_OTP"]:
+    urlpatterns += [
+        path(
+            "login/code/",
+            RedirectView.as_view(pattern_name="account_login", query_string=True),
+            name="account_request_login_code",
+        ),
+        path("login/auth", UserAuthView.as_view(), name="account_confirm_login_code"),
+    ]
--- a/peachjam/context_processors.py
+++ b/peachjam/context_processors.py
@@ -24,6 +24,7 @@ def general(request):
         "SEARCH_SEMANTIC": settings.PEACHJAM["SEARCH_SEMANTIC"],
         "CUSTOMERIO_JS_KEY": settings.PEACHJAM["CUSTOMERIO_JS_KEY"],
         "CUSTOMERIO_JOURNEYS_SITE_ID": settings.PEACHJAM["CUSTOMERIO_JOURNEYS_SITE_ID"],
+        "AUTH_OTP": settings.PEACHJAM["AUTH_OTP"],
         # this object will be injected into Javascript to provide configuration settings to the Javascript app
         "PEACHJAM_JS_CONFIG": {
             "appName": settings.PEACHJAM["APP_NAME"],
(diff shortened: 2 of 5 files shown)


peachjam/views/accounts.py
```diff
@@ -9,13 +14,151 @@
...
+            return self._handle_set_password(request)
+        elif action == "signup_password":
+            return self._handle_signup_password(request)
+        elif action == "verify_otp_for_password":
+            return self._handle_verify_otp_for_password(request)
+
+        return super().post(request, *args, **kwargs)
+
+    def _handle_resend(self, request):
+        email, user = self._get_email_and_user()
```
longhotsummer: guard against email and user both being None -- add a test for this

peachjam/views/accounts.py
```diff
@@ -9,13 +14,151 @@
...
+        elif action == "set_password":
+            return self._handle_set_password(request)
+        elif action == "signup_password":
+            return self._handle_signup_password(request)
+        elif action == "verify_otp_for_password":
+            return self._handle_verify_otp_for_password(request)
+
+        return super().post(request, *args, **kwargs)
+
+    def _handle_resend(self, request):
```
longhotsummer: Can we drop this, and instead get the HTML to submit the form to the usual login page, as though the user has just typed in their email address? Otherwise we're duplicating the logic.

peachjam/views/accounts.py
```diff
@@ -9,13 +12,113 @@
...
+
+        if action == "resend":
+            return self._handle_resend(request)
+        elif action == "password_login":
+            return self._handle_password_login(request)
+        elif action == "signup_password":
+            return self._handle_signup_password(request)
+
+        response = super().post(request, *args, **kwargs)
+        if request.user.is_authenticated and not request.user.first_name:
```
longhotsummer: Handle updating the user's details in a separate GET view. Building it all into one view is too confusing and complex.

Dennismwangi-k: do you want us to render new page or we can just separate the views but still reuse the same templates to avoide rendering new page when filling the first name and last name?

peachjam/views/accounts.py
```diff
@@ -1,21 +1,158 @@
...
+
+            next_url = self._get_success_url(request)
+            profile_url = reverse("complete_profile")
+            if next_url:
+                profile_url += "?" + urlencode({"next": next_url})
+            return HttpResponseRedirect(profile_url)
+        return response
+
+    def _get_success_url(self, request):
```
longhotsummer: I think this falls away.

peachjam/views/accounts.py
```diff
@@ -1,21 +1,158 @@
...
+
+    def get_object(self):
+        return self.request.user
+
+    def get_form(self, form_class=None):
+        form = super().get_form(form_class)
+        form.fields["first_name"].required = True
+        return form
+
+    def _get_next_url(self):
```
longhotsummer: use the `NextRedirectMixin` and call `get_success_url()` and it will do all of this for you.

peachjam/urls/accounts.py
```diff
@@ -7,10 +7,21 @@
     EditAccountView,
     LoggedOutView,
 )
+from peachjam.views.accounts import CompleteProfileView, PeachjamConfirmLoginCodeView
 
 urlpatterns = [
+    path(
+        "login/code/confirm/",
```
longhotsummer: `login/auth`

peachjam/templates/account/user_auth.html
```diff
@@ -0,0 +1,233 @@
...
+                {% endif %}
+              </div>
+              <div class="mb-3">{{ signup_form.captcha }}</div>
+              <div class="d-grid mb-3">
+                <button type="button" class="btn btn-primary py-2 fs-5" id="pw-continue-btn">{% trans 'Continue' %}</button>
+              </div>
+            </div>
+            <div id="pw-step2" style="display: none;">
+              <div class="mb-3">
+                <label class="form-label" for="id_pw_first_name">{% trans 'First name' %}</label>
```
longhotsummer: Entering the name is for the CompleteProfile view only, not here. I've seen too many users enter passwords that don't match or aren't valid, and they have to start again, and having this extra step makes it worse.

peachjam/forms.py
```diff
@@ -643,6 +644,19 @@ class PeachjamLoginForm(LoginForm):
     captcha = get_recaptcha_field()
 
 
+class PasswordSignupForm(PeachjamSignupForm):
+    password1 = PasswordField(label=_("Password"), autocomplete="new-password")
+    password2 = PasswordField(label=_("Password (again)"), autocomplete="new-password")
+
+    def clean(self):
+        cleaned = super().clean()
+        if not cleaned.get("first_name"):
```
longhotsummer: Leave name entry entirely for the CompleteProfile view.

Dennismwangi-k:
> I think this is fine to merge, because it is not turned on.

thank you, I’ll proceed with merging.

PR #2968, opened 2026-02-18
UI tweaks for ADR Hub

mugiwaramunyi:


peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -300,15 +300,15 @@ <h1 class="doc-title d-flex">
...
-                            {% trans 'Judgment date' %}
-                          </dt>
-                        {% else %}
-                          <dt>
-                            {% trans 'Date' %}
-                          </dt>
-                        {% endif %}
+                        <dt>
+                          {% block document-metadata-date-label %}
+                            {% if document.doc_type == "judgment" %}
```
longhotsummer: you can now move this override into `judgment_detail.html` rather than changing it here. So we have a standard mechanism to override this label, which now both judgments and arbitration awards do in the same way -- in their templates.

Jonathan-Ngondi: Good catch, I’ll keep leaning on extension and shared interfaces first, instead of introducing custom logic. Will spend more time learning the best approaches to such problems.

longhotsummer (approved):
great

PR #2971, opened 2026-02-18
criminal justice analysis data models

actlikewill:
This adds models for the criminal justice dashboards.

### DIFF
--- a/peachjam/signals.py
+++ b/peachjam/signals.py
@@ -1,3 +1,5 @@
+import logging
+
 import allauth.account.signals as allauth_signals
 from asgiref.sync import async_to_sync
 from django.contrib.auth import get_user_model
@@ -25,6 +27,7 @@ from peachjam.models import (
     UserFollowing,
     UserProfile,
     Work,
+    pj_settings,
 )
 from peachjam.models.lifecycle import after_attribute_changed
 from peachjam.tasks import (
@@ -36,6 +39,8 @@ from peachjam.tasks import (
 from peachjam_search.models import SavedSearch
 
 User = get_user_model()
+
+log = logging.getLogger(__name__)
 
 
 # a user has requested a password reset
@@ -230,6 +235,10 @@ def judgment_content_changed_generate_summary(sender, instance, **kwargs):
 
 @receiver(signals.pre_save, sender=DocumentContent)
 def judgment_content_changed_extract_criminal_data(sender, instance, **kwargs):
+    if not pj_settings().allow_criminal_data_extraction:
+        log.info("Criminal data extraction is disabled.")
+        return
+
     if instance.document.doc_type != "judgment":
         return
 
--- a/peachjam/analysis/criminal_data/extractor.py
+++ b/peachjam/analysis/criminal_data/extractor.py
@@ -3,7 +3,7 @@ import os
 
 from django.db import transaction
 
-from peachjam.models import Judgment, JudgmentOffence, Offence, Sentence, pj_settings
+from peachjam.models import Judgment, JudgmentOffence, Offence, Sentence
 
 from .agent import extract_case_type_filing_year, extract_offences_and_sentences
 
@@ -16,9 +16,6 @@ class CriminalDataExtractor:
     """
 
     def __init__(self):
-        if not pj_settings().allow_criminal_data_extraction:
-            raise ValueError("Criminal data extraction is disabled.")
-
         if not os.environ.get("OPENAI_API_KEY"):
             raise ValueError("OPENAI_API_KEY is not configured.")
 


peachjam/analysis/criminal_data/offence.py
```diff
@@ -0,0 +1,129 @@
...
+from .types import MatchOffenceResult, MentionedOffences
+
+log = logging.getLogger(__name__)
+
+
+class OffenceMentionExtractor(BaseExtractor):
+    def build_prompt(self) -> str:
+        judgment_text = self.judgment.get_content_as_text()
+        return f"""
+            You are a legal information extraction engine.
```
longhotsummer: This is the system or developer prompt. The LLM gives prompts with "role: system" (can only be one, right at the start) or "role: developer" special consideration.

Then provide the document text as "role: user". That helps to separate instructions from content.

longhotsummer: See https://github.com/laws-africa/peachjam/blob/0368f36ba3e6a18ed23a0d26ab639e5f8fd8bc8b/peachjam/chat/agent.py#L123

peachjam/analysis/criminal_data/base.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        Call the LLM service
+        """
+        prompt = self.build_prompt()
+
+        response = self.client.responses.parse(
+            model=self.model,
+            input=[
+                {"role": "system", "content": self.system_prompt()},
+                {"role": "user", "content": prompt},
```
longhotsummer: I've found it useful to be able to pass multiple messages as input, using the different roles available. 

Use `instructions=self.system_prompt()` then `input=self.build_prompt()` which will allow `build_prompt()` to return a list of messages, rather than one, which is useful.

peachjam/analysis/criminal_data/offence_tool.py
```diff
@@ -0,0 +1,193 @@
...
+    def tools(self):
+        return [SEARCH_OFFENCES_TOOL]
+
+    def tool_handlers(self):
+        return {"search_offences": search_offences}
+
+    def build_prompt(self) -> str:
+        judgment_text = self.judgment.get_content_as_text()
+        return f"""
+            <ROLE>
```
longhotsummer: make this your system prompt

peachjam/analysis/criminal_data/agent.py
```diff
@@ -0,0 +1,252 @@
...
+log = logging.getLogger(__name__)
+
+
+KeywordInput = Union[str, Sequence[str]]
+_SPLIT = re.compile(r"[,\n;|]+")
+
+
+@function_tool
+def search_offences(
+    keywords: KeywordInput,
```
longhotsummer: I suggest to avoid being ambiguous: either make this a single string that is comma-separated, or a list of strings, not both. Don't make the LLM make a choice it doesn't have to.

peachjam/tests/test_criminal_data_extraction.py
```diff
@@ -0,0 +1,85 @@
...
+            title="Criminal trespass",
+            description="""
+            A person who unlawfully enters into or upon property in the possession of another with "
+            intent to commit an offence or to intimidate, insult or annoy any person in possession of
+            the property, or who, having lawfully entered, unlawfully remains there with such intent,
+            commits criminal trespass.
+            """,
+        )
+
+    def test_extract_robbery_with_violence(self):
```
longhotsummer: great, much simpler!

You need some tests that cover other edge cases for the fields in your pydantic models, such as "minimum sentence applied", or poorly matching tool results, or ambiguous judgment text where other sentences are mentioned that aren't the actual sentence.

peachjam/analysis/criminal_data/agent.py
actlikewill: i've removed the config parameters from function call

peachjam/signals.py
```diff
@@ -226,6 +228,26 @@ def judgment_content_changed_generate_summary(sender, instance, **kwargs):
         generate_judgment_summary(judgment.pk)
 
 
+@receiver(signals.pre_save, sender=DocumentContent)
+def judgment_content_changed_extract_criminal_data(sender, instance, **kwargs):
```
longhotsummer: I think this should check if criminal data extraction is enabled, and do nothing if it is not. Otherwise it's going to queue up background tasks which fail due to the ValueError raised when the extractor is created.

I think having the check here is better than the extractor's constructor raising an exception. I think of this as the settings flag controlling whether the work is queued up to be done, rather than the actual analysis logic itself checking if it's allowed to be run.

PR #2989, opened 2026-02-24
Bug 2974 fix

mugiwaramunyi:
Closes https://github.com/laws-africa/peachjam/issues/2974


Jonathan-Ngondi:
> I wonder how this happened 🤷

Quite a bizarre case, of a ghost user....

PR #2997, opened 2026-02-25
Feat helpscout beacon 2972

mugiwaramunyi:
Implements changes to the HelpScout Beacon so that if it is configured, the Contact Us button in the footer opens the Beacon window, and if it's not, the user is routed to the default mailto: address.


longhotsummer (changes requested):
great. please also remove the existing beacon from lawlibrary's about.html

PR #3004, opened 2026-02-26
Only create a following if it doesn't already exist

### LINKED ISSUE #2975: ValidationError: {'__all__': ['Constraint “unique_user_court” is violated.']}
Sentry Issue: [LII-3X2](https://lawsafrica.sentry.io/issues/7106640381/?referrer=github_integration)

```
ValidationError: {'__all__': ['Constraint “unique_user_court” is violated.']}
(9 additional frame(s) were not displayed)
...
  File "peachjam/views/mixins.py", line 12, in post
    return super().post(request, *args, **kwargs)
  File "peachjam/models/user_following.py", line 235, in save
    self.full_clean()
```

Sandravaphilips:
Closes #2975 

- Reproduced by double-clicking the follow button.
- Added a post method that creates a following only if it doesn't exist. 

P.S.: Double-clicking to delete a following throws a 404 on the second execution, but also doesn't affect the user experience.

### DIFF
--- a/peachjam/views/user_following.py
+++ b/peachjam/views/user_following.py
@@ -1,7 +1,6 @@
 from django import forms
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.http import HttpResponse, HttpResponseRedirect
-from django.shortcuts import redirect
 from django.urls import reverse
 from django.views.generic import CreateView, DeleteView, FormView, ListView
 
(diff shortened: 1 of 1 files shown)


peachjam/views/user_following.py
```diff
@@ -110,6 +111,15 @@ def get_form_kwargs(self):
         kwargs["data"] = self.request.GET or self.request.POST
         return kwargs
 
+    def post(self, *args, **kwargs):
```
longhotsummer: it's annoying that django makes it hard to do this :(

longhotsummer (approved):

PR #3017, opened 2026-03-02
Feat break contenthtml cycle

mugiwaramunyi:
This PR refactors document content storage so CoreDocument no longer stores content_html directly. Instead, HTML is persisted in DocumentContent (source_html + content_html), with CoreDocument exposing compatibility properties and save behaviour.

The goal is to reduce model coupling, avoid content update cycles, and make upload/import/admin flows consistent around a single canonical source (source_html).

  ## Why

  `content_html` on CoreDocument created tight coupling between document metadata and large mutable content blobs, this completed a loop between the content_html object and other processes. We want to break this cycle so that our data ingestion pipeline maintains a directed acyclical graph. 

  ## What changed

  ### Data model + migrations

  - Added fields to DocumentContent:
      - source_html
      - content_html
  - Backfilled existing CoreDocument.content_html into DocumentContent:
      - sets both source_html and content_html where missing
  - Removed CoreDocument.content_html DB field

  Migrations:

  - 0274_add_documentcontent_source_html_and_more
  - 0275_backfill_documentcontent_html
  - 0276_remove_coredocument_content_html

  ### Core document behavior

  - CoreDocument now manages content_html/source_html through properties backed by DocumentContent
  - Added internal caching and save_document_content() to persist pending HTML changes after CoreDocument.save()
  - get_content_as_text() now regenerates text when missing and avoids stale-cache reads
  - DOCX extraction flow now treats source_html as canonical and derives content_html from it

  ### Lifecycle + analysis flow

  - DocumentContent now uses lifecycle hook:
      - BEFORE_SAVE on content_html change triggers citation extraction task
  - SourceFile now owns file-change side effects via lifecycle hooks:
      - clear stale file_as_pdf before save
      - regenerate pdf + re-extract document content after save
  - Removed now-duplicated signal-based SourceFile PDF conversion
  - Removed form-level run_analysis() / duplicate extraction calls
  - Added raw guard in DocumentContent post-save summary signal to avoid fixture-load side effects

  ### Admin / import / API compatibility

  - Admin form now edits source_html (instead of direct content_html)
  - Import resource supports source_html and keeps backward compatibility by accepting content_html input and mapping it to source flow
  - Anonymisation serializer accepts optional content_html and maps it to source_html for compatibility

  ### Fixtures / tests

  - Updated fixtures to use peachjam.documentcontent records instead of coredocument.content_html
  - Updated document test expectation that assumed absence of document_content
  - Fixes the failing tests with this change.

### DIFF
--- a/peachjam/models/judgment.py
+++ b/peachjam/models/judgment.py
@@ -18,6 +18,7 @@ from django_lifecycle import AFTER_SAVE
 
 from peachjam.analysis.summariser import JudgmentSummariser
 from peachjam.decorators import CauseListDecorator, JudgmentDecorator
+from peachjam.helpers import current_year
 from peachjam.models import (
     CoreDocument,
     DocumentContent,
@@ -26,7 +27,6 @@ from peachjam.models import (
     on_attribute_changed,
 )
 from peachjam.tasks import create_anonymised_source_file_pdf, generate_judgment_summary
-from peachjam.helpers import current_year
 
 log = logging.getLogger(__name__)
 


peachjam/models/core_document.py
```diff
@@ -557,6 +561,15 @@ class Meta:
     def __str__(self):
         return f"{self.doc_type} - {self.title}"
 
+    def __init__(self, *args, **kwargs):
+        content_html = kwargs.pop("content_html", UNSET)
```
longhotsummer: this feels helluva clunky. Can't we just read/write to `document.document_content.content_html`?

peachjam/analysis/citations.py
```diff
@@ -19,31 +19,44 @@ class CitationAnalyser:
...
             # markup the source file by extracting the text and assuming we'll render it as a PDF
             return self.extract_citations_from_source_file(document)
 
     def extract_citations_from_html(self, document):
+        doc_content = self.get_document_content(document)
+        if not doc_content or not doc_content.content_html:
+            return False
+
         try:
-            html = document.content_html_tree
```
longhotsummer: Move `content_html_tree` onto `DocumentContent`

peachjam/models/core_document.py
```diff
@@ -1070,20 +1172,87 @@ class Meta:
...
+        self.toc_json = toc_json
+
+    def update_content_text_from_html(self):
+        if self.content_html:
+            root = self.content_html_tree
+            self.content_text = " ".join(root.itertext())
+        else:
+            self.content_text = ""
+
+    def sync_document_html_cache(self):
```
longhotsummer: This is odd to me, I don't see why it's necessary?

peachjam/models/core_document.py
```diff
@@ -535,9 +533,9 @@ class CoreDocument(AttributeHooksMixin, PolymorphicModel):
         default=False,
         help_text=_("Restrict access to this document to selected groups."),
     )
-
-    # for caching parsed html
-    _content_html_tree = None
+    # Legacy compatibility field retained because fixtures and some schemas still
+    # load CoreDocument rows with this column before DocumentContent takes over.
+    content_html_is_akn = models.BooleanField(_("content HTML is AKN"), default=False)
```
mugiwaramunyi: `content_html_is_akn` is not nullable.

So this is back for a reason, because there's no RemoveField migration for this 
- DB table still has a non-null, 
- CoreDocument doesn't have the field
- Django fixtures omit the column 
- Postgres raises NOT NULL violations during setUPClass.

My thoughts were I'd leave this in here for now, and then remove this in the next PR
If you have a better way to go about this, I'm open.

PR #3033, opened 2026-03-06
Made facets in listing views quick-searchable

### LINKED ISSUE #3026: Make facets in listing views quick-searchable
Image

Use a data-component to put a small text box at the top of the list if there are many items (more than 10?). That must live-filter as the user types based on the checkboxes in the page.

Something like this:

Image

bonus points:

* render checked items at the top of the list

Sandravaphilips:
Closes #3026 
- The clear and clear all buttons also clear the search terms
- I chose 8 as the minimum because that's how many I could immediately see on my desktop. Figured anything more than that should be searchable
- The checked items appear first in the results.
- Coding with AI is always the bomb🤩

https://www.loom.com/share/8b9c3bb7b917475c9f8e739c234da6b3

### DIFF
--- a/peachjam/js/components/FilterFacets/SingleFacet.vue
+++ b/peachjam/js/components/FilterFacets/SingleFacet.vue
@@ -206,12 +206,4 @@ export default {
   max-height: 25vh;
   overflow-y: auto;
 }
-
-.facet-search-sticky {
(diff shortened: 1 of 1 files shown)


longhotsummer (approved):
great

PR #3034, opened 2026-03-06
Implement tabbed navigation for cited cases and legislation

Dennismwangi-k:
Implement tabbed navigation for cited cases and legislation within law report volumes and refactor law report detail views
https://www.loom.com/share/61841191330c4c47a7752388595ea8d4


actlikewill:
@Dennismwangi-k this looks good and works okay too which is great. however, we have an established pattern for tabbed listing views that we use for legislation listing view. it would be great if we could replicate that pattern. i think its a bit more extendable. if we could get these views more closely aligned in both functionality and visuals it would be great. they are not that far off as it is.

longhotsummer (changes requested):

Dennismwangi-k:
> @Dennismwangi-k this looks good and works okay too which is great. however, we have an established pattern for tabbed listing views that we use for legislation listing view. it would be great if we could replicate that pattern. i think its a bit more extendable. if we could get these views more closely aligned in both functionality and visuals it would be great. they are not that far off as it is.

  hey @actlikewill  , sure i can try to ease things and make it more cleaner and follow the code acitecture, but i might not follow exactly what you have done on legislation coz on liiweb i have seen you have used variants which honestly this will work when you are using 1 model, mine will be worse coz im using more than w1 model, please confrim over my code which will change variables

peachjam/views/law_reports.py
```diff
@@ -15,58 +26,263 @@ class LawReportListView(ListView):
...
-        context["law_report_volume_groups"] = chunks(context["law_report_volumes"], 3)
-        context["entity_profile"] = self.law_report.entity_profile.first()
-        context["entity_profile_title"] = self.law_report.title
+        context["law_report_volume"] = self.law_report_volume
+        context["active_tab"] = self.active_tab
         return context
 
 
-class LawReportVolumeDetailView(LawReportDetailView):
+class LawReportJudgmentChildrenMixin:
```
Dennismwangi-k: @longhotsummer  i hope im not too much complex on this, instead of repeating myself, let me just inherit this, this is because we have also drop down toggle to show cited judgements as well on the judgement page, so we need to get those judgements as well, if we follow the same logic we might repeat ouselves alot, so intead i got 1 function to get all those and reuse the code on both tabs

peachjam/templates/peachjam/_document_table_row.html
```diff
@@ -4,12 +4,12 @@
   {% if doc_table_toggle %}
     <td class="cell-toggle">
       {% if document.children %}
-        <button class="btn btn-sm bg-none collapsed"
+        <button class="btn btn-sm bg-none{% if not doc_table_children_expanded %} collapsed{% endif %}"
                 type="button"
-                title="{{ PEACHJAM_SETTINGS.subleg_label }}"
+                title="{{ doc_table_toggle_title|default:PEACHJAM_SETTINGS.subleg_label }}"
```
longhotsummer: rather than having a default, ensure it is set in all views that use it

peachjam/views/law_reports.py
Dennismwangi-k: this will fall away

peachjam/templates/peachjam/layouts/document_detail.html
```diff
@@ -38,7 +38,7 @@
             <div>
               <h1 class="doc-title d-flex">
                 <span>{{ document.title }}</span>
-                <span id="saved-document-star--{{ document.pk }}"></span>
+                <span class="saved-document-star--{{ document.pk }}"></span>
```
Dennismwangi-k, line 41: @longhotsummer  here i had to change from id to class since im rendering same documents more than once in the same page

peachjam/templates/peachjam/saved_document/_create.html
```diff
@@ -10,7 +10,8 @@
   <i class="bi bi-star"></i>
   {% trans 'Save' %}
 </button>
-<span hx-swap-oob="true" id="saved-document-star--{{ document.id }}"></span>
```
Dennismwangi-k, line 13: this will avoid breaking if we have more than 1 document rendered on the same page

peachjam/templates/peachjam/layouts/document_detail.html
longhotsummer: Noted. As long as the functionality still works correctly.

peachjam/templates/peachjam/_document_table_row.html
```diff
@@ -18,9 +18,11 @@
   {% endif %}
   {% if document.is_group %}
     <td id="{% if document.group_id %}{{ document.group_id }}{% endif %}"
-        class="cell-group">{{ document.title }}</td>
+        class="cell-group">
+      <div class="{% if document.is_table_child %}ps-lg-3 ms-lg-1{% endif %}">{{ document.title }}</div>
```
longhotsummer: look at how our SCSS files set up padding using a CSS rule on `table.doc-table .doc-table-children .cell-title` -- apply the same mechanism to `table.doc-table .doc-table-children .cell-group` rather than using custom classes here.

peachjam/templates/peachjam/_document_table_row.html
```diff
@@ -44,7 +46,7 @@
           {% include 'peachjam/_document_taxonomies.html' with taxonomies=document.listing_taxonomies %}
         </div>
       {% endif %}
-      <div id="saved-document-table-detail--{{ document.pk }}"></div>
+      {% if not document.is_table_child %}<div id="saved-document-table-detail--{{ document.pk }}"></div>{% endif %}
```
longhotsummer: Why don't we want this?

longhotsummer (approved):
Great :)

PR #3046, opened 2026-03-10
display dedicated restricted access page for taxonomies and standardize restricted content messaging

Dennismwangi-k:
Before:
Screenshot 2026-03-10 at 2 11 32 PM

After:
Screenshot 2026-03-10 at 2 32 32 PM


longhotsummer (approved):
great

PR #3065, opened 2026-03-13
Changes for attribute tracking

longhotsummer:
* ensure track_changes when modifying document content
* simplify and clean up some tests
* remove unnecessary document content stuff on CoreDocument

I've seen that cascading changes made during hooks don't trigger future hooks, which we can hack around now but need to bear in mind.

### DIFF
--- a/peachjam/adapters/indigo.py
+++ b/peachjam/adapters/indigo.py
@@ -375,7 +375,7 @@ class IndigoAdapter(RequestsAdapter):
             defaults={**field_data, **frbr_uri_data},
         )
         doc_content = created_doc.get_or_create_document_content(True)
-        doc_content.source_html = content_html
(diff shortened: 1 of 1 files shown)


peachjam/adapters/indigo.py
```diff
@@ -373,8 +374,8 @@ def update_document(self, url):
             expression_frbr_uri=expression_frbr_uri,
             defaults={**field_data, **frbr_uri_data},
         )
-        doc_content = created_doc.get_or_create_document_content()
-        doc_content.content_html = content_html
+        doc_content = created_doc.get_or_create_document_content(True)
+        doc_content.source_html = content_html
```
mugiwaramunyi: Should I just go ahead and set_source_html here?

PR #3096, opened 2026-03-23
feat: Add filtering and faceting to the subsidiary legislation view

mugiwaramunyi:
Add filtering and faceting to the subsidiary legislation tab

  Summary

  - Subsidiary legislation is now loaded lazily into the legislation detail page via HTMX when the tab is first clicked, instead of being rendered eagerly with the page.
  - A new LegislationSubsidiaryView (backed by FilteredDocumentListView) serves the subsidiary list at /akn/.../subsidiary, providing full filtering, faceting, sorting, and pagination — identical to the standalone legislation list page.
  - The legislation detail page now stores a child_documents_count (a cheap COUNT query) instead of fetching the full child queryset on every page load.  
  
Screenshot 2026-03-23 at 22 03 32

### DIFF
--- a/peachjam/templates/peachjam/document/_legislation_subsidiary.html
+++ b/peachjam/templates/peachjam/document/_legislation_subsidiary.html
@@ -1,3 +1,3 @@
 {% load i18n %}
-<h4>{% trans "Subsidiary legislation" %}</h4>
-{% include "peachjam/_document_table_form.html" %}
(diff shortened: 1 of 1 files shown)


peachjam/templates/peachjam/legislation_subsidiary.html
longhotsummer: this is fundamentally not a full template because it's just a fragment that is included with htmx. So its name should start with an underscore.

PR #3099, opened 2026-03-25
case outcomes extraction

actlikewill:
this adds case outcome extraction for criminal data


peachjam/analysis/criminal_data/agent.py
```diff
@@ -464,6 +467,102 @@ def extract_offences_and_sentences(judgment_text: str) -> JudgmentOffenceExtract
...
+
+# Examples
+
+Example 1
+
+<text>
+The appeal against conviction is dismissed, but the sentence is reduced from ten years to five years.
+</text>
+
+Return the exact matching canonical labels from the provided list only.
```
longhotsummer: Which label would be applied in this case?

PR #3108, opened 2026-03-30
chore: fix the broken subsidiary fragment behaviour on filter/sort

mugiwaramunyi:
Fixes the bug for filtering and sorting on subsidiary legislation
- Adds opt out to hx-push-url in the shared document table form
- Allows filtering and sorting to work without replacing the pages url with the htmx fragments endpoint
- A flag is passed in the LegislationSubsidiaryView to disable hx-push-url

Screenshot 2026-03-31 at 10 37 24

URL: http://127.0.0.1:8000/en/akn/zm/act/1965/67/eng@1996-12-31


longhotsummer:
Screenshot please :)

PR #3115, opened 2026-03-31
Navigation links to the topic page

Dennismwangi-k:
I have added  navigation links to the flynote topic pages . I also fixed a breadcrumb error in decorators.py: some judgments exist without an associated court(im not sure if its my database but you can confirm as well on your side), and breadcrumb generation was assuming document.court was always present, which caused an exception. The fix guards the court- based breadcrumb steps so the judgment page still renders even when court metadata is missing

Screenshot 2026-03-31 at 9 25 19 PM

Screenshot 2026-05-08 at 9 21 22 AM

Screenshot 2026-05-08 at 9 21 10 AM


peachjam/templates/peachjam/_linked_flynotes.html
longhotsummer: please move under templates/peachjam/document/

peachjam/templates/peachjam/home.html
```diff
@@ -3,7 +3,15 @@
 {% block full-title %}{{ APP_NAME }}{% endblock %}
 {% block body-class %}home-template{% endblock %}
 {% block page-content %}
-  <section>
-    Page content
+  <section class="py-5">
```
longhotsummer: not necessary

liiweb/templates/liiweb/home.html
```diff
@@ -34,13 +34,16 @@ <h2>
...
     <div class="container pt-2">
-      <div class="mt-3 d-lg-none row row-cols-2">
+      <div class="mt-3 d-lg-none row row-cols-3">
         <div class="col mb-2">
           <a href="{% url 'judgment_list' %}" class="btn btn-primary d-block">{% trans 'Judgments' %}</a>
         </div>
         <div class="col mb-2">
           <a href="{% url 'legislation_list' %}" class="btn btn-primary d-block">{% trans 'Legislation' %}</a>
         </div>
+        <div class="col mb-2">
```
longhotsummer: not necessary, since the topics are included under the judgments link.

peachjam/templates/peachjam/judgment_detail.html
```diff
@@ -95,17 +95,23 @@
 {% endblock %}
 {% block extra-metadata %}
   {{ block.super }}
+  {% if document.flynote %}
```
longhotsummer: why is this changing?

peachjam/templates/peachjam/judgment_detail.html
```diff
@@ -95,17 +95,23 @@
...
+    <dl class="document-metadata-list d-md-flex">
+      <dt>
+        {% trans 'Flynote' %}
+      </dt>
+      <dd>
+        {% if document.blurb %}<div class="mb-3">{{ document.blurb|default_if_none:"" }}</div>{% endif %}
+        {% include "peachjam/_linked_flynotes.html" with linked_flynotes=linked_flynotes flynote_text=document.flynote %}
```
longhotsummer: simplest to pass in the document object directly, and make linked_flynotes a helper method on the judgment class.

peachjam/models/judgment.py
```diff
@@ -492,6 +493,38 @@ def case_duration(self):
...
+            flynote = judgment_flynote.flynote
+            nodes = [*flynote.get_ancestors(), flynote]
+            linked_nodes_by_path[tuple(node.name for node in nodes)] = nodes
+
+        linked_lines = []
+        text = parser.clean(self.flynote)
+        if not text:
+            return linked_lines
+
+        for line in parser.normalise_multiline_text(text).splitlines():
```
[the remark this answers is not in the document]
longhotsummer: this is no longer necessary, we can rely on the Flynote objects (if they exist).

longhotsummer:
@Dennismwangi-k pls ensure the flynote topics (and the links to the Topics page) are only rendered if the `PEACHJAM["SUMMARISE_USE_FLYNOTE_TREE"]` setting is `True`.

peachjam/templates/peachjam/document/_linked_flynotes.html
```diff
@@ -1,14 +1,11 @@
 {% if linked_flynotes %}
   <div class="mb-2 fst-italic">
-    {% for line in linked_flynotes %}
-      {% for item in line %}
-        {% for node in item.nodes %}
-          <a href="{{ node.get_absolute_url }}">{{ node.name }}</a>
-          {% if not forloop.last %}—{% endif %}
-        {% endfor %}
-        {% if not forloop.last %};{% endif %}
+    {% for item in linked_flynotes %}
```
longhotsummer: Please make each flynote a `li`, and the whole thing inside `<ul class="list-unstyled">`

peachjam/models/judgment.py
```diff
@@ -599,6 +621,11 @@ def pre_save(self):
             self.assign_mnc()
             self.assign_title()
 
+        # flynote_raw is the editable/source field; keep flynote populated immediately
```
longhotsummer: Why is this necessary?

Dennismwangi-k: coz if there is a flynote raw and the hook to sync it didnt run the flyning note will be empty and since its the one being displayed it will not display hence breaking and thats why my tests were failing

longhotsummer: `flynote` is updated when `flynote_raw` changes by the `BEFORE_SAVE` hook, but the object must have `track_changes()` called to ensure that happens. Forcing them to be the same here in `pre_save()` is technically allowed, but it means that your tests aren't actually correct, because they aren't calling `track_changes()` and so the hook is not being run.

So the best fix here is to ensure that your tests only set `flynote_raw` after calling `track_changes()`.

Dennismwangi-k: ok let me remove them

Dennismwangi-k: no need to force them if track_changes() makes sure they are always sync

peachjam/settings.py
```diff
@@ -189,6 +189,8 @@ def build_allowed_hosts(*default_hosts):
     "SUMMARISE_JUDGMENTS": True,
     # should we use the flynote tree and match flynotes to our database of known flynotes?
     "SUMMARISE_USE_FLYNOTE_TREE": False,
+    # should flynote topic navigation and linked flynote UI be shown when the flynote tree is enabled?
+    "SHOW_FLYNOTE_TOPICS": False,
```
Dennismwangi-k, line 193: @longhotsummer  this will only control 2 things, how the flying note will be rendered on the detail page like if its on it will show the links, else it will fall back to plain text also the "" Explore case law by topic "". will be hidden if its false. hope thats only where we should apply that?

PEACHJAM["SHOW_FLYNOTE_TOPICS"] = True

Dennismwangi-k: i kinda love it like that coz when we make it premium  we can turn it on for few months and make them clickable then later we turn off they see plain text and will see the benefits of paying it coz now they can navigate to the topics they want, what do you think??

longhotsummer (approved):

PR #3130, opened 2026-04-08
fixed the null issue on judgement field coart

Dennismwangi-k:
fixed the null issue on judgement field coart to avoide having judgements without coarts, i also added some restrains on the object to avoid RelatedObjectDoesNotExist and instead just get None

### DIFF
--- a/peachjam/models/judgment.py
+++ b/peachjam/models/judgment.py
@@ -584,8 +584,6 @@ class Judgment(CoreDocument):
             court = self.court
             if court.country:
                 self.jurisdiction = court.country
-
-            if court.country:
                 self.locality = court.locality
 
         self.doc_type = "judgment"
--- a/peachjam/models/judgment.py
+++ b/peachjam/models/judgment.py
@@ -313,7 +313,7 @@ class Judgment(CoreDocument):
         CIVIL = "civil", _("Civil")
 
     court = models.ForeignKey(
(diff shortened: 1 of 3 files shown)


actlikewill:
@Dennismwangi-k django usually asks what you want to do with existing null values, which should be handled in by setting a value that will populate it. (you could set it to pk=1 maybe). It probably won't happen in prod but you should test it just in case. add a judgment without a court then run the migration and see what happens

peachjam/models/judgment.py
```diff
@@ -580,12 +580,13 @@ def pre_save(self):
...
-            if self.court.country:
-                self.jurisdiction = self.court.country
+        if self.court_id:
+            court = self.court
+            if court.country:
+                self.jurisdiction = court.country
 
-            if self.court.country:
-                self.locality = self.court.locality
+            if court.country:
```
actlikewill: you can remove this `if court.country`,  its repeated above

Dennismwangi-k:
> @Dennismwangi-k django usually asks what you want to do with existing null values, which should be handled in by setting a value that will populate it. (you could set it to pk=1 maybe). It probably won't happen in prod but you should test it just in case. add a judgment without a court then run the migration and see what happens

yeah this is a very good point @actlikewill  but @longhotsummer  assured me there is no judgement that exists without a court, and since this is a guard that protects ever not to have a judgement without a court then that default will be a dead code, or not useful

actlikewill (approved):

PR #3135, opened 2026-04-13
Sentrybug code fix, for 403 errors

mugiwaramunyi:
Add 403 to the list of ignored errors.

### DIFF
--- a/peachjam/js/components/DocumentContent/pdf-renderer.ts
+++ b/peachjam/js/components/DocumentContent/pdf-renderer.ts
@@ -354,7 +354,7 @@ class PdfRenderer {
       err.innerText = message;
     }
 
-    if (response && response.status !== 404) {
+    if (response && response.status !== 404 && response.status !== 403) {
       throw new Error(`Error fetching PDF: ${message}`);
     }
   }


peachjam/js/components/DocumentContent/pdf-renderer.ts
```diff
@@ -354,7 +354,7 @@ class PdfRenderer {
       err.innerText = message;
     }
 
-    if (response && response.status !== 404) {
+    if (response && response.status !== 404 && response.status !== 403) {
```
actlikewill, line 357: 403 errors would indicate something is misconfigured somewhere. i think we would want to be aware of them

actlikewill: if this is for this issue: https://github.com/laws-africa/peachjam/issues/2534, the issue is kind of complex and the cause is not always one thing. Recently we were able to identify that the docker app ips can change when we do a deployment, which can trigger this issue. We are yet to figure out how to properly fix it. but we did add a config to ALLOWED_HOSTS so that now we should be getting 403s instead of 404s. We can review together whether that change actually did that.

longhotsummer (approved):

peachjam/js/components/DocumentContent/pdf-renderer.ts
mugiwaramunyi: @actlikewill noted, will go ahead and merge this fix since we repro-ed it and confirmed the cloud flare, but like we discussed in our sync, we can improve on the fix by building on our cloudflare config.

PR #3138, opened 2026-04-14
coveralls

actlikewill:
- This adds test coverage reports from coveralls

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,7 +65,7 @@ jobs:
           python -m pip install --upgrade "setuptools<82" wheel
           pip install .[dev]
           pip install psycopg[binary]==3.2.12 unittest-xml-reporting
-          npm ci
+          npm ci --no-audit --ignore-scripts --only=prod
           npm i -g sass
       - name: Run tests
         env:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,17 +1,31 @@
 name: Test
 on: [push, pull_request]
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
+
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: pip
+      - name: Install lint deps
+        run: |
+          python -m pip install --upgrade "setuptools<82" wheel
+          pip install .[dev]
+      - name: Run pre-commit
+        run: pre-commit run --all-files
+
   test:
+    needs: lint
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        app:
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -1,64 +1,101 @@
 name: Test
-on: [push, pull_request]
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
```
actlikewill, line 7: @longhotsummer i've changed tests to only run on pull request and on push to main. The test suite was running twice per app, for no real reason. We have rules that say only merge via an approved PR, so I think this is safe.

actlikewill: @longhotsummer please also confirm in the repo settings whether we have read and write permissions on for actions: Check Repo Settings → Actions → General This will allow coveralls to write a comment on the pr. but the status updates may be enough.
image

mugiwaramunyi (dismissed):

.github/workflows/test.yml
longhotsummer: I think changing the triggers will mean we won't have tests running on a push to a non-main branch, which I don't think is what we want?

.github/workflows/test.yml
```diff
@@ -1,64 +1,101 @@
...
-        node-version: '18'
-
-    - name: Install dependencies
-      run: |
-        sudo apt-get -y update
-        sudo apt-get install -y libreoffice poppler-utils pandoc
-        python -m pip install --upgrade "setuptools<82" wheel
-        pip install .[dev]
-        pip install psycopg[binary]==3.2.12 unittest-xml-reporting
-        npm ci --no-audit --ignore-scripts --only=prod
```
longhotsummer, line 48: why this change? this is (or at least was) the recommended way of doing this to prevent dodgy scripts from running during install, and we only need production dependencies.

longhotsummer:
The perms in the repo look correct. Should we be seeing a coveralls annotation somewhere on this PR?

.github/workflows/test.yml
actlikewill: a sanity check recommended that we allow all dev dependecies and scripts to install during ci tests. especially because we are end up doing them in manually (i.e sass is a dev dependancy and we do npm -i g, and pupeteer has post install scripts.) but closer inspection reveals that we do this for good reasons. sass needs a global install and pupeteer needs the chrome browser install, during deployment. i'll put the lines back

actlikewill:
> The perms in the repo look correct. Should we be seeing a coveralls annotation somewhere on this PR?

the checks have a coverage report if you expand it. coveralls should also be posting a comment but i dont see it.

.github/workflows/test.yml
actlikewill: @longhotsummer if you push a branch without a pull request, tests will not run. they will only trigger on creating a pull request and on updating a pull request. butif you push a branch without a pull request, tests will not run. they will only trigger on creating a pull request and on updating a pull request.  i dont see why one would push a branch without intending to make a pr. updates to prs will trigger tests as normal

longhotsummer: ok, let's see how it plays out.

longhotsummer (approved):

PR #3143, opened 2026-04-16
Fix Flynote.DoesNotExist error during delete

Dennismwangi-k:
Fixed he issue of trying to delete none existing flying notes, before were bulk deleting them all together which some flying notes were not exisitng and raise an error, currently  we get all the list of empty flying notes and delete them 1 by 1 #3120

### DIFF
--- a/peachjam/models/flynote.py
+++ b/peachjam/models/flynote.py
@@ -159,9 +159,13 @@ class FlynoteDocumentCount(models.Model):
                 root.slug,
                 root.pk,
             )
-            Flynote.objects.filter(path__startswith=root_path).filter(
-                document_count_cache__isnull=True
-            ).delete()
+            empty_flynotes = list(
+                Flynote.objects.filter(path__startswith=root_path)
+                .filter(document_count_cache__isnull=True)
+                .order_by("-depth", "-path")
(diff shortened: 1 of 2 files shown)


peachjam/models/flynote.py
```diff
@@ -159,9 +159,13 @@ def refresh_for_flynote(cls, root):
                 root.slug,
                 root.pk,
             )
-            Flynote.objects.filter(path__startswith=root_path).filter(
-                document_count_cache__isnull=True
-            ).delete()
+            empty_flynotes = list(
+                Flynote.objects.filter(path__startswith=root_path)
+                .filter(document_count_cache__isnull=True)
```
Dennismwangi-k, line 164: we are also avoiding complex deleting of whole tree instead small portion at a time

longhotsummer: I'm not sure this will fix the problem (which I think is a race condition), but let's try it.

longhotsummer (approved):

peachjam/models/flynote.py
Dennismwangi-k: sure lets try

PR #3146, opened 2026-04-17
Simple Guard to retry generating the flying notes

Dennismwangi-k:
Simple Guard to retry generating the flying notes when they are suspiciously less than 20 Characters


Dennismwangi-k:
@longhotsummer  put in mind if the second attempt is still short it will be saved

longhotsummer (approved):

PR #3147, opened 2026-04-17
ensure document content is saved after extracting citations

longhotsummer:
previously saving document was enough, but that is no longer used at all since the content doesn't live on it.

We will need to re-extract citations on any HTML documents since 1 Jan 2026.


longhotsummer:
Update logic:

```
from peachjam.models import CoreDocument
from peachjam.tasks import extract_citations

ids = list(CoreDocument.objects.filter(document_content__content_html_is_akn=False, document_content__source_html__isnull=False, updated_at__gte='2026-01-01').exclude(document_content__source_html="").values_list("pk", flat=True))
for i in ids:
  extract_citations(i)
```

mugiwaramunyi (approved):

longhotsummer:
Backfilling done.

Issue #3144, opened 2026-04-17
WCAG 2.1 AA Compliance: Identified Issues & Proposed Fix Plan


Dennismwangi-k:
This issue consolidates accessibility findings identified during WCAG 2.1 AA review across liiweb

1. Skip Links
      - [x] Update skip link wording to “Skip to main content” where applicable
      - [x] Ensure skip link wording is correct per page type (not always “document”)
      - [x]  Replace #top skip target with a proper main content anchor
      - [x]  Fix skip link visibility/clipping at 200% zoom
2. Focus Visibility
      - [x] Improve focus styles on solid buttons eg search buttons etc.
3. Keyboard Accessibility (Filters & Navigation)
      - [x] Make alphabet filter keyboard reachable
      - [x]  Add skip link from filters → results on listing pages
      - [x]  Add skip link to skip all filters → list items
 4. Automatic Filters (HTMX behaviour)
      - [x] Inform users that filters auto-update results
      - [ ]  Add visible feedback when results update eg color flash
5. Sort Controls
      - [x] Add clear semantics to sort controls {instead of Title use something like Alphabetical}
      - [x]  Indicate ascending/descending order {{make it bolder}}
      - [x]  Ensure sort state is communicated to screen readers{{ eg when sorting lall listing pages we need to communicate to      them the sort state }}    
6. Target="_blank" Links
      - [ ]  Remove unnecessary target="_blank" usage serach results page.{{Make sure back button works}}
7. Form Accessibility
      - [x]  Ensure all inputs have visible labels, fix unclear labels (e.g. “sort” dropdown → more descriptive)
8. Color Contrast
      - [x]  Fix contrast for Advanced Search link {Ensure all text meets minimum 4.5:1 contrast ratio}
      - [x]  Fix contrast for Help link {{Ensure all text meets minimum 4.5:1 contrast ratio}}

PR #3148, opened 2026-04-17
listing pages a11y improvements

actlikewill:
This adds a11y improvements for filtered listing pages
- adds skip links to the filters, hidden by default, visible on focus, 
- adds aria-live for the document results, allows screen readers to announce live content updates for htmx
- adds skip links for the long year lists and registry lists, reduces tab fatigue
- adds skip to results, to bypass filters entirely

https://www.loom.com/share/c5ede7a21b8b4ce6bd79a83f4a4bb44c

https://www.loom.com/share/b12bf52a382441f5b2c384811be2783a

addresses https://github.com/laws-africa/peachjam/issues/3144


peachjam/js/components/document-filter-form.ts
```diff
@@ -86,6 +87,37 @@ export default class DocumentFilterForm {
...
 
       optionEl.classList.toggle('d-none', !isVisible);
+      optionEl.hidden = !isVisible;
+      optionEl.setAttribute('aria-hidden', String(!isVisible));
     });
   }
+
+  setupSkipLinks () {
+    this.root.querySelectorAll('[data-skip-link]').forEach((link) => {
+      link.addEventListener('click', (event: Event) => {
```
longhotsummer, line 97: what is this doing?

longhotsummer:
Interesting!

peachjam/js/components/document-filter-form.ts
actlikewill: it closes the off canvas element on mobile when you skip a link to the results

actlikewill:
@Dennismwangi-k @longhotsummer i'm happy to merge this. I've been doing a lot testing with screen readers and tweaking it and its now better than it before. Particularly, adding fieldset and legend markup to the filters greatly improves the experience. It now announces the panels as you land on them which is much better than before. The aria-live also calls out page updates. 
The main idea with screen readers is that you are presenting a stream of information, much like how you would navigate a self service menu on a customer care phone number. the call outs  should be clear and help orient the user on the page. The best way to achieve this is to manually test with a screen reader thoroughly and use your best judgment on how it should be designed. There is still room for improvement, and i plan to keep making incremental improvments as we go.

peachjam/views/generic_views.py
```diff
@@ -261,13 +261,31 @@ def get_context_data(self, **kwargs):
...
+            if facet.get("options")
+        ]
+        for index, item in enumerate(rendered_facets):
+            next_item = (
+                rendered_facets[index + 1] if index + 1 < len(rendered_facets) else None
+            )
+            item["next_target_id"] = (
+                f'{context["doc_table_form_id"]}-group-{next_item["name"]}'
+                if next_item
+                else f'{context["doc_table_id"]}-results-heading'
```
Dennismwangi-k: This is assuming every filtered table is rendering  #{{ doc_table_id }}-results-heading, but some tables like in this 
http://127.0.0.1:8000/unconstitutional-provisions/#doc-table-unconstitutional-provisions-results-heading
dont render that target so the link or tab will point to a missing anchor. No need to show it or even show the link tab in such case

actlikewill: @Dennismwangi-k thanks. added a default value for such forms so we always have tab to land on

actlikewill:
@Dennismwangi-k @longhotsummer this is ready in to be merged. the coverage drop is a marginal statistical drop :sweat_smile:

longhotsummer:
@actlikewill would it be possible to record a short loom showing some of the changes, in particular:

* that the offcanvas filters work and close correctly on desktop and mobile
* the new skiplinks functionality
* what the updated alphabetical filters look like

peachjam/templates/peachjam/provision_enrichment/_unconstitutional_table.html
```diff
@@ -1,5 +1,14 @@
 {% load i18n peachjam %}
-<div id="doc-table">
```
longhotsummer, line 2: pls include in the video that this still works correctly, because it's quite different to what it was.

peachjam/views/generic_views.py
```diff
@@ -261,13 +261,31 @@ def get_context_data(self, **kwargs):
...
+            )
+        context["rendered_facets"] = rendered_facets
```
longhotsummer, line 287: if this is just setting `next_target_id`, can you not replace `facet_data` directly? Do we need both `rendered_facets` and `facet_data`?

actlikewill: the next target id is different for the last item in the list. It goes to the results rather than the next filter group. The data structure changes to a list to enforce the ordering. Existing tests still check the facet_data object at various points so the object still has relevance in the grand scheme of things. but i can see that it could be a problem if we have too much data in the context at some point.

actlikewill:
@longhotsummer @Dennismwangi-k i've added the videos. Like i mention, if you are able to fire up your screen reader and test for yourself it would be great. Thats where a lot of the value of the work we have done reveals itself. This can be improved further, and i plan on doing so, but we are much better than where we started. My focus has been on the listing pages, but i'm sure that more work can be done on all the pages. Its a journey that never ends!

longhotsummer:
This is excellent and very interesting, thanks Wilson.

longhotsummer (dismissed):

PR #3149, opened 2026-04-17
fix the Flynote.DoesNotExist

Dennismwangi-k:
So yeah @longhotsummer  ignoring it was ok but lets treat it coz i found the issue was we have 2 workers that were doing the same thing, so when one was refresh counting the other one was deleting so i just locked the flying note  root before refreahing. This im kinda confidence will fix that issue

### DIFF
--- a/peachjam/tasks.py
+++ b/peachjam/tasks.py
@@ -394,11 +394,10 @@ def update_flynote_taxonomy(judgment_id):
 
 
 @background(queue="peachjam", remove_existing_tasks=True, schedule={"priority": -1})
-@transaction.atomic
 def refresh_flynote_document_count(root_id):
     from peachjam.models.flynote import Flynote, FlynoteDocumentCount
 
-    root = Flynote.objects.select_for_update().filter(pk=root_id).first()
+    root = Flynote.objects.filter(pk=root_id).first()
     if not root:
         log.info(f"No flynote root with id {root_id} exists, ignoring.")
         return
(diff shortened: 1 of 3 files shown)


peachjam/tasks.py
```diff
@@ -394,10 +394,11 @@ def update_flynote_taxonomy(judgment_id):
 
 
 @background(queue="peachjam", remove_existing_tasks=True, schedule={"priority": -1})
+@transaction.atomic
 def refresh_flynote_document_count(root_id):
     from peachjam.models.flynote import Flynote, FlynoteDocumentCount
 
-    root = Flynote.objects.filter(pk=root_id).first()
+    root = Flynote.objects.select_for_update().filter(pk=root_id).first()
```
longhotsummer: Let's see if locking helps, it should. It makes more sense to me to do this locking inside refresh_for_flynote, rather than here, since that already has transactions and locking.

peachjam/tests/test_flynotes.py
longhotsummer, line 1: this is a lot of work which probably isn't worth testing.

longhotsummer: Does this test fail if the changes to tasks.py are reverted? If not, then it's not a useful test.

Dennismwangi-k: ooh yeah ok i can remove them was trying to replicate same issue on my local machine by using test file 
let me remove them

Dennismwangi-k: didnt need it so just do away

Dennismwangi-k:
@longhotsummer  yeah this one might fix now one worker will have to wait for the other and exits cleanly if its deleted

longhotsummer (approved):

PR #3150, opened 2026-04-17
Fix button focus styles

Dennismwangi-k:
#3144   For this PR im trying to improve the focus around the solids buttons to allign with WCAG 2.1 AA compliannce. I have used black round ring for this coz, i wanted to use blue since other links were using blue eg advanced button but the issue is some sites are using that blue as primary color for their buttons like tanzlii so using Black waas the best option

Before:
Screenshot 2026-04-17 at 7 47 06 PM

After:
Screenshot 2026-04-17 at 7 47 20 PM

### DIFF
--- a/peachjam/static/stylesheets/components/_buttons.scss
+++ b/peachjam/static/stylesheets/components/_buttons.scss
@@ -1,0 +1,22 @@
+@mixin solid-button-focus-ring {
+  outline: 3px solid $gray-900;
+  outline-offset: 2px;
+  box-shadow: 0 0 0 0.2rem $white;
+}
+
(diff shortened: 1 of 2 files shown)


longhotsummer:
I'm surprised Bootstrap doesn't do something like this already. What do the bootstrap docs say about accessibility and buttons?

Dennismwangi-k:
> I'm surprised Bootstrap doesn't do something like this already. What do the bootstrap docs say about accessibility and buttons?

let me answer it like yes it does but not distinct enough in our theme solid buttons to meet WCAG so i had to strengthen it to meet that

longhotsummer (approved):

PR #3151, opened 2026-04-20
I have fixed the color contrast for advanced and help link so as to meet the WCAG requirements

Dennismwangi-k:
In this Pr im trying to make sure the links (advanced link and help link) meets minimum 4.5:1 contrast ratio as suggested by WCAG, you might not notice the difference but the aim here is to make sure if the background is pale make sure the links are well visible as welll

Before:
Screenshot 2026-04-20 at 11 30 41 AM

After:

Screenshot 2026-04-20 at 11 34 26 AM

### DIFF
--- a/liiweb/static/stylesheets/components/_hero.scss
+++ b/liiweb/static/stylesheets/components/_hero.scss
@@ -23,8 +23,7 @@
   background-color: transparentize($primary, 0.90);
 }
 
-.hero-search__aux-link,
-.hero-search__aux-links a {
+.hero-search a:not(.btn) {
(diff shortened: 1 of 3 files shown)


longhotsummer:
It would make more sense to me to set the visited colour for any link inside `.hero-search` -- there's nothing special about advanced search.

Dennismwangi-k:
> It would make more sense to me to set the visited colour for any link inside `.hero-search` -- there's nothing special about advanced search.

Had only saw the issue on those 2 links only, but youare right we should target all normal links inside the hero search, tahnks

longhotsummer (approved):

PR #3153, opened 2026-04-20
Fix flynote update failures

Dennismwangi-k:
so in this branch im fixing 2 issues, @longhotsummer  i havent addressed child-first pruning coz this i had already done, so in this i have addressed only the 'NoneType' / 'add_sibling' error, whereby i refetch and lock the parent node before calling add_child() and retry transient tree-state failures instead of crashing immediately and for the second error the update_for_judgment() now parses first and only deletes existing JudgmentFlynote links after parsing succeeds. If parsing raises IndexError, we log the judgment id and a short flynote excerpt, then return early.

### DIFF
--- a/peachjam/analysis/flynotes.py
+++ b/peachjam/analysis/flynotes.py
@@ -2656,22 +2656,58 @@ class FlynoteUpdater:
             self.node_cache[expected_slug] = existing
             return existing
 
-        try:
-            if parent:
-                # Treebeard uses parent state such as numchild when adding a child.
-                # Cached parent instances can go stale across multiple judgments,
-                # so refresh before inserting under an existing node.
-                parent.refresh_from_db(fields=["path", "depth", "numchild"])
-                node = parent.add_child(name=name, slug=expected_slug)
-            else:
+        if not parent:
+            try:
                 node = Flynote.add_root(name=name, slug=expected_slug)
-            self.node_cache[expected_slug] = node
-            return node
-        except IntegrityError:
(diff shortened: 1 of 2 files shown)


Dennismwangi-k:
@longhotsummer  i made some few changes and im not sure if they make sense, so before we were deleting the old JudgmentFlynote link first, then afterwards try ti create new, so if it fails might end up with no flynote link, now what im doing is im trying to rebuild first then replace only of the rebuild was success this will make flynote relinking fail safely

peachjam/analysis/flynotes.py
```diff
@@ -30,6 +30,7 @@
 from peachjam.tasks import refresh_flynote_document_count
 
 log = logging.getLogger(__name__)
+DEFER_FLYNOTE_UPDATE = object()
```
Dennismwangi-k, line 33: this will be our unique identifire that will help us abort the udate and keep the old link

peachjam/analysis/flynotes.py
```diff
@@ -2656,41 +2657,86 @@ def get_or_create_node(self, parent, name):
...
-            return existing
+                self.node_cache[expected_slug] = node
+                return node
+            except IntegrityError:
+                existing = Flynote.objects.filter(slug=expected_slug).first()
+                if existing:
+                    self.node_cache[expected_slug] = existing
+                return existing
+
+        for attempt in range(1, 4):
```
Dennismwangi-k, line 2671: we retry 3 times, if still fail we stop relink attempt and the existing remains unchanged

PR #3156, opened 2026-04-21
fix(chat): guard against non-string content in setMessageContent

### LINKED ISSUE #3119: TypeError: e.content.trim is not a function
Sentry Issue: [LII-486](https://lawsafrica.sentry.io/issues/7377135275/?referrer=github_integration)

```
TypeError: e.content.trim is not a function
  at Cm.methods.setMessageContent (peach-jam/./peachjam/js/components/chat/DocumentChat.vue:305:52)
  at apply (peach-jam/./peachjam/js/components/chat/DocumentChat.vue:255:14)
```

armorbreak001:
## Summary

Fixes #3119 — `TypeError: e.content.trim is not a function` crash in DocumentChat.

## Root Cause

`setMessageContent()` assigns `content || ''` to `message.content`, but when the SSE payload delivers a non-string value (object, number, etc.), the fallback `|| ''` doesn't catch it (since objects are truthy). The subsequent `.trim()` call then throws.

## Fix

Coerce `content` to string explicitly: use `typeof` check + `String()` fallback instead of relying on truthiness.

**Changed file:** `peachjam/js/components/chat/DocumentChat.vue` (1 line)

### DIFF
--- a/peachjam/js/components/chat/DocumentChat.vue
+++ b/peachjam/js/components/chat/DocumentChat.vue
@@ -301,7 +301,7 @@ export default {
       return message;
     },
     setMessageContent (message, content) {
-      message.content = content || '';
+      message.content = typeof content === 'string' ? content : (content != null ? String(content) : '');
       if (message.role === 'ai' && message.content.trim().length > 0) {
         message.content_html = marked.parse(message.content);
       } else {


longhotsummer:
No longer required.

PR #3158, opened 2026-04-21
main page heading

actlikewill:
This adds the main-page-heading. 
- makes main content a focusable tab stop with an aria-labelledby
- allows screen readers to announce the main page heading on navigation.

### DIFF
--- a/peachjam/templates/peachjam/journal/journal_list.html
+++ b/peachjam/templates/peachjam/journal/journal_list.html
@@ -3,10 +3,11 @@
 {% block title %}
   {% trans 'Journals' %}
 {% endblock %}
+{% block main-aria-labelledby %}aria-labelledby="main-page-heading"{% endblock %}
 {% block page-content %}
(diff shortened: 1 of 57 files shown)


peachjam/templates/peachjam/journal/journal_list.html
```diff
@@ -3,10 +3,11 @@
 {% block title %}
   {% trans 'Journals' %}
 {% endblock %}
+{% block main-aria-labelledby %}aria-labelledby="main-page-heading"{% endblock %}
```
longhotsummer: what if you just make this the default for the block, and it assumes that all pages have an element with `id="main-page-heading"` ?

actlikewill: thanks @longhotsummer much better. some more small changes on the share document links, adding aria labels

longhotsummer (approved):

Dennismwangi-k (approved):

PR #3159, opened 2026-04-21
allow MNCMatcher to span lines in pdfs

longhotsummer:
this handles citations that cross lines and so don't have whitespace in their extracted text.

image


Dennismwangi-k (approved):
lgtm

PR #3160, opened 2026-04-21
Fixed the issue of skip to document where unnecessary

Dennismwangi-k:
#3144 
In this Pr i have addressed WCAG issue and a bug, so the WCAG  issue is the skip links were not visible at all they were not visible at all especially when you zoom to 200 so i have fixed it to bivisible as below.
https://www.w3.org/WAI/WCAG22/Understanding/resize-text.html
https://www.w3.org/TR/WCAG22/#resize-text
for the bug was that we were showing skip to document evn on the pages where we have no documents, you find it appearing to all pages

Before:
Screenshot 2026-04-21 at 8 09 13 PM
Screenshot 2026-04-21 at 8 09 50 PM

After:

Screenshot 2026-04-21 at 8 10 14 PM
Screenshot 2026-04-21 at 8 10 33 PM

### DIFF
--- a/peachjam/templates/peachjam/_header.html
+++ b/peachjam/templates/peachjam/_header.html
@@ -3,10 +3,10 @@
   {% block skip-links %}
     <div class="skip-links">
       {% if 'akn/' in request.path %}
-        <a href="#document-content">{% trans 'Skip to document content' %}</a>
+        <a class="bg-white" href="#document-content">{% trans 'Skip to document content' %}</a>
       {% endif %}
-      <a href="#main-nav">{% trans 'Skip to main menu' %}</a>
-      <a href="#header-search">{% trans 'Skip to search' %}</a>
+      <a class="bg-white" href="#main-nav">{% trans 'Skip to main menu' %}</a>
+      <a class="bg-white" href="#header-search">{% trans 'Skip to search' %}</a>
     </div>
   {% endblock %}
   {% block top-bar %}<div>Top bar content</div>{% endblock %}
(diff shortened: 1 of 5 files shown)


peachjam/templates/peachjam/_header.html
```diff
@@ -2,8 +2,9 @@
 <header>
   {% block skip-links %}
     <div class="skip-links">
-      <a class="bg-white"
-         href="{% if 'akn/' in request.path %}#document-content{% else %}#top{% endif %}">{% trans 'Skip to document content' %}</a>
+      {% if 'akn/' in request.path %}
```
actlikewill, line 5: the idea here was to show skip to document content on document detail pages. i've actually removed this conditional entirely, just so we can be more consistent in all our pages. the skip to document content link now exists above the document table of contents where it is most useful.

actlikewill: @Dennismwangi-k  we still need the skip to main content link, i suggest that you ignore the changes in this pr and i will add the z-index on my own pr.

Dennismwangi-k: @actlikewill  yeah but we have skip to main menu which they act as the same, if we add both of them i them, i had done that on the first push but when testing i realised thy are doing the same so there was no need to add both we can just keep skip tp main menu. About skip to document before the idea was to show that link when necessary but i think the condition was not working so i tightened it. When you mean i ignore the changes i dont understand @actlikewill  you mean i discard the changes? if yes why are you not checking the issue list on the tasks you are doing please so to avoide double work please? please follow the issue list we created together coz now there is alot of confusion

Dennismwangi-k:
@actlikewill @longhotsummer  the skip to document now looks ok coz now it works well its shown only on the required pages, please you can test and let me know

peachjam/templates/peachjam/_header.html
actlikewill: no they are different,

PR #3164, opened 2026-04-23
handling target blank links

actlikewill:
This handles links that open new tabs
- by default all internal links should not open new tabs. 
- external links can open new tabs, but include a visually hidden call out for speech readers. Ideally we should also have an icon, but i've ignored that.

### DIFF
--- a/africanlii/templates/africanlii/_lii_info.html
+++ b/africanlii/templates/africanlii/_lii_info.html
@@ -7,7 +7,7 @@
       {% endblocktrans %}
     </div>
     <div>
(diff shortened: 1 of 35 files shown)


longhotsummer (commented):
I'm surprised screen readers don't handle this automatically, since they can determine from the link markup that it will open a new tab.

actlikewill:
> I'm surprised screen readers don't handle this automatically, since they can determine from the link markup that it will open a new tab.

no they dont. They may announce it after the window or tab opens. its best practice to communicate this in advance.

longhotsummer (approved):

PR #3165, opened 2026-04-23
Match flynotes to database entries when generating summaries

longhotsummer:
* change the flynote summariser to an agent, so that it can handle multiple tool calls
* add a feature flag for matching flynotes against the database
* if enabled, the agent uses a tool to search the flynote database and adjusts its flynotes based on what's available in the db, preferring well-used flynotes over new ones


peachjam/analysis/summariser.py
```diff
@@ -93,55 +289,89 @@ def normalise_summary(self, summary):
...
+                "content": FLYNOTE_MATCH_PROMPT + self.top_level_flynotes_prompt(),
+            }
+        )
+
+        # give the agent tools and change its output type
+        agent = self.agent.clone(tools=[search_flynotes], output_type=Flynote)
+        run_result = Runner.run_sync(agent, input)
+        log_agent_reasoning(self.run_result)
+
+        summary.flynote = run_result.final_output.flynote
```
Dennismwangi-k, line 338: we have updated the summary.flynote here but i dont think you have carried it to the translated summury so making it rebuild from the old state

peachjam/analysis/summariser.py
```diff
@@ -93,55 +289,89 @@ def normalise_summary(self, summary):
...
+        )
+
+    def translate_summary(self, summary, language):
+        if not language or language == "English":
+            return summary
+
+        log.info(f"Translating judgment summary into {language}")
+
+        # build on the original input for the translation, to give the model more context to work with
+        input = self.run_result.to_input_list()
```
Dennismwangi-k, line 366: so this ends up using the old one not the updtaed one

peachjam/analysis/summariser.py
```diff
@@ -93,55 +289,89 @@ def normalise_summary(self, summary):
...
+        input.append(
+            {
+                "role": "developer",
+                "content": FLYNOTE_MATCH_PROMPT + self.top_level_flynotes_prompt(),
+            }
+        )
+
+        # give the agent tools and change its output type
+        agent = self.agent.clone(tools=[search_flynotes], output_type=Flynote)
+        run_result = Runner.run_sync(agent, input)
```
Dennismwangi-k, line 335: maybe you should update this to use the new updated one

peachjam/analysis/summariser.py
longhotsummer: good catch, fixed.

peachjam/analysis/summariser.py
longhotsummer: I don't want the translator to be confused by the flynote matching prompt, so I've kept them separate.

Dennismwangi-k: ooh ok

Dennismwangi-k (approved):
lgtm now

Issue #3145, opened 2026-04-17
Set up continuous accessibility testing using axe-core + puppeteer


Dennismwangi-k:
To prevent accessibility regressions and improve WCAG 2.1 AA compliance across LII sites, we need to introduce automated accessibility testing into our CI pipeline.

PR #3166, opened 2026-04-23
Ci files for running accessibility tests for WCAG

Dennismwangi-k:
In this PR I’m addressing issue #3145  I have added puppeteer and axe-core to automate accessibility testing in our sites. Puppeteer will open the mapped pages in a browser, while axe-core it will be running automated WCAG 2.1 AA checks against those pages and prints the findings directly in the terminal for local runs. I have also added a separate accessibility GitHub Actions workflow that runs on a two-week schedule and can also be triggered manually when needed. The workflow will discover the site shells in the repo, runs the accessibility scan against the mapped pages for each site, and if blocking issues are found, it updates a single GitHub issue with the findings from the failing sites.
This will gives us:
  - repeatable automated WCAG checks for key shared page types
  - local terminal output that is easy to read while fixing issues
  - scheduled monitoring without slowing down the normal PR test workflow
I have  attached a short video below to  show you how to run the checks locally and how to read the output.

https://www.loom.com/share/cd89956abdc440bf90fcf5eba2951fac


.github/workflows/a11y.yml
```diff
@@ -0,0 +1,299 @@
+name: Accessibility
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 3 1,15 * *"
```
Dennismwangi-k, line 6: this scheduled it to be running after 2 weeks meaning 2 times a month, and will create an issue that document all the findings

actlikewill: we don't have a dedicated server where this will run reliably, we'll have to figure it out

Dennismwangi-k: @actlikewill  does a a github action really need a server i doubt it it runs in github hosted runners unless im wrong

PR #3168, opened 2026-04-24
Improve management of flynote tree

longhotsummer:
* remove flynote slugs; they are difficult to maintain and not really useful
* make it possible to deprecate flynotes, which triggers re-summarisation of associated judgments but prevents the AI from re-using the deprecated parts of the tree
* add flynote_raw. This allows us to avoid circular updates between the raw flynote, the Flynote objects, and the rendered flynote
* add admin area support for flynotes
* flynote parser now assumes that flynotes are clean (because they are generated in a well-structured way), and so by default doesn't try to re-interpret the flynote structure.
* add a management command to bulk-serialise Flynote tree topics into text, so that the flynote tree and the rendered text are the same
* when flynote name or path changes (ie. it was moved in the tree), then linked judgments must re-serialise their flynotes

The net result of all of this is that we can now manipulate the flynote tree (eg. rename nodes, move nodes, deprecate nodes) and the linked judgments will be updated. This means we can manage the entire flynote system in bulk, by managing the flynote tree.


peachjam/migrations/0296_remove_flynote_slug.py
longhotsummer, line 1: I must move this into a separate PR otherwise it'll break existing code.

peachjam/models/judgment.py
```diff
@@ -711,6 +712,28 @@ def generate_summary(self):
...
 
+    @on_attribute_changed(BEFORE_SAVE, ["flynote_raw"], ["flynote", "flynote_tree"])
+    def sync_flynote(self):
+        """The flynote_raw field has changed, copy it to flynote and populate the flynote tree (if configured)."""
+        self.flynote = self.flynote_raw
+
+        if settings.PEACHJAM["SUMMARISE_USE_FLYNOTE_TREE"]:
+            # This will eventually update both flynote and flynote_raw to match the flynote tree.
+            # We set flynote above since this background task may take a while to run.
+            update_flynote_taxonomy(self.pk)
```
Dennismwangi-k, line 723: wount this affect the new judgement since no id is created, since we can only get the ID after we save to database, so if we run this hook before we save the tree wount be build and will fail silently, can you please try checking on this.

longhotsummer: good catch, fixed. I now check for self.pk.

PR #3169, opened 2026-04-24
fixed the error for vue function

Dennismwangi-k:
this pr is addressing the issue #3121 ,so when it was trying to show a fallback transalated message it failed coz "$t" was not available for "this" so it crushed. so the fix. also as i was fixing i realised another bug whereby when Try again button called "load" but the real method it should have been calling was "loadThread()" . Also another bug was successful retry didnt clear the old error, so the warning could remain still visible after recovery.

### DIFF
--- a/peachjam/js/components/chat/DocumentChat.vue
+++ b/peachjam/js/components/chat/DocumentChat.vue
@@ -171,7 +171,7 @@ export default {
       } catch (err) {
         console.error(err);
         this.threadId = '';
-        this.error = this.fallbackErrorMessage(err);
+        this.error = err?.message?.trim() || this.$t('Something went wrong. Please try again.');
       }
     },
     /**
@@ -202,7 +202,7 @@ export default {
         }
       } catch (err) {
         console.error(err);
-        this.error = this.fallbackErrorMessage(err);
+        this.error = err?.message?.trim() || this.$t('Something went wrong. Please try again.');
       }
     },
     async submit () {
@@ -270,7 +270,7 @@ export default {
 
       es.addEventListener('error', err => {
         console.error(err);
-        this.error = this.fallbackErrorMessage(err);
+        this.error = err?.message?.trim() || this.$t('Something went wrong. Please try again.');
         // remove the message so that the user can try again
         const lastMessage = this.messages[this.messages.length - 1];
         if (!lastMessage || lastMessage.role !== 'ai') {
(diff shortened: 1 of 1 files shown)


longhotsummer:
All our other vue.js code uses `this.$t(...)` -- why doesn't it work in this case?

Dennismwangi-k:
> All our other vue.js code uses `this.$t(...)` -- why doesn't it work in this case?

i focused more on adding a gurad to prevent the error and avoid relyong on this.$t() this will work but i can keep it and fix on the root whereby when we unmount the stream still emit error coz the unmount didnt really close well, so making it run against som component instance that are already torn down so some function crush. So if i apply this unmounted() { this.closeStream(this.eventSource); } this should work as well coz i will close the stream when the component unmounts

longhotsummer:
> > All our other vue.js code uses `this.$t(...)` -- why doesn't it work in this case?
> 
> i focused more on adding a gurad to prevent the error and avoid relyong on this.$t() this will work but i can keep it and fix on the root whereby when we unmount the stream still emit error coz the unmount didnt really close well, so making it run against som component instance that are already torn down so some function crush. So if i apply this unmounted() { this.closeStream(this.eventSource); } this should work as well coz i will close the stream when the component unmounts

So are you saying that `this.$t()` goes away when the component is unmounted?

Dennismwangi-k:
> > > All our other vue.js code uses `this.$t(...)` -- why doesn't it work in this case?
> > 
> > 
> > i focused more on adding a gurad to prevent the error and avoid relyong on this.$t() this will work but i can keep it and fix on the root whereby when we unmount the stream still emit error coz the unmount didnt really close well, so making it run against som component instance that are already torn down so some function crush. So if i apply this unmounted() { this.closeStream(this.eventSource); } this should work as well coz i will close the stream when the component unmounts
> 
> So are you saying that `this.$t()` goes away when the component is unmounted?

yes

peachjam/js/components/chat/DocumentChat.vue
```diff
@@ -136,6 +136,9 @@ export default {
   mounted () {
     this.loadThread();
   },
+  unmounted () {
```
Dennismwangi-k, line 139: this will now close the running events during unmount, so it should not raise that error anymore

longhotsummer (approved):

PR #3170, opened 2026-04-24
Flynote merge

longhotsummer:
Adds a mechanism for an editor to safely merge flynotes.

image

* the user can search for similar flynotes to merge into the current one, and choose which ones should be  merged in
* the flynotes are updated, all linked judgments are updated, and the old flynotes (and their entire trees) are deleted

### DIFF
--- a/peachjam/models/flynote.py
+++ b/peachjam/models/flynote.py
@@ -9,6 +9,7 @@ from django_lifecycle import AFTER_SAVE, LifecycleModelMixin
 from treebeard.mp_tree import MP_Node
 
 from peachjam.models.lifecycle import on_attribute_changed
+from peachjam.tasks import refresh_flynote_document_count
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


peachjam/models/flynote.py
```diff
@@ -129,6 +130,99 @@ def cascade_deprecated(self):
...
+                        judgment_flynote.delete()
+                    else:
+                        judgment_flynote.flynote = target
+                        judgment_flynote.save(update_fields=["flynote"])
+
+                for child in list(source.get_children()):
+                    # moving nodes updates path attributes, so always work with latest info from db
+                    child.refresh_from_db()
+                    target.refresh_from_db()
+                    child.move(target, pos="sorted-child")
```
Dennismwangi-k, line 219: @longhotsummer  after we merge the other branch we will review this again just to confirm after we merge the parent and  had a child under that parent and a judgement was linked to that child  if the Judgment.flynote and Judgment.flynote_raw   get reserialised as well coz i can see here we are only moving them(the children). I don’t see what would reserialise judgments linked under that moved child branch. Also add a test for this please.

Dennismwangi-k (changes requested):

peachjam/models/flynote.py
longhotsummer: There's a post-save signal on JudgmentFlynote which will cause `serialise_judgment_flynote_tree` to run.

longhotsummer: I've added a test.

Dennismwangi-k: ok cool

Dennismwangi-k (approved):
lgtm

PR #3173, opened 2026-04-28
updated the flynote for refresh and prune to be runnning after 24 hours

Dennismwangi-k:
This pr updates the flynote refresh and prune flow so that it can run after a delay of 24 hours


peachjam/tasks.py
```diff
@@ -431,6 +433,27 @@ def refresh_flynote_document_count(root_id):
     FlynoteDocumentCount.refresh_for_flynote(root)
 
 
+def queue_refresh_flynote_document_count(root_id, schedule=FLYNOTE_REFRESH_DELAY):
+    existing_task = (
```
longhotsummer: I don't think this does anything useful. The background task system already de-duplicates tasks that aren't yet already running (`locked_at__isnull=True`), so I don't think this is doing anything extra.

We are getting "duplicate" tasks because the task takes a long time to run, and so a new one is queued up because there is already one running. I suggest we simply run this for all flynotes every 24 hours. I'll send you details of how to do that.

peachjam/tasks.py
```diff
@@ -431,6 +433,27 @@ def refresh_flynote_document_count(root_id):
...
+            kwargs={},
+        )
+        .filter(locked_at__isnull=True)
+        .order_by("run_at")
+        .first()
+    )
+    if existing_task:
+        return existing_task
+
+    return refresh_flynote_document_count(
```
longhotsummer: If we configure it like this, then any existing task remains scheduled (without changing the scheduled time):

```
schedule={'run_at': FLYNOTE_REFRESH_DELAY, 'action': TaskSchedule.CHECK_EXISTING}
```

and drop `remove_existing_tasks` (it defaults to False).

longhotsummer: All of this can be done via the `@background` decorator, so this helper function is not necessary.

peachjam/models/flynote.py
```diff
@@ -364,17 +366,28 @@ def refresh_for_flynote(cls, root):
...
                 .filter(document_count_cache__isnull=True)
+                .annotate(has_linked_judgments=Exists(linked_judgments))
+                .filter(has_linked_judgments=False)
                 .order_by("-depth", "-path")
             )
             for flynote in empty_flynotes:
-                flynote.delete()
+                try:
+                    flynote.delete()
+                except ProtectedError:
```
longhotsummer, line 391: if you also catch, log and ignore `Flynote.DoesNotExist` then it continues even if a flynote has been deleted under the hood.

peachjam/tasks.py
```diff
@@ -415,20 +417,92 @@ def update_flynote_taxonomy(judgment_id):
...
+            if cleared_pending:
+                Flynote.get_root_nodes().filter(pk=root.pk).update(
+                    document_count_refresh_pending=True
+                )
+            raise
+        refreshed += 1
+    return refreshed
 
 
 @background(queue="peachjam", remove_existing_tasks=True, schedule={"priority": -1})
```
longhotsummer:
```suggestion
@background(queue="peachjam", schedule={"priority": -1, "run_at": FLYNOTE_REFRESH_DELAY, "action": TaskSchedule.CHECK_EXISTING})
```

peachjam/signals.py
```diff
@@ -304,3 +306,20 @@ def judgment_flynote_saved_serialise_judgment(sender, instance, raw, **kwargs):
 @receiver(signals.post_delete, sender=JudgmentFlynote)
 def judgment_flynote_deleted_serialise_judgment(sender, instance, **kwargs):
     serialise_judgment_flynote_tree(instance.document_id)
+
+
+@receiver(signals.pre_delete, sender=Judgment)
```
longhotsummer: these can both be replaced by additions to the two JudgmentFlynote-related functions above

peachjam/tasks.py
```diff
@@ -13,6 +13,8 @@
 
 log = logging.getLogger(__name__)
 
+FLYNOTE_REFRESH_DELAY = 24 * 60 * 60
```
longhotsummer:
```suggestion
FLYNOTE_REFRESH_DELAY = 12 * 60 * 60
```

peachjam/signals.py
```diff
@@ -298,9 +299,17 @@ def chat_thread_deleted(sender, instance, **kwargs):
...
+        root_id = instance.flynote.get_root().pk
+        transaction.on_commit(
+            lambda root_id=root_id: refresh_flynote_document_count(root_id)
+        )
         serialise_judgment_flynote_tree(instance.document_id)
 
 
 @receiver(signals.post_delete, sender=JudgmentFlynote)
 def judgment_flynote_deleted_serialise_judgment(sender, instance, **kwargs):
+    root_id = instance.flynote.get_root().pk
```
longhotsummer: note that this may fail if `instance.flynote` has not been loaded from the database, because it now can't load it because `instance` has been deleted. The work-around is to do something like `flynote = Flynote.objects.filter(instance.flynote_id).first()`

peachjam/tasks.py
```diff
@@ -415,20 +417,37 @@ def update_flynote_taxonomy(judgment_id):
...
+    affected_root_ids = FlynoteUpdater().update_for_judgment(judgment)
+    for root_id in affected_root_ids:
+        refresh_flynote_document_count(
+            root_id,
+        )
 
 
-@background(queue="peachjam", remove_existing_tasks=True, schedule={"priority": -1})
-def refresh_flynote_document_count(root_id):
+def refresh_flynote_roots_now(root_ids):
```
longhotsummer: This is not necessary. The logic to load the flynote and call `refresh_for_flynote` can just be in `refresh_flynote_document_count()`.

longhotsummer (changes requested):

peachjam/management/commands/update_flynote_taxonomies.py
```diff
@@ -111,8 +114,8 @@ def handle(self, *args, **options):
         if last_pk:
             msg += f" Last pk processed: {last_pk}."
 
-        if not skip_counts and processed > 0:
-            FlynoteDocumentCount.refresh_for_all_flynotes()
+        if not skip_counts and affected_root_ids:
+            refresh_flynote_roots_now(affected_root_ids)
```
longhotsummer: This should call FlynoteDocumentCount.xxx directly; going through tasks.py doesn't make sense. tasks.py is an "entrypoint" that other code should not call directly. Instead, both tasks.py and other code should call into the real business logic, which is in FlynoteDocumentCount. If you need a helper method to load flynote ids, check they exist, and then update their counts, then put a helper method on FlynoteDocumentCount.

peachjam/signals.py
longhotsummer: Actually, I may be wrong... it may be able to load the flynote if it just uses instance.flynote_pk, just like the above. 🤷

longhotsummer (dismissed):

longhotsummer:
Looks good. Just missing the admin command to forcibly refresh flynote counts

Dennismwangi-k:
> Looks good. Just missing the admin command to forcibly refresh flynote counts

@longhotsummer  sorry for delay but now i understand your logic back to bakc, task.py should always have the code for running background tasks, nothing more, then also my logic of having them all in 1 was kinda good but not good for now, coz for now you will have to run all roots in admin, you will not know what roots are scheduled or not

peachjam/admin.py
```diff
@@ -1427,6 +1429,34 @@ def get_action_queryset_roots(self, queryset):
...
 
+    def get_action_target_roots(self, queryset):
+        root_paths = {
+            path[: self.model.steplen]
+            for path in queryset.values_list("path", flat=True)
+        }
+        return self.model.get_root_nodes().filter(path__in=root_paths).order_by("path")
+
+    @admin.action(description=_("Refresh selected flynote roots now (background)"))
+    def refresh_document_counts_now(self, request, queryset):
```
longhotsummer, line 1438: No, this must just run`FlynoteDocumentCount.refresh_for_flynote(root)` - the point is that I want to run the update immediately, not as a background task at all.

Dennismwangi-k: then no need to add any que at all when you call that function you just need to run it direct let me drop it had removed 12 hour and made it run now but its ok let me drop it

longhotsummer (approved):

Dennismwangi-k:
finally

PR #3176, opened 2026-04-30
generate summary immediately

longhotsummer:
this is useful when an editor wants to force a new summary and review it

also:
* summary task priority = -1
* show num documents in flynote admin


Dennismwangi-k (dismissed):
ooh yeah instead of adding a que, same idea with the flying note refresh

PR #3183, opened 2026-05-04
make update work citations trigger more granular

longhotsummer:
otherwise we update work citations unneccesarily on minor metadata changes, including when bulk re-generating judgment summaries.


mugiwaramunyi (approved):
Looks good!

PR #3185, opened 2026-05-04
add a command to regenerate the flying notes and summary, also fiz th…

Dennismwangi-k:
…e linebreak on the search html page


peachjam_search/templates/peachjam_search/_search_hit.html
```diff
@@ -70,7 +70,7 @@ <h5 class="card-title">
         {% if hit.document.matter_type %}<div>{{ hit.document.matter_type }}</div>{% endif %}
         {% if hit.document.doc_type == "judgment" %}
           {% if hit.document.blurb %}<div class="my-2">{{ hit.document.blurb }}</div>{% endif %}
-          {% if hit.document.flynote %}<div class="fst-italic my-2">{{ hit.document.flynote|safe }}</div>{% endif %}
+          {% if hit.document.flynote %}<div class="fst-italic my-2">{{ hit.document.flynote|linebreaksbr }}</div>{% endif %}
```
longhotsummer: Perhaps split these two lines into a separate include that we call in the 3-4 places in various templates where we render this.

NOT search_alert.email, since that deliberately uses simpler html.

longhotsummer (approved):

PR #3200, opened 2026-05-07
change flying note results from div ut ul for better accesibility

Dennismwangi-k:
put heading and filter inside the padding and card body and list items outside the card body, flush to the card edges 

Before:
Screenshot 2026-05-07 at 2 55 01 PM
Screenshot 2026-05-07 at 2 55 14 PM

Screenshot 2026-05-07 at 1 38 39 PM

After:

Screenshot 2026-05-07 at 2 56 08 PM

Screenshot 2026-05-07 at 1 39 08 PM

### DIFF
--- a/peachjam/templates/peachjam/flynote/_filter_form.html
+++ b/peachjam/templates/peachjam/flynote/_filter_form.html
@@ -14,4 +14,6 @@
     <div class="progress-bar progress-bar-striped progress-bar-animated w-100"></div>
   </div>
 </form>
-<div id="flynote-list-container" class="mt-1">{% include 'peachjam/flynote/_list.html' %}</div>
+{% if not omit_results %}
+  <div id="flynote-list-container" class="mt-1">{% include 'peachjam/flynote/_list.html' %}</div>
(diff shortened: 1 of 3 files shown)


peachjam/templates/peachjam/flynote/_filter_form.html
```diff
@@ -14,4 +14,6 @@
     <div class="progress-bar progress-bar-striped progress-bar-animated w-100"></div>
   </div>
 </form>
-<div id="flynote-list-container" class="mt-1">{% include 'peachjam/flynote/_list.html' %}</div>
+{% if not omit_results %}
```
longhotsummer, line 17: why is this necessary? won't this mean that there is no container for results to be injected into?

Dennismwangi-k: @longhotsummer its only omitted on teh subtopics card layout only, the result container still exxist in pupular html. The default behaviour is still unchanged elsewhere.

longhotsummer (approved):

PR #3203, opened 2026-05-07
Clear the node_order_by on Flynote

longhotsummer:
we don't need the Flynote tree to be ordered; we order when necessary. With node_order_by set, when we change flynote names we need to recompute paths, which we aren't doing.

Simplest is to just clear ordering. This also simplifies move operations.


Dennismwangi-k (dismissed):
Lgtm

PR #3204, opened 2026-05-07
Remove node_order_by from Taxonomy

longhotsummer:
We don't need the taxonomy to maintain its order, we order nodes during display. This simplifies tree management and prevents treebeard's MP_Node's path from getting broken when we change the name of a node, but don't re-fix the tree.

See also https://github.com/laws-africa/peachjam/pull/3203


Dennismwangi-k (approved):
lgtm

PR #3208, opened 2026-05-11
Arbitration judgment view

mugiwaramunyi:
Adds the necessary detail view for viewing Arbitration related judgments.
Also adds static image for the card view.
Adds some tests for the new view as well.

Screenshot 2026-05-11 at 10 35 54

Screenshot 2026-05-11 at 10 41 51


peachjam/views/arbitration.py
```diff
@@ -55,6 +59,135 @@ class ArbitrationAwardListView(FilteredDocumentListView):
...
+        context["facet_data"] = {}
+        self.add_taxonomies_facet(context)
+        self.add_courts_facet(context)
+        self.add_years_facet(context)
+        self.add_judges_facet(context)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["arbitration_categories"] = self.get_arbitration_categories()
+        context["doc_table_show_jurisdiction"] = True
```
Dennismwangi-k, line 148: this line will make the page crush coz enabling the jurisdiction column causes the template to render jurisdiction for group rows too to the template and its not designed to hundle the group data only the real data. Is it running well on your machine coz i tried and it crushed

PR #3212, opened 2026-05-13
CLean flying note script

Dennismwangi-k:
This script wil clean the second level flying note thats duplicate the first lavel name and add slash between them. like this one.
- Administrative law > Administrative/constitutional duty
- Defamation > Defamation/communications

### DIFF
--- a/peachjam/management/commands/clean_flynote_child_prefixes.py
+++ b/peachjam/management/commands/clean_flynote_child_prefixes.py
@@ -1,0 +1,193 @@
+import re
+from collections import defaultdict
+
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from peachjam.analysis.flynotes import FlynoteParser
+from peachjam.models import Judgment
+from peachjam.models.flynote import Flynote, FlynoteDocumentCount, JudgmentFlynote
+
+LAW_SUFFIX_RE = re.compile(r"\s+law$", re.I)
+
+
+def clean_spaces(value):
+    return " ".join((value or "").split()).strip()
+
+
+def prefix_variants(root_name):
+    root_name = clean_spaces(root_name)
+    variants = [root_name]
+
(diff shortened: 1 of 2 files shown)


peachjam/management/commands/clean_flynote_child_prefixes.py
```diff
@@ -0,0 +1,193 @@
...
+from peachjam.models.flynote import Flynote, FlynoteDocumentCount, JudgmentFlynote
+
+LAW_SUFFIX_RE = re.compile(r"\s+law$", re.I)
+
+
+def clean_spaces(value):
+    return " ".join((value or "").split()).strip()
+
+
+def prefix_variants(root_name):
```
longhotsummer, line 18: I think this is being too smart to guess the prefixes, and making this command too specific and dangerous.

I think it will be better for us to be able to explicitly say "under this root, replace the prefix X with Y". We will want to do that multiple times, with carefully chosen prefixes.

For example, all these depth=2 flynotes need to be cleaned, and we should be able to do so manually in bulk - I don't want the code guessing at what changes it must make, we need to retain precise control.

- Administrative/foo
- Administrative / foo
- Administrative law / foo
- Administrative law/foo

A helper classmethod on the Flynote object that takes a root, a regular expression, and a replacement pattern, would be a very flexible building block.

Dennismwangi-k: ok what if we dry run it and export the excel sheet to see how its gonna behave, coz i tried on my local data and it did a very good job you can see it on the notion documentation 

https://www.notion.so/Flynote-cleanup-for-redundant-parent-prefixes-in-child-topics-35e233d841b1801f992ffa4c8e775f9a

maybe if i can do the same i dry run it on server and see how it will behave and investigate the data how will locate and see if it will work soo smart like that we go with it

longhotsummer (approved):

PR #3223, opened 2026-05-14
this is a fix on the flyong note script to avoide the duplicates of the children

Dennismwangi-k:
i have updated this script to avoide createing same flying note on the second level after renaming it. so i did a dry run on my local data and this is how should look like in below data but focus on this example
""TARGET  47185: 'Public finance' from ROOT 2443: Contract law -> merge target in ROOT 3769: Procurement law""

It means:
  - there is already a flynote node called Public finance
  - its id is 47185
  - right now it is under the root Contract law
  - while processing the root Procurement law, the script found a child like:
      - Procurement law / public finance
  - after cleaning that child name, the final label becomes:
      - public finance
  - instead of creating a new Public finance under Procurement law, the script
    wants to reuse the existing Public finance node
  - so on apply, it would move node 47185 from Contract law to Procurement law
  - then it would merge the prefixed procurement duplicate into that moved
    node

  Before:

  Contract law
    Public finance

  Procurement law
    Procurement law / public finance

  After:

  Contract law
    [Public finance no longer here]

  Procurement law
    Public finance

  and then:

  Procurement law / public finance

  gets merged into that Public finance

Here is teh rest of dry run

ROOT 3769: Procurement law
TARGET  47185: 'Public finance' from ROOT 2443: Contract law -> merge target in ROOT 3769: Procurement law
MERGE   62812: 'Procurement law / public finance' -> 47185
MERGE   61594: 'Procurement law / public procurement' -> 3770
MERGE   64690: 'Procurement/tender law' -> 13146
MERGE   40591: 'Procurement/tender review' -> 25758
TARGET  17028: 'tenders' from ROOT 2681: Administrative law -> merge target in ROOT 3769: Procurement law
MERGE   11746: 'Procurement/tenders' -> 17028
ROOT DONE: targets_moved=2, renamed=0, merged=5, judgments_updated=0

### DIFF
--- a/peachjam/management/commands/clean_flynote_child_prefixes.py
+++ b/peachjam/management/commands/clean_flynote_child_prefixes.py
@@ -63,6 +63,7 @@ class Command(BaseCommand):
             )
 
         total_roots_touched = 0
+        total_targets_moved = 0
         total_renamed = 0
         total_merged = 0
         total_judgments_updated = 0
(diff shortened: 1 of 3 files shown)


longhotsummer (approved):
So much cleaner!

PR #3233, opened 2026-05-19
fix the issue of add child none in the flynote path

Dennismwangi-k:
This pr addresses this issue #3221  whereby the issue was under the flying note updater when creating the child node under the existing parent whereby parent is saying it has a child but in actual the metadata is lying there is no child existing so it returns none.

### DIFF
--- a/peachjam/analysis/flynotes.py
+++ b/peachjam/analysis/flynotes.py
@@ -2832,7 +2832,6 @@ class FlynoteUpdater:
         can proceed safely.
         """
         locked_parent = Flynote.objects.select_for_update().get(pk=parent.pk)
-        locked_parent.refresh_from_db(fields=["path", "depth", "numchild"])
 
         if locked_parent.numchild and locked_parent.get_last_child() is None:
             log.warning(
--- a/peachjam/analysis/flynotes.py
+++ b/peachjam/analysis/flynotes.py
@@ -2822,6 +2822,29 @@ class FlynoteUpdater:
         self.node_cache = {}
         self.update_counts = update_counts
 
+    def get_parent_for_child_insert(self, parent):
+        """Lock and refresh the parent row before inserting a child.
+
+        Treebeard decides whether a node is a leaf from ``numchild`` on the
+        in-memory parent. In production we have seen parent rows where
+        ``numchild`` says children exist but the last-child lookup returns
+        nothing. Repair that drift before calling ``add_child()`` so the insert
+        can proceed safely.
+        """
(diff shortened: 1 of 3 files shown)


peachjam/analysis/flynotes.py
```diff
@@ -2822,6 +2822,29 @@ def __init__(self, assume_clean=True, update_counts=True):
...
+        """Lock and refresh the parent row before inserting a child.
+
+        Treebeard decides whether a node is a leaf from ``numchild`` on the
+        in-memory parent. In production we have seen parent rows where
+        ``numchild`` says children exist but the last-child lookup returns
+        nothing. Repair that drift before calling ``add_child()`` so the insert
+        can proceed safely.
+        """
+        locked_parent = Flynote.objects.select_for_update().get(pk=parent.pk)
+        locked_parent.refresh_from_db(fields=["path", "depth", "numchild"])
```
Dennismwangi-k: no need for the second refresh

Dennismwangi-k: let me drop this

peachjam/tests/test_admin.py
longhotsummer, line 1: this can all be reverted

peachjam/analysis/flynotes.py
```diff
@@ -2822,6 +2822,28 @@ def __init__(self, assume_clean=True, update_counts=True):
...
+        """
+        locked_parent = Flynote.objects.select_for_update().get(pk=parent.pk)
+
+        if locked_parent.numchild and locked_parent.get_last_child() is None:
+            log.warning(
+                "Repairing flynote %s before add_child(): numchild=%s but no last child exists.",
+                locked_parent.pk,
+                locked_parent.numchild,
+            )
+            Flynote.objects.filter(pk=locked_parent.pk).update(numchild=0)
```
longhotsummer, line 2842: I hate the fact that we even have to do this, because it feels like a hack, but let's see if it fixes things.

longhotsummer (changes requested):

peachjam/analysis/flynotes.py
Dennismwangi-k: yeah please coz we have to try to lock it so we see if its going to fix the race issue

Dennismwangi-k: if there is a way we can think of in a better way to lock this roots and make sure there is no race issue can be teh best coz we are patching the lock code almost everywhere

longhotsummer (approved):

PR #3236, opened 2026-05-20
ensure libreoffice is installed in test workflow

Dennismwangi-k:
lets install the soffice with apt and stop relying on caching, it will still be fast but not as relying on caching but now we need to solve the issue of soffice not get installed well

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,6 +64,14 @@ jobs:
           packages: libreoffice poppler-utils pandoc
           version: test-apt-v1
           execute_install_scripts: true
+      - name: Inspect cached system packages
+        run: |
+          which pandoc || true
+          pandoc --version || true
+          which pdftotext || true
+          pdftotext -v || true
+          which soffice || true
+          soffice --version || true
       - name: Ensure system packages are installed
         run: |
           sudo apt-get update
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,7 @@ jobs:
           packages: libreoffice poppler-utils pandoc
           version: test-apt-v1
           execute_install_scripts: true
-      - name: Install system packages
+      - name: Install libreoffice
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -62,8 +62,12 @@ jobs:
         uses: awalsh128/cache-apt-pkgs-action@v1.6.0
         with:
           packages: libreoffice poppler-utils pandoc
-          version: test-apt-v1
+          version: test-apt-v3
           execute_install_scripts: true
+      - name: Install system packages
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libreoffice poppler-utils pandoc
```
longhotsummer:
```suggestion
          sudo apt-get install -y libreoffice
```

.github/workflows/test.yml
```diff
@@ -62,8 +62,12 @@ jobs:
         uses: awalsh128/cache-apt-pkgs-action@v1.6.0
         with:
           packages: libreoffice poppler-utils pandoc
-          version: test-apt-v1
+          version: test-apt-v3
           execute_install_scripts: true
+      - name: Install system packages
```
longhotsummer:
```suggestion
      - name: Install libreoffice
```

.github/workflows/test.yml
```diff
@@ -62,8 +62,12 @@ jobs:
         uses: awalsh128/cache-apt-pkgs-action@v1.6.0
         with:
           packages: libreoffice poppler-utils pandoc
-          version: test-apt-v1
+          version: test-apt-v3
```
longhotsummer:
```suggestion
          version: test-apt-v1
```

Changing this is clearing the cache, which means everything installs cleanly again as if there is no cache, which is masking the behaviour.

longhotsummer (changes requested):
I've made small suggestions.

longhotsummer (approved):

PR #3240, opened 2026-05-21
update the email template so to make the flying notes italic and also…

Dennismwangi-k:
update the email template so to make the flying notes italic and also create some spacing between the list. i didnt use exact padding coz i have seen the structure of styling email templates we are using the margin, so i have just followed it and still will achieve the same goal we want.

### DIFF
--- a/peachjam/templates/peachjam/emails/_alert_document_list_item.html
+++ b/peachjam/templates/peachjam/emails/_alert_document_list_item.html
@@ -1,6 +1,6 @@
 {% load peachjam %}
 <li style="margin-bottom: 1rem">
   <a href="{% absolute_url site url_path %}?utm_campaign={{ utm_campaign }}&utm_source=alert&utm_medium=email">{{ title }}</a>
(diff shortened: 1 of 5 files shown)


peachjam/templates/peachjam/emails/_alert_document_list_item.html
```diff
@@ -0,0 +1,6 @@
+{% load peachjam %}
+<li style="margin-bottom: 1rem">
+  <a href="{% absolute_url site url_path %}?utm_campaign={{ utm_campaign }}&utm_source=alert&utm_medium=email">{{ title }}</a>
+  {% if blurb %}<div>{{ blurb|striptags }}</div>{% endif %}
```
longhotsummer:
```suggestion
  {% if blurb %}<div>{{ blurb }}</div>{% endif %}
```

longhotsummer: blurb is plain text

longhotsummer (dismissed):

longhotsummer (approved):

PR #3241, opened 2026-05-21
"All users" permission group that supports anonymous permissions

longhotsummer:
This allows us to assign permissions (especially feature permissions) to anonymous users, not just authenticated users.

This means we can enable features for anonymous users on some websites, but restrict them on others.


Dennismwangi-k (approved):
lgtm

PR #3254, opened 2026-05-27
improve flynote manager to add a created at time stamp 

Dennismwangi-k:
improve flynote manager to add a created at time stamp so that the admin can trace whne the flynote was created. So it will show the tags of new if that flynote is 2 days old once is older than 2 days it will disapear. this will help the admin trace the flynotes that are created in 2 days, we can add more days if you want us to track maybe flynotes which are more than 3,4,5 or 6 days preferring with what we think is the best.

Screenshot 2026-05-27 at 5 23 43 PM
Screenshot 2026-05-27 at 5 24 01 PM


peachjam/views/flynotes.py
```diff
@@ -42,6 +44,7 @@ class FlynoteManagerView(TemplateView):
 
 class FlynoteManagerMixin(FlynoteViewMixin):
     model = Flynote
+    new_flynote_window = timezone.timedelta(days=2)
```
longhotsummer, line 47: Check with Thirusha -- perhaps 7 days?

longhotsummer (approved):
looks good, we may need to tweak the threshold for "new" -- I think 7 days is better, otherwise by Monday everything created on Friday is old.

PR #3274, opened 2026-06-03
Improvements to the comparison view (beta feature)

longhotsummer:
This makes the comparison view more robust and useful:

* the user can change the documents to compare
* the user can change the provisions to compare
* use embeddings to provide suggestions of similar documents, where possible
* the page is loaded with two uris: uri-a and uri-b, which contain the full FRBR URI including the portion to look at. If one is missing, or a portion is missing, the user is prompted to choose a document and/or a provision.

This view is only currently used with the Law Reader on AfricanLII. I want to push this into production to see how well it works on sites that have actual embedding data. Later we will link it into the actual sites.

image

image

image


Dennismwangi-k (approved):
lgtm

PR #3286, opened 2026-06-05
This is he abbitration legislation feature

Dennismwangi-k:
This is teh feature for abitration legislation, whereby can help user be able to compare the legistlation acts for abitratioon between countries
Screenshot 2026-06-09 at 10 16 20 AM
Screenshot 2026-06-09 at 10 16 04 AM
Screenshot 2026-06-09 at 10 15 57 AM
Screenshot 2026-06-09 at 10 15 47 AM
Screenshot 2026-06-09 at 10 15 30 AM


peachjam/adapters/indigo.py
```diff
@@ -970,6 +971,225 @@ def check_for_deleted(self, docs):
...
+    * token: API token
+    * api_url: URL to API base (no ending slash)
+    * dataset_id: Indigo enrichment dataset ID
+    * taxonomy_topic_root: src-root:target-root mapping for taxonomy import
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        auth_scheme = self.settings.get("auth_scheme", "Bearer")
+        self.client.headers.update(
```
longhotsummer: IndigoAdapter already sets this

peachjam/adapters/indigo.py
```diff
@@ -970,6 +971,225 @@ def check_for_deleted(self, docs):
...
+        missing_topic_slugs = set()
+        missing_work_frbr_uris = set()
+        for enrichment in dataset.get("enrichments", []):
+            topic_slug = self.normalize_taxonomy_topic_slug(
+                enrichment.get("taxonomy_topic")
+            )
+            topic = tree_mapping.get(topic_slug)
+            if topic is None and topic_slug in root_mapping:
+                topic = Taxonomy.objects.filter(slug=root_mapping[topic_slug]).first()
+            if topic is None:
```
longhotsummer, line 1079: if the topic doesn't exist, the import must fail, because we're otherwise losing important data. `IndigoAdapter.import_taxonomy_tree()` should mean that the whole tree exists by this point.

peachjam/adapters/indigo.py
```diff
@@ -970,6 +971,225 @@ def check_for_deleted(self, docs):
...
+                "provision_eid": enrichment.get("provision_id") or None,
+                "topic": topic,
+            }
+            text = remote_topics.get(topic_slug, {}).get("name", topic.name)
+            existing = ProvisionTopicEnrichment.objects.filter(**attrs).order_by("pk")
+            obj = existing.first()
+            if obj is None:
+                obj = ProvisionTopicEnrichment.objects.create(**attrs, text=text)
+            elif obj.text != text:
+                obj.text = text
```
longhotsummer: the text field is not used for ProvisionTopicEnrichment and can be ignored entirely.

Dennismwangi-k: ok

peachjam/adapters/indigo.py
```diff
@@ -970,6 +971,225 @@ def check_for_deleted(self, docs):
...
+            slug_without_root_namespace = root_slug.removeprefix("enrichments-")
+            if (
+                slug != root_slug
+                and not slug.startswith(root_slug)
+                and slug.startswith(slug_without_root_namespace)
+            ):
+                return f"enrichments-{slug}"
+        return slug
+
+    def ensure_taxonomy_roots(self):
```
longhotsummer: pls check this against the superclass's import_taxonomy_tree, which probably makes this redundant.

peachjam/adapters/indigo.py
```diff
@@ -970,6 +971,225 @@ def check_for_deleted(self, docs):
...
+                "slug", flat=True
+            )
+        )
+        for src, target in root_mapping.items():
+            if target in existing_roots:
+                continue
+            remote_root = remote_roots[src]
+            self.create_taxonomy_root(remote_root, target)
+
+    def get_taxonomy_topic_root_mapping(self):
```
longhotsummer: superclass should already do this?

peachjam/adapters/indigo.py
Dennismwangi-k: yeah i have seen we have authorisation in line 80 let me remove that redudant code

peachjam/adapters/indigo.py
```diff
@@ -970,6 +971,225 @@ def check_for_deleted(self, docs):
...
+            raise ValueError("dataset_id setting is required")
+        if not self.settings.get("taxonomy_topic_root"):
+            raise ValueError("taxonomy_topic_root setting is required")
+
+        self.dataset_id = self.settings["dataset_id"]
+        self.taxonomy_topic_root = self.settings["taxonomy_topic_root"]
+
+    def check_for_updates(self, last_refreshed):
+        """Refresh the dataset whenever the ingestor runs."""
+        self.import_dataset()
```
longhotsummer: `check_for_updates` should check the `updated_at` field of the dataset, and only if that is after `last_refreshed` should it return something to indicate that `update_document()` should be called because the dataset must be re-imported.

Otherwise we'll re-import the whole dataset every time this runs.

peachjam/views/legislation.py
```diff
@@ -132,6 +141,215 @@ def get_context_data(self, **kwargs):
         return context
 
 
+@method_decorator(never_cache, name="dispatch")
+class ArbitrationLegislationTopicListView(TemplateView):
```
longhotsummer: these views are specific to the arbitration hub. Please move them into the `arbitration_hub` django app.

peachjam/adapters/indigo.py
```diff
@@ -970,6 +980,208 @@ def check_for_deleted(self, docs):
...
+            if (
+                slug != root_slug
+                and not slug.startswith(root_slug)
+                and slug.startswith(slug_without_root_namespace)
+            ):
+                return f"enrichments-{slug}"
+        return slug
+
+    def ensure_taxonomy_roots(self):
+        """Create local taxonomy roots before importing their descendants.
```
longhotsummer: I would rather we don't automatically create roots locally, they must be created manually, so that we know it's connecting to the correct thing. So the superclass's implementation should be sufficient and you shouldn't need this.

longhotsummer (approved):

PR #3293, opened 2026-06-09
fix the small bug to show the  division specific labels that editors …

Dennismwangi-k:
fix the small bug to show court codes for court registry choices

Before:
Screenshot 2026-06-09 at 4 44 31 PM

Now:
Screenshot 2026-06-09 at 4 47 41 PM

### DIFF
--- a/peachjam/admin.py
+++ b/peachjam/admin.py
@@ -2246,7 +2246,8 @@ class ExternalDocumentAdmin(DocumentAdmin):
 class CourtRegistryAdmin(BaseAdmin):
     help_topic = "site-admin/add-court-registries"
     readonly_fields = ("code",)
-    list_display = ("name", "code")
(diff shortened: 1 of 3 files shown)


peachjam/admin.py
```diff
@@ -2246,7 +2246,8 @@ def get_form(self, request, obj=None, **kwargs):
 class CourtRegistryAdmin(BaseAdmin):
     help_topic = "site-admin/add-court-registries"
     readonly_fields = ("code",)
-    list_display = ("name", "code")
+    list_display = ("court", "name", "code")
+    list_select_related = ("court",)
```
longhotsummer, line 2250: Redundant, because CourtRegistryManager already does this

longhotsummer (approved):

PR #3294, opened 2026-06-10
enrichment dataset adapter also fetches works

longhotsummer:
updates the enrichment dataset ingestor to also update works/documents that are enriched, so that we don't need a second ingestor to fetch those works.


Dennismwangi-k (approved):
lgtm

PR #3310, opened 2026-06-11
default metadata_json to be {}

longhotsummer:
this means we don't need to guard against None in our code

fixes https://github.com/laws-africa/peachjam/issues/3305


Dennismwangi-k (dismissed):
lgtm

mugiwaramunyi (approved):

PR #3315, opened 2026-06-12
Ingestor improvements

longhotsummer:
* fix issue when queue up task for new ingestors that either didn't happen, or happened with a null id
* link ingestor background tasks to the ingestor
* show a list of pending background tasks linked to an ingestor
* ingestor admin commands queue up tasks immediately


peachjam/models/ingestors.py
longhotsummer, line 1: this changes file line endings from windows to unix

mugiwaramunyi (approved):

PR #3316, opened 2026-06-12
Similar provisions in provision detail view

longhotsummer:
* adds a re-usable LegislationProvisionListView and templates for use in views that list legislation provisions
* expands the current "provision citations" view into a provision detail view
* adds "Similar provisions" to the provision detail view
* standardise mechanism for showing jurisdiction

image

image

image


Dennismwangi-k (approved):
lgtm

PR #3318, opened 2026-06-15
When hovering over an AKN provision, show a context menu

longhotsummer:
This makes context-related options more available.

Actions (similar/compare only available on sites with embeddings)

* show similar provisions
* compare with other provisions
* comment

image

The text-selection popup also has these new options on sites with embeddings:

image


Dennismwangi-k (approved):
lgtm

PR #3319, opened 2026-06-15
I have Implemented entity search results for courts and judges.

Dennismwangi-k:
Search now keeps the existing document search flow intact, but also matches the raw query against known Court and Judge records. Strong entity matches are rendered above the normal document results using the existing server rendered `results_html` path.

Screenshot 2026-06-15 at 7 57 38 PM

Screenshot 2026-06-15 at 8 56 01 PM

### DIFF
--- a/peachjam/js/components/FindDocuments/index.vue
+++ b/peachjam/js/components/FindDocuments/index.vue
@@ -196,7 +196,6 @@
           <div class="col-md-12 col-lg-9 position-relative">
             <div>
               <FacetBadges v-model="facets" :permissive="searchInfo.count === 0" />
-              <div id="saved-search-button" />
               <div
                 id="saved-search-modal"
                 class="modal fade"
@@ -206,6 +205,11 @@
               >
                 <div id="saved-search-modal-dialog" class="modal-dialog" />
               </div>
+              <div
+                v-if="searchInfo.entity_results_html"
+                v-html="searchInfo.entity_results_html"
+              />
+              <div id="saved-search-button" />
               <div v-if="searchInfo.count">
                 <div class="my-3 d-flex">
                   <div class="me-2">
--- a/peachjam_search/entity_matcher.py
+++ b/peachjam_search/entity_matcher.py
@@ -2,10 +2,11 @@ import re
 import string
 from dataclasses import dataclass
 from typing import Iterable
+from urllib.parse import urlencode
 
 from django.core.cache import cache
 from django.db.models import Model, QuerySet
-from django.urls import reverse
+from django.urls import NoReverseMatch, reverse
 from django.utils.translation import gettext_lazy as _
 
 from peachjam.models import Court, Judge, Locality
@@ -145,10 +146,14 @@ class LocalityEntityProvider(EntityProvider):
         return super().get_queryset().select_related("jurisdiction")
 
     def get_url(self, entity) -> str:
-        return reverse(
-            "locality_legislation_list",
-            kwargs={"code": entity.place_code()},
-        )
+        try:
+            return reverse(
+                "locality_legislation_list",
+                kwargs={"code": entity.place_code()},
+            )
+        except NoReverseMatch:
+            # Some site URL configs don't have a locality legislation route.
+            return f"{reverse('legislation_list')}?{urlencode({'localities': entity.name})}"
 
     def match(self, query: str, normalized_query: str) -> list[CandidateMatch]:
         matches = []
--- a/peachjam_search/entity_matcher.py
+++ b/peachjam_search/entity_matcher.py
@@ -5,9 +5,10 @@ from typing import Iterable
 
 from django.core.cache import cache
 from django.db.models import Model, QuerySet
+from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
-from peachjam.models import Court, Judge
+from peachjam.models import Court, Judge, Locality
 
 
 @dataclass(frozen=True)
@@ -134,8 +135,47 @@ class JudgeEntityProvider(EntityProvider):
         return set(query_tokens).issubset(set(name_tokens))
 
 
+class LocalityEntityProvider(EntityProvider):
+    entity_type = "locality"
+    type_label = _("Locality")
+    model = Locality
+    fields = ("id", "name", "code", "jurisdiction")
+
+    def get_queryset(self) -> QuerySet:
+        return super().get_queryset().select_related("jurisdiction")
+
+    def get_url(self, entity) -> str:
+        return reverse(
+            "locality_legislation_list",
+            kwargs={"code": entity.place_code()},
+        )
+
+    def match(self, query: str, normalized_query: str) -> list[CandidateMatch]:
+        matches = []
+
+        for locality in self.get_entities():
+            normalized_names = [
+                normalize(locality.name),
+                normalize(re.sub(r"\s*\([^)]*\)", "", locality.name)),
+            ]
+            normalized_place_code = normalize(locality.place_code())
+
+            if query == locality.name:
+                matches.append(CandidateMatch(locality, "exact", 1.0))
+            elif normalized_query in normalized_names:
+                matches.append(CandidateMatch(locality, "normalized exact", 0.98))
+            elif normalized_query == normalized_place_code:
+                matches.append(CandidateMatch(locality, "place code exact", 0.98))
+
+        return matches
+
+
 class EntityMatcher:
-    default_providers = [CourtEntityProvider, JudgeEntityProvider]
+    default_providers = [
+        CourtEntityProvider,
+        JudgeEntityProvider,
+        LocalityEntityProvider,
+    ]
     max_query_length = 50
     _instance = None
 
--- a/peachjam_search/entity_matcher.py
+++ b/peachjam_search/entity_matcher.py
@@ -3,6 +3,7 @@ import string
 from dataclasses import dataclass
 from typing import Iterable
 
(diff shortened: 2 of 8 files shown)


peachjam_search/entity_matcher.py
```diff
@@ -0,0 +1,158 @@
...
+                provider.build_hit(match)
+                for match in provider.match(query, normalized_query)
+            )
+
+        return sorted(matches, key=lambda hit: hit.confidence, reverse=True)
+
+
+def normalize(value: str) -> str:
+    value = value.casefold()
+    value = value.translate(str.maketrans("", "", string.punctuation))
```
longhotsummer, line 222: comment explaining what this does, pls

Dennismwangi-k: sure

peachjam_search/entity_matcher.py
```diff
@@ -0,0 +1,158 @@
...
+    value = re.sub(r"\s+", " ", value).strip()
+    return value
+
+
+def tokenize(value: str) -> list[str]:
+    return [token for token in value.split() if len(token) >= 3]
+
+
+def is_judge_token_match(query_tokens: list[str], name_tokens: list[str]) -> bool:
```
longhotsummer: move this onto `JudgeEntityProvider`

longhotsummer: docstring explaining how this works pls

peachjam_search/entity_matcher.py
```diff
@@ -0,0 +1,158 @@
...
+class EntitySearchHit:
+    entity_type: str
+    entity_id: int
+    label: str
+    url: str
+    match_type: str
+    confidence: float
+
+    @property
+    def type_label(self):
```
longhotsummer: why not put this into the provider and populate it during `build_hit`, so that you don't need a lookup dict, and can configure it on the provider at the same time `entity_type` is provided

longhotsummer: that means a provider is self-contained

peachjam_search/views/search.py
```diff
@@ -231,6 +234,14 @@ def make_search_engine(self, form):
 
         return engine
 
+    def make_entity_matcher(self):
+        return EntityMatcher()
+
+    def match_entities(self, engine):
+        if engine.page != 1 or engine.field_queries:
```
longhotsummer, line 247: if the length of the query is very long, then there's no point in running this matcher. eg more than 50 chars.

peachjam_search/views/search.py
longhotsummer, line 135: if you send the entity results back as a separate entry in this dict, then you can position them separately to the main search results block. eg `entity_results_html` or something.

peachjam_search/templates/peachjam_search/_search_hit_list.html
```diff
@@ -1,4 +1,15 @@
 <ul class="list-unstyled search-result-list">
+  {% if entity_hits %}
+    <li class="mb-4">
+      <div class="row g-3">
+        {% for entity_hit in entity_hits %}
+          <div class="{% if entity_hits|length > 1 %}col-md-6{% else %}col-12{% endif %}">
```
longhotsummer: I think always using `col-12 col-md-4` is better, because it gives the entity a mini card, rather than being full page width, so it stands out from a usual hit.

peachjam_search/entity_matcher.py
```diff
@@ -0,0 +1,160 @@
...
+    entity_type = "judge"
+    type_label = _("Judge")
+    model = Judge
+
+    def match(self, query: str, normalized_query: str) -> list[CandidateMatch]:
+        matches = []
+        token_matches = []
+        query_tokens = tokenize(normalized_query)
+
+        for judge in self.get_queryset().only("id", "name"):
```
longhotsummer: we're going to be repeating this for every search, when this data changes maybe once or twice a day. 

The engine and the matchers are stateless. Perhaps create a singleton matcher that gets re-used, and it creates all its providers up front, and they can then use django's caching decorators to cache this data for, say, 1 hour or something.

peachjam_search/entity_matcher.py
```diff
@@ -32,11 +32,22 @@ class CandidateMatch:
...
     model = None
+    fields = ("id", "name")
+    cache_timeout = 60 * 60
 
     def get_queryset(self) -> QuerySet:
         return self.model.objects.all()
 
+    def get_entities(self) -> list[Model]:
+        cache_key = f"peachjam-search-entity-provider:{self.entity_type}:v1"
+        return cache.get_or_set(
```
Dennismwangi-k, line 45: This provider level cache is kinda better @longhotsummer  so we don’t reload the court/judge/locality lists on every search. Maybe this is much lighter than the similar provisions path, since entity matching is now  cached DB entity lists plus string matching, while similar provisions seems to do heavier work around ContentChunk.get_similar_provisions(...)what do you think?.

peachjam_search/templates/peachjam_search/_entity_search_hit_list.html
```diff
@@ -0,0 +1,11 @@
+{% if entity_hits %}
+  <ul class="list-unstyled search-result-list p-2 bg-light rounded my-3">
```
longhotsummer: just make this a div? since it only has one li

peachjam_search/templates/peachjam_search/_entity_search_hit.html
```diff
@@ -1,10 +1,8 @@
 <article class="card h-100">
```
longhotsummer: `bg-white`

peachjam_search/templates/peachjam_search/_entity_search_hit.html
```diff
@@ -1,10 +1,8 @@
...
-  <div class="card-body">
-    <div class="mb-2">
-      <span class="badge text-bg-secondary">{{ entity_hit.type_label }}</span>
-    </div>
-    <h5 class="card-title">
+  <div class="card-body d-flex align-items-start gap-2 bg-light">
```
longhotsummer:
```suggestion
  <div class="card-body d-flex align-items-start gap-2">
```

peachjam_search/templates/peachjam_search/_entity_search_hit.html
```diff
@@ -1,10 +1,8 @@
...
+    <h5 class="card-title mb-0 flex-grow-1">
       <a class="h5 text-primary" href="{{ entity_hit.url }}">{{ entity_hit.label }}</a>
```
longhotsummer:
```suggestion
      <a href="{{ entity_hit.url }}">{{ entity_hit.label }}</a>
```

peachjam_search/templates/peachjam_search/_entity_search_hit.html
```diff
@@ -1,10 +1,8 @@
 <article class="card h-100">
-  <div class="card-body">
-    <div class="mb-2">
-      <span class="badge text-bg-secondary">{{ entity_hit.type_label }}</span>
-    </div>
-    <h5 class="card-title">
-      <a class="h5 text-primary" href="{{ entity_hit.url }}">{{ entity_hit.label }}</a>
+  <div class="card-body d-flex align-items-start gap-2">
+    <h5 class="card-title mb-0 flex-grow-1">
```
longhotsummer: remove the heading; this should just be a regular `<a>`

peachjam/js/components/FindDocuments/index.vue
```diff
@@ -206,6 +205,11 @@
               >
                 <div id="saved-search-modal-dialog" class="modal-dialog" />
               </div>
+              <div
+                v-if="searchInfo.entity_results_html"
+                v-html="searchInfo.entity_results_html"
+              />
+              <div id="saved-search-button" />
```
longhotsummer: pls move the saved search button between the search count/sort row, and the result list.

peachjam_search/templates/peachjam_search/_entity_search_hit_list.html
longhotsummer:
```suggestion
  <ul class="list-unstyled search-result-list p-2 bg-light rounded mb-3">
```

peachjam_search/templates/peachjam_search/_entity_search_hit_list.html
```diff
@@ -0,0 +1,11 @@
+{% if entity_hits %}
+  <ul class="list-unstyled search-result-list p-2 bg-light rounded mb-3">
+    <li class="mb-0">
```
longhotsummer: This can just be a div? the `ul` doesn't do anything useful since there is only one `li`

peachjam_search/templates/peachjam_search/_entity_search_hit.html
```diff
@@ -1,10 +1,6 @@
...
-  <div class="card-body">
-    <div class="mb-2">
-      <span class="badge text-bg-secondary">{{ entity_hit.type_label }}</span>
-    </div>
-    <h5 class="card-title">
-      <a class="h5 text-primary" href="{{ entity_hit.url }}">{{ entity_hit.label }}</a>
-    </h5>
+<article class="card h-100 bg-white">
+  <div class="card-body d-flex align-items-start gap-2">
+    <a class="flex-grow-1" href="{{ entity_hit.url }}">{{ entity_hit.label }}</a>
```
longhotsummer:
```suggestion
    <a class="flex-grow-1 stretched-link" href="{{ entity_hit.url }}">{{ entity_hit.label }}</a>
```

